PR #60, opened 2020-09-18
Add PHP linting with inline comments

aristath:
(includes a deliberate error to test it)


jeffikus (commented):
Just remove the intentional errors and we can get this merged.

PR #62, opened 2020-09-18
Composer improvements

### LINKED ISSUE #9: Code quality: Add Github actions for pull requests
Each pull request should be checked for PHP compatibility and to ensure that coding standards are followed.

dingo-d:
I've made some updates to `composer.json` (structure, packages, scripts). I've added a basic `phpstan` config and added a new GH action.

This will override #60 and fix #9


dingo-d:
@aristath Should I remove the extra phpcs lint action that was merged? Since this is covered by one of the steps in the code quality checks one.

aristath:
Definitely :+1: 
The one we currently have will need to be removed when this gets merged.
I think the only blocker for this one is https://github.com/WPTT/WPThemeReview/pull/257

dingo-d:
Yeah, we'll have the block on phpcs until WPThemeReview is resolved and merged

dingo-d:
Alternatively, we can just silence this error in the `phpcs.xml.dist` 🤷‍♂️

aristath:
> Alternatively, we can just silence this error in the `phpcs.xml.dist` 🤷‍♂️ 

I suppose that would be an acceptable solution so we can get this merged... We know we're using it right and the errors reported are false-positives so we can do that, and once the fix lands in the WPThemeReview CS we can remove the exception :+1:

dingo-d:
I mean, you can merge it and then do a PR with phpcs exclusion and fixes for PHPStan issues? That way we keep these two separate, as this PR is just for adding actions and configuration fixes.

Issue #68, opened 2020-09-18
Remove duplicate width/height attributes from SVG icons


audrasjb:
We shouldn’t use width and height attributes in `Twenty_Twenty_One_SVG_Icons` class.
This currently results in duplicated attributes in SVG icons.

Worth noting that there is a DocBlock with this information in the class-twenty-twenty-onesvg-icons.php file:
> * Place each <svg> source on its own array key, without adding the
> * both `width` and `height` attributes, since these are added dnamically,
> * before rendering the SVG code.

PR #70, opened 2020-09-18
Update DocBlock in class-twenty-twenty-one-svg-icons.php

pattonwebz:
I spotted that this DocBlock has a typo and doesn't read very well in #68


audrasjb (approved):
Looks good to me 👍

PR #88, opened 2020-09-20
Use aria-current attribute on footer homepage link when home is the current page

### LINKED ISSUE #87: Use aria-current attribute on footer homepage link when home is the current page
It would be nice to use aria-current attribute on footer homepage link when home is the current page.
Eg: replace
```
<div class="site-name">
	<a href="http://localhost:8888/twentytwentyone/" rel="home">Testing 2021</a>
</div>
```
 with:
```
<div class="site-name">
	<a href="http://localhost:8888/twentytwentyone/" rel="home" aria-current="page">Testing 2021</a>
</div>
```
When we are at the homepage URL (blog posts or page on front).

audrasjb:
Fixes: #87 

It would be nice to use aria-current attribute on footer homepage link when home is the current page.
Eg: replace
```
<div class="site-name">
	<a href="http://localhost:8888/twentytwentyone/" rel="home">Testing 2021</a>
</div>
```
 with:
```
<div class="site-name">
	<a href="http://localhost:8888/twentytwentyone/" rel="home" aria-current="page">Testing 2021</a>
</div>
```
When we are at the homepage URL (blog posts or page on front).


aristath (approved):
Code looks good and makes sense. :+1:

sabernhardt:
`is_front_page()` should be sufficient for both home page conditions, and the `$aria_current` part is no longer needed:

```
					if ( is_front_page() ) :
						?>
						<?php bloginfo( 'name' ); ?>
					<?php else : ?>
						<a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a>
					<?php endif;
```

pattonwebz:
> `is_front_page()` should be sufficient for both home page conditions, and the `$aria_current` part is no longer needed:

Did you confirm that for sure? I have ran across cases where `is_front_page()` and `is_home()` can be false when expecting to be true (in some cases both can be false when expecting one to be true) depending on differenc configurations of static homepage settings. IE if you set a static home page but do not set a blog page both are false I believe.

sabernhardt:
It was tested for [Trac #37011](https://core.trac.wordpress.org/ticket/37011#comment:8).

If `is_front_page()` returns false, then `( is_front_page() && is_home() )` also would return false. (I would have used these two conditionals before Sergey pointed out the redundancy.)

pattonwebz:
> It was tested for [Trac #37011](https://core.trac.wordpress.org/ticket/37011#comment:8).
> 
> If `is_front_page()` returns false, then `( is_front_page() && is_home() )` also would return false. (I would have used these two conditionals before Sergey pointed out the redundancy.)

Awesome, thanks for confirming! :D

carolinan:
Since this needs to be updated, can we also add the check for the custom logo? If there is a logo, that will show instead of the title.

audrasjb:
> `is_front_page()` should be sufficient for both home page conditions, and the `$aria_current` part is no longer needed:
> 
> ```
> 					if ( is_front_page() ) :
> 						?>
> 						<?php bloginfo( 'name' ); ?>
> 					<?php else : ?>
> 						<a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a>
> 					<?php endif;
> ```

Makes sense to me, I'll update the PR accordingly after testing this.

audrasjb:
> Since this needs to be updated, can we also add the check for the custom logo? If there is a logo, that will show instead of the title.

Yeah I'll add support for  custom logo as well :+1:

audrasjb:
I don't know what happened, but my pull request failed somehow 😞

audrasjb:
There is no conflict on my side…

Issue #36, opened 2020-09-16
Implement the design for forms, inputs and buttons


carolinan:
Implement the designs for forms/inputs/buttons.

- [x] Buttons block -filled
- [x] Buttons block -outlined

- [x] HTML buttons  (including disabled state)
- [x] Input elements
- [x]  Select lists (including blocks)

- [x] Tested standard form elements as well as widgets and blocks.
- [x] Tested focus and hover states.

carolinan:
Currently testing the color changes/contrasts when the body background is changed, could use some more eyes.

Should the border that is used on the default button block be used when the user changes the button color?

nielslange:
@carolinan & @melchoyce I'm diving into the `:hover` and `:focus` states now.

carolinan:
Button blocks with custom styles (radius, colors,) still needs to be tested more, but lets close this and open individual issues for any new problems that we run into.

PR #97, opened 2020-09-21
Update forms, inputs and buttons

carolinan:
For https://github.com/WordPress/twentytwentyone/issues/36

Updates search form markup so that the label is visible per the design.
Adjusts borders and background colors for inputs
Adds styles to buttons and button blocks.

Sample content for testing: https://gist.github.com/carolinan/53cf8c92b77c8ad6babd3b38d0a07530

- [x] -Styling for visited button blocks are added, to override the hard coded default colors in the block library.

- [x] Label styles for when a label is combined with a radio or checkbox.

- [x] Styles for disabled items

- [ ] Styles for range.

- [ ] Make sure that the default button block border styles or colors are not applied when custom styles are used.

Known issues:
-Border styles for radio and check boxes only works in some browsers

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -283,6 +283,10 @@ figcaption {
 	padding: 15px 30px;
 }
 
+.wp-block-button__link:visited {
+	color: #d1e4dd;
+}
+
 .wp-block-button__link:hover {
 	color: #39414d;
 	background-color: #d1e4dd;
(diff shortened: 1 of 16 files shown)


jffng (commented):
Looking good, I have two comments: 

1. Should the range focus styles be dotted like the other focus styles?

Screen Shot 2020-09-21 at 3 47 08 PM

2. Should radio and checkbox inputs have some default margins? Also I think the label text should have consistent font size and weight.

Screen Shot 2020-09-21 at 3 48 00 PM

aristath (commented):
Everything looks good in Chrome & Firefox on Ubuntu, with the exception of range inputs: On firefox they are not visible

Screenshot_2020-09-22 test https github com WordPress twentytwentyone pull 97 – local

melchoyce:
I wonder if it makes sense to use the browser default focus styling for range sliders in this case?

Issue #37, opened 2020-09-16
Pages - How to implement two page heading styles


carolinan:
In the design, pages with multiple columns (assumed to be columns blocks in content),
position the page title differently compared to when the content only has a single column.

Single column: Centered page title
Multiple columns: Left aligned wide page title.

How do we best implement this?
Perhaps using a body class?

trsenna:
Hi,

why Is QA harder with two custom page templates than using a checkbox? Don't you have to test CSS changes with checkbox checked and unchecked anyway?

IMHO I'd suggest page-templates 'cause it feels more natural for theme developers. A checkbox can be nice too - but as a developer what Do I do when I need to create a new page-template that has a title always? So maybe now I have to decide if I remove the checkbox options or change my markup and styles to support the feature I do not need. This is the reason why page-templates makes more sense to me.

But probably it's possible for you to make easier for developer to just disable the checkbox button programmatically. In this case it's OK and acceptable for me to go with a checkbox option.

pattonwebz:
> why Is QA harder with two custom page templates than using a checkbox? Don't you have to test CSS changes with checkbox checked and unchecked anyway?

The difference is testing against 2 pages (which are entirely replicated other than just the differnce in title) vs testing against 2 states of the same page. 

For 2 pages we need to test both pages but also ensure any code changes to one template is ported across both templates (since the everything except title markup is replicated). With one page and a toggle we just need to test both states (no code porting required).

trsenna:
Thanks @pattonwebz for your answer :+1: 

But now thinking as just changing the page state - with page templates is it not just changing states of a page too?

pattonwebz:
> with page templates is it not just changing states of a page too?

Sure is, but the template option here comes along with 95% reused code and the need to replicate changes from one into the other every time one is made. 2 checks AND 2 places to copy/paste code between.

melchoyce:
One other issue a page template could help us solve is how to align the heading here, without resorting to creating a new columns block just to contain it:

![image](https://user-images.githubusercontent.com/2846578/94048008-dd713500-fda0-11ea-9519-900be9c04771.png)

pattonwebz:
> One other issue a page template could help us solve is how to align the heading here, without resorting to creating a new columns block just to contain it:
> 
> ![image](https://user-images.githubusercontent.com/2846578/94048008-dd713500-fda0-11ea-9519-900be9c04771.png)

Is the position of this title not the same position as what is in the narrow default template?

[...]

PR #101, opened 2020-09-21
Try a "narrow title" template

kjellr:
Tries out having a separate page template for a narrower page title. As brought up in https://github.com/WordPress/twentytwentyone/issues/37#issuecomment-696235187. 

Just an idea to see how we think it's working. 

---

Screen Shot 2020-09-21 at 4 09 40 PM

Standard Post/Page Header: 

gutenberg test__p=1637

With "Narrow Title" template active: 

gutenberg test__p=1637 (1)

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -3559,6 +3559,11 @@ nav a {
 	margin-bottom: 90px;
 }
 
+.post-template-template-narrow-title .entry-header {
+	border-bottom: none;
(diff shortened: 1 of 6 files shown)


melchoyce:
Per https://github.com/WordPress/twentytwentyone/issues/37#issuecomment-700699386, we're only going to include the one page template in the theme. Closing, but thanks for working on this, @kjellr!

Issue #28, opened 2020-09-16
Update font sizes


carolinan:
The font sizes have been increased in the design and both the CSS variables
and the editor-font-sizes theme support needs to be updated.

melchoyce:
Aside from #102, what's left to update here?

carolinan:
tested and merged, looks nice and much more readable, thank you.

PR #102, opened 2020-09-21
Add mobile type sizes

jffng:
Continues addressing #28 by adding support for [mobile type sizes](https://github.com/WordPress/twentytwentyone/issues/28#issuecomment-694919289).


assets/sass/01-settings/global.scss
```diff
@@ -223,3 +223,12 @@ $baseline-unit: 10px;
 	--separator--height: 1px;
 	--separator--width: calc(6 * var(--global--spacing-horizontal));
 }
+
+@media only screen and (min-width: 652px) {
```
jffng: It would be nice to use the breakpoint mixin here but it's not available due to the order that they're currently loaded.

kjellr: Might be worth adding a comment to clarify why this doesn't use that mixin?

carolinan:
Tested, LGTM

PR #103, opened 2020-09-21
Fix search block styling

jffng:
This PR addresses a small bug in the way the search block was displayed on the front-end and in the editor.

**Before**
Screen Shot 2020-09-21 at 4 37 41 PM

**After**
Screen Shot 2020-09-21 at 4 37 16 PM


melchoyce:
Does this fix impact existing search blocks in core's version of Gutenberg? If not, let's merge 👍

Issue #44, opened 2020-09-17
Continue the code clean up


carolinan:
**Describe the bug**
There is still code inherited from Seedlet that needs to be removed.

-Block patterns, including any related CSS (Keep the file and functions for when we are ready to add the new patterns!)
-Left over CSS from removing the gradients
-The foreground color that is no longer used.
-Empty CSS selectors.

If you find anything else that you feel needs to be mentioned here, that you do not have time to work on, 
you can add it to the issue.

kjellr:
Once #118 is merged, can we close this issue and open individual issues for anything that's left? It's becoming a little hard to follow.

PR #118, opened 2020-09-23
General cleanup

carolinan:
For #44

Removes unused CSS for the site footer, two unused files, and one unused package.json script.

Updates the style.scss table of content to match the current file structure better.

Changes some of the comment styles in the scss files.

@luminuu For the SCSS structure, I think the code for the following are still in the wrong files and needs to be moved here:
```
ELEMENTS
Headings.............H1–H6
Links................Default links.
Lists................Default lists.

UTILITIES
Color Palette........Classes for the color palette colors.
Editor Font Sizes....Editor Font Sizes.
```


aristath (approved):
LGTM :+1:

assets/sass/02-tools/extends.scss
```diff
@@ -1,4 +1,5 @@
-// Button Placeholder style
+/* Button extends */
```
jffng, line 1: Do these comments need to make it into the compiled CSS? 
```suggestion
// Button extends
```

assets/sass/02-tools/mixins.scss
```diff
@@ -1,15 +1,4 @@
...
-Example:
-@mixin typography {
-	color: var(--color-charcoal);
-	font-family: var(--font-family);
-	font-size: var(--font-size-global);
-	line-height: var(--line-height-global);
-	word-wrap: break-word;
-}
-*/
+/* Mixins */
```
jffng, line 1:
```suggestion
// Mixins
```

assets/sass/01-settings/file-header.scss
```diff
@@ -11,7 +11,7 @@ Version: 1.0.0
 License: GNU General Public License v2 or later
 License URI: LICENSE
 Text Domain: twentytwentyone
-Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-menu, custom-logo, editor-style, featured-images, footer-widgets, rtl-language-support, sticky-post, threaded-comments, translation-ready
```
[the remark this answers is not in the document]
jffng: Should we add the `block-patterns` tag while we're here, or we'll do a pass at the end to ensure the theme has accurate tags?

jffng (approved):
Looks good, just a few small comments, thanks!

assets/sass/02-tools/extends.scss
carolinan: Without them in the css file the code felt out of place because there is no heading or explaination to why the button styles are loaded this early.

PR #125, opened 2020-09-23
Revise menu and submenu spacing

jffng:
This PR continues work started in #115 to get the menu styles in place. 

Specifically, the following changes are made:

- Aligns submenus according to the design
- Site branding is given some padding in case the menu wraps
- Uses `var(--primary-nav--padding)` in a more straightforward manner

Before | After
------ | ------
Screen Shot 2020-09-23 at 11 36 49 AM | Screen Shot 2020-09-23 at 11 36 15 AM


carolinan:
The spacing between the sub menu item was increased in comparison to the design to make sure that the click target size is at least 44x44 to follow WCAG level AAA: https://www.w3.org/TR/WCAG21/#target-size

jffng:
@carolinan does increasing the padding on the element that contains the link (not the link itself) actually increase the target size to pass AAA accessibility?

I.e. should hovering + clicking anywhere in this mouse area trigger its active state + allow the user to follow the link? 

Kapture 2020-09-23 at 12 49 50
_Screen capture is from trunk, not this PR_.

carolinan:
Hmm no you are right, it would need to be on the link itself.  While spacing between the items will reduce the chance of miss clicks, it doesn't help if you are using a pointer.

@jffng Found one good article on the topic: https://www.nngroup.com/articles/touch-target-size/

carolinan:
But we need to consider the bottom border style while adding the padding.

melchoyce:
It's hard for me to read through the code, but I wanted to make sure we're also using `text-decoration` instead of a border, for consistency: https://codepen.io/melchoyce/pen/jOqdNPm

jffng:
I increased the target size for submenu items and used `text-decoration` instead of a border. The results don't look so great given that `text-decoration` lacks customizability:

Kapture 2020-09-25 at 00 52 12

melchoyce:
Chrome! :shakes-fist:

Maybe at this point we should stick with the borders and try to make `text-decoration` work in a new PR?

jffng:
I think this is ready for another review. To summarize the changes, this PR: 

- Updates the menu link styles and padding to be more consistent with design
- Gives padding to the site title + description in case the menu wraps
- Increases the target area of the menu items per [this comment](https://github.com/WordPress/twentytwentyone/pull/125#issuecomment-697685018)

The responsive menu still needs some work but I'd prefer to get this in and work on that as a follow up.

Note that there is a small visual bug in Chrome with the text-underline that will hopefully be addressed when they add support for `text-underline-offset`.

melchoyce:
The menu button moves when you open it, which is kind of weird:

menu

I also find it disorienting that the page content fades out to the left, but the menu itself fades in from the bottom. Let's ditch the move in/out entirely for now.

Otherwise, LGTM 👍

carolinan (approved):

PR #124, opened 2020-09-23
add @wordpress/env to provide a easy local environment

fabiankaegy:
This PR adds the [`@wordpress/env`](https://www.npmjs.com/package/@wordpress/env) to the project to make it easier to contribute to the project when working locally. 

Everything a user needs to do after cloning the repository is run the `npm run wp-env start` command and a local WordPress installation will start running on the `localhost:8888` url. 

This resolves #123

Issue #123, opened 2020-09-23
Adding `@wordpress/env` for local development into the project


fabiankaegy:
**Is your feature request related to a problem? Please describe.**
When working with themes locally it is always a hurdle to setup a separate dev environment for it. With the [`@wordpress/env`](https://npmjs.com/package/@wordpress/env) package that was build by the Gutenberg theme this is very easy to just have one npm command: `npm run wp-env start` to start a local environment server. 

I can happily create a PR for this. But I would love your thoughts about having this as a part of the repository.

jffng:
Adding `wp-env` as a development dependency seems reasonable to me and nice a way to increase the inclusivity of who can contribute, since we're already requiring node + npm to compile the styles.

carolinan:
+1 As long as the documentation to use it is added to the contributing file for the repo.
It is actually not as straight forward to set up for the first time, especially not on Windows, so it is important that contributors know that they are not required to use it.

fabiankaegy:
@carolinan I have updated my PR (#124) to include some basic documentation about how to get started. 

Can you elaborate how granular you would like the doc's in the theme to get? / at which point should we link to the docs of the `wp-env` project?

carolinan:
You can combine what you wrote with the commands added here:
https://github.com/WordPress/twentytwentyone/pull/127/files#diff-6a3371457528722a734f3c51d9238c13R87

PR #127, opened 2020-09-23
Add wp-env with some supporting docs.

whyisjake:
To aid local development, I added some docs on how to use `wp-env` for local testing.

[Full docs are here](https://developer.wordpress.org/block-editor/packages/packages-env/).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,13 +12,13 @@ Twenty Twenty One should have all issues reported on GitHub at https://github.co
 
 ## Testing a Pull Request
 
-If you're using Git locally, you can test a pull request by pulling down the associated branch, creating a zip file of the contents, and uploading to your site. This repository includes all compiled files, so it should install just like any other uploaded theme. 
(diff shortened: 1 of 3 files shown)


carolinan:
Hi, there is already a pull request for this, can you work together with @fabiankaegy on it?

Both PR's include scripts for linting. css files. This was removed earlier because it was pointed out that if the scss lint passes, so does the css.

It is unclear to me, why was the order of the scripts changed?

fabiankaegy:
Hi @whyisjake, The PR I've been working on is #123 with the issue #124 to track the conversations about it. Would love your insights / thoughts especially on the docs side of things.

jffng:
> why was the order of the scripts changed?

Perhaps best practice to keep the scripts in alphabetical order, since @whyisjake added a few related to `wp-env`.

carolinan:
Sorry about the confusion about the script for linting CSS files. It hasn't been merged yet but was fresh in my memory.

luminuu:
Already done in #124.

PR #130, opened 2020-09-23
Add href to the site-title element

### LINKED ISSUE #109: Site title should be linked on the paged blog
**Describe the bug**

When navigating the pages on the list of latest posts, (/page/2 etc). the site title should be a link back to the front page.

metodiew:
I have added a `href` using `home_url()` function. I guess this is simple enough as a fist patch here, but let me know if there is anything I'm missing?


jffng (approved):
Works, thank you @metodiew!

PR #133, opened 2020-09-24
fix link styling to match the design provided by @melchoyce

### LINKED ISSUE #110: Update link states to match the design
The links need to be updated to match the new styles:

Regular links have a solid underline, hovered links have a dotted underlined, and focused links have a solid box shadow with a border radius

fabiankaegy:
The styles were provided in this coment: https://github.com/WordPress/twentytwentyone/issues/110#issuecomment-697896362

This resolves #110

### DIFF
--- a/assets/sass/03-generic/reset.scss
+++ b/assets/sass/03-generic/reset.scss
@@ -71,15 +71,8 @@ body {
 
 // Links styles
 a {
-	border-bottom: 1px solid var(--global--color-secondary);
 	color: var(--wp--style--color--link, var(--global--color-primary));
-	text-decoration: none;
 	transition: border-color 0.1s ease-out;
-
-	&:hover,
-	&:focus {
-		border-bottom-color: transparent;
-	}
 
 	&:hover {
 		color: var(--global--color-primary-hover);
--- a/assets/sass/04-elements/links.scss
+++ b/assets/sass/04-elements/links.scss
@@ -1,0 +1,24 @@
+/*
+ * text-underline-offset doesn't work in Chrome at all 👎
+ * But looks nice in Safari/Firefox, so let's keep it and
+ * maybe Chrome will support it soon.
+ */
+a {
+	color: var(--global--color-black);
+	text-underline-offset: 3px;
+	transition: all .15s ease-in-out;
+  }
+
+a:hover {
+	text-decoration-style: dotted;
+}
+
+/*
+ * The reason for the !important style here is the global focus styles created
+ * by the `.site *:focus {` rules in the 03-generic/reset.scss file.
+ */
+a:focus {
+	outline: none !important;
+	border: 1px solid var(--global--color-black);
+	border-radius: 5px;
+}
--- a/assets/sass/style.scss
+++ b/assets/sass/style.scss
@@ -61,6 +61,7 @@
 @import "04-elements/forms";
 @import "04-elements/media";
 @import "04-elements/misc";
+@import "04-elements/links";
 
 /* Category 05 is all about adjusting the default block styles to the given layout. I only added three blocks as examples. */
 @import "05-blocks/blocks";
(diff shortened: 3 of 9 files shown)


assets/sass/03-generic/reset.scss
```diff
@@ -71,16 +71,9 @@ body {
 
 // Links styles
 a {
-	border-bottom: 1px solid var(--global--color-secondary);
 	color: var(--wp--style--color--link, var(--global--color-primary));
-	text-decoration: none;
 	transition: border-color 0.1s ease-out;
```
carolinan, line 75: Is this still needed?

carolinan, line 75: Is the border transition still needed?

assets/sass/04-elements/links.scss
```diff
@@ -0,0 +1,24 @@
+/*
+ * text-underline-offset doesn't work in Chrome at all 👎
+ * But looks nice in Safari/Firefox, so let's keep it and
+ * maybe Chrome will support it soon.
+ */
+a {
+	color: var(--global--color-black);
```
carolinan, line 7: If the color needs to be overwritten here I believe it still needs to be     `color: var(--wp--style--color--link, var(--global--color-primary));` and not black?
Hopefully @kjellr can confirm.

assets/sass/04-elements/links.scss
```diff
@@ -0,0 +1,24 @@
...
+a:hover {
+	text-decoration-style: dotted;
+}
+
+/*
+ * The reason for the !important style here is the global focus styles created
+ * by the `.site *:focus {` rules in the 03-generic/reset.scss file.
+ */
+a:focus {
+	outline: none !important;
```
carolinan, line 21: We can not use outline none for accessibility reasons, to work on high contrast mode, you can set this to transparent instead.

assets/sass/03-generic/reset.scss
carolinan: And the border on .has-link-color ?

assets/sass/04-elements/links.scss
```diff
@@ -0,0 +1,24 @@
...
+	border: 1px solid var(--global--color-black);
```
carolinan, line 22: The problem with using a border here is that it takes up space, so when a link is focused there is a visible "jump" or shift of not only the link but the content that follows, because the extra pixels are added.

This makes a text difficult to read because the shifts are distracting.

visible-shift-of-content

The shift can be avoided by adding a transparent border to the default state of the link, 
but that would create extra spacing and move the entire link so it is not a viable option.

carolinan: Ah, as per comment https://github.com/WordPress/twentytwentyone/issues/110#issuecomment-698423081 this would be changed back to an outline. :)

assets/sass/style.scss
```diff
@@ -61,6 +61,7 @@
 @import "04-elements/forms";
 @import "04-elements/media";
 @import "04-elements/misc";
+@import "04-elements/links";
```
carolinan, line 64: Please also include the new file in the table of contents :)

assets/sass/04-elements/links.scss
kjellr: Yes, that's right. That fallback stack should get this working nicely with the Custom Link color functionality in Gutenberg.

assets/sass/style.scss
jffng: Which table of contents are you referring to? 

Looks like links are already listed in the TOC of this file: https://github.com/WordPress/twentytwentyone/blob/76ca10c37ba9be894be64467b2d9d854d6c8d235/assets/sass/style.scss#L22

carolinan: Yeah, the toc and the actual files in the folder are out of sync

PR #135, opened 2020-09-24
Adds exceptions for false-positives in PHPCS checks

aristath:
Addition to https://github.com/WordPress/twentytwentyone/pull/62


carolinan (approved):
These exceptions look correct to me.

PR #137, opened 2020-09-24
Fix PHPStan errors

aristath:
Additional fixes - Continuation of #62 & #135

### DIFF
--- a/classes/class-twenty-twenty-one-custom-colors.php
+++ b/classes/class-twenty-twenty-one-custom-colors.php
@@ -75,9 +75,11 @@ class Twenty_Twenty_One_Custom_Colors {
 	 * Customizer & frontend custom color variables.
 	 *
 	 * @access public
+	 *
+	 * @return void
 	 */
 	public function custom_color_variables() {
-		if ( 'D1E4DD' !== get_theme_mod( 'background_color', 'D1E4DD' ) ) {
+		if ( 'd1e4dd' !== strtolower( get_theme_mod( 'background_color', 'D1E4DD' ) ) ) {
 			wp_add_inline_style( 'twenty-twenty-one-style', $this->generate_custom_color_variables() );
 		}
 	}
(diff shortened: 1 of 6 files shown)


classes/class-twenty-twenty-one-custom-colors.php
```diff
@@ -86,9 +88,16 @@ public function custom_color_variables() {
...
+	 *
+	 * @return void
 	 */
 	public function editor_custom_color_variables() {
-		wp_enqueue_style( 'twenty-twenty-one-custom-color-overrides', get_template_directory_uri() . '/assets/css/custom-color-overrides.css', array(), wp_get_theme()->get( 'Version' ) );
+		wp_enqueue_style(
+			'twenty-twenty-one-custom-color-overrides',
+			get_theme_file_uri( 'assets/css/custom-color-overrides.css' ),
+			array(),
+			(string) filemtime( get_theme_file_path( 'assets/css/custom-color-overrides.css' ) )
```
dingo-d, line 99: This is ok for development, but not that ok for production. This would invalidate any caching set in place (`filemtime`).

classes/class-twenty-twenty-one-customize-color-control.php
```diff
@@ -45,7 +45,7 @@ public function enqueue() {
 			'twentytwentyone-control-color',
 			get_theme_file_uri( 'assets/js/palette-colorpicker.js' ),
 			array( 'customize-controls', 'jquery', 'customize-base', 'wp-color-picker' ),
-			filemtime( get_theme_file_path( 'assets/js/palette-colorpicker.js' ) ),
+			(string) filemtime( get_theme_file_path( 'assets/js/palette-colorpicker.js' ) ),
```
dingo-d, line 48: Same as the above comment, about caching.

inc/custom-css.php
```diff
@@ -7,39 +7,32 @@
...
+ * @return string
+ */
+function twenty_twenty_one_generate_css( $selector, $style, $value, $prefix = '', $suffix = '', $echo = true ) {
+	$return = '';
+
+	/*
+	* Bail early if we have no $selector elements or properties and $value.
+	*/
+	if ( ! $value || ! $selector ) {
+		return;
```
dingo-d: Should return string
```suggestion
		return '';
```

classes/class-twenty-twenty-one-custom-colors.php
aristath: :thinking: Why would it not be OK for production?
Caches will only be invalidated when the modification date of the file changes, which would indicate that the contents of the file have also changed. On a production environment the files don't change, and if they do then caches have to be invalidated (which wouldn't happen with a static version number). Unless I'm missing something?
I'd agree if it was `time()`, that is definitely not suitable for production :sweat_smile:

dingo-d: Oh, my bad. It's ok for caching, but I know there was a reason it is discouraged:

https://github.com/WordPress/WordPress-Coding-Standards/pull/1378#issuecomment-398784410

😄

dingo-d (approved):

PR #139, opened 2020-09-24
Revert "Add href to the site-title element"

carolinan:
Reverts WordPress/twentytwentyone#130

@metodiew I am reverting this only because the intention to was to add the link on the paged blog archives, and not on the home page.
-I meant paged as in when you click on the "newer posts" link and your url is something like http://localhost:8888/page/2/.
For this, you can use https://developer.wordpress.org/reference/functions/is_paged/ and I think it would be an easy fix for you.

The home page should not link to itself, you can find background information about why the link was removed here:
https://core.trac.wordpress.org/ticket/37011

### DIFF
--- a/template-parts/header/site-branding.php
+++ b/template-parts/header/site-branding.php
@@ -18,7 +18,7 @@
 	if ( ! empty( $blog_info ) ) {
 		if ( is_front_page() || ( is_front_page() && is_home() ) ) {
 			?>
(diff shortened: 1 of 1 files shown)


metodiew:
Thanks for the feedback, @carolinan!  Any preferences how this should be implemented? Happy to use `is_paged()` even if we'll have a chain of a few `ifs`

Would you like me to continue the discussion in https://github.com/WordPress/twentytwentyone/issues/109 and send an update once I'm ready?

PR #142, opened 2020-09-24
Update forms, inputs and buttons, replaces Pr/97

carolinan:
For https://github.com/WordPress/twentytwentyone/issues/36

Replaces https://github.com/WordPress/twentytwentyone/pull/97

Updates search form markup so that the label is visible per the design.
Adjusts borders and background colors for inputs
Adds styles to buttons and button blocks.

Sample content for testing: https://gist.github.com/carolinan/53cf8c92b77c8ad6babd3b38d0a07530

- [x] -Styling for visited button blocks are added, to override the hard coded default colors in the block library.

- [x] Label styles for when a label is combined with a radio or checkbox.

- [x] Styles for disabled items

- [x] Styles for range, use the default focus style.

- [ ] Make sure that the default button block border styles or colors are not applied when custom styles are used.

Known issues:
-Border styles for radio and check boxes only works in some browsers


carolinan:
10b5395 Adds styles for the range element, resets the label font weight to normal, and adjust colors.

Animation of the range element being adjusted, with dotted focus styles.


The styles now work better in the customizer when the background color is changed.
-But, it has the side effect that on dark background colors where the form element border is white,
the border is not visible when the element is in focus or selected, because the background is also white:

white border on text area



I was not able to get the text color inside select lists, date, and color elements to maintain a high color contrast when the body background is dark, so I changed it back to white.

The left most element is the before, and the other two are after the changes:

background color of select lists

assets/js/customize-preview.js
```diff
@@ -1,49 +1,55 @@
...
-        result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec( hex );
-        return result ? {
-            r: parseInt( result[1], 16 ),
-            g: parseInt( result[2], 16 ),
-            b: parseInt( result[3], 16 )
-        } : null;
-    }
-
-    // Add listener for the "background_color" control.
+	/**
```
carolinan, line 3: Replaced spaces with tabs.

carolinan:
This is the search block (before) and a category block in the editor and the front.
I think the select list looks clunky, did I add to much padding?

Would the thick border stand out too much the editor?
block-forms


Search forms after:

comparison of differnt search forms

search form over a pink background

carolinan:
Because the styles for the border radius is inline, I was not able to adjust the dotted line for the outlined button when it is selected:

button-border-radius-editor

assets/sass/05-blocks/button/_editor.scss
```diff
@@ -7,21 +7,47 @@
 	line-height: var(--button--line-height);
 	background-color: var(--button--color-background);
 	border-radius: var(--button--border-radius);
+	border: var(--button--border-width) solid var(--button--color-background);
 	padding: var(--button--padding-vertical) var(--button--padding-horizontal);
 
-	&:hover,
-	&:focus,
-	&.has-focus {
+	&:visited {
```
jffng, line 13: Did you try using `%extend button-style` here and migrating as much as possible to that mixin for ease of maintenance? I wonder why we didn't do that before 🤔

assets/sass/04-elements/forms.scss
```diff
@@ -5,38 +5,228 @@ input[type="password"],
...
+	background: var(--global--color-white);
 }
 
 select {
 	border: var(--form--border-width) solid var(--form--border-color);
+	padding: calc(0.5 * var(--form--spacing-unit));
+	color: var(--form--color-text);
 }
 
 textarea {
```
jffng, line 66: Do you think we should also give `input[type="text"]` a default width of `100%`?

assets/sass/05-blocks/button/_editor.scss
jffng, line 13: What about using `%extend button-style` here and migrating as much as possible to that mixin for ease of maintenance? I wonder why we didn't do that before 🤔

jffng (commented):
This is looking good in my testing, I left a couple minor comments.

> Because the styles for the border radius is inline, I was not able to adjust the dotted line for the outlined button when it is selected

I'm not sure what a good option here is if we want to preserve the focus styles having an outline inside the button. Is there a way we could set the `outline-offset` when a custom border radius is set so that it looks like this? 

Screen Shot 2020-09-28 at 12 01 19 PM

carolinan:
> Is there a way we could set the `outline-offset` when a custom border radius is set so that it looks like this?

There is no CSS class added when the border radius is changed. I don't know how to check if the style attribute contains a radius.

assets/sass/04-elements/forms.scss
carolinan: No, because there might be forms with two text inputs on the same line?

melchoyce:
Thanks for adding that testing data @carolinan, it's super helpful!

Overall things are looking great 👍 

Noticed the color picker isn't showing the currently selected color, it looks like it's cut off a little:

image

> I think the select list looks clunky, did I add to much padding?

I think it could use a little more padding — I'd go with `var(--form--spacing-unit)` instead:

image

I also noticed the arrow isn't respecting the padding you've set on the element, so the balance appears off. I'm looking in Chrome:

image

If the arrowed adhered to that padding, it'd probably be fine.

melchoyce:
Also: can we punt the button issue to a new PR, since it seems like a more complicated problem? That way we can get the rest of this PR merged, since it's really close.

aristath:
> There is no CSS class added when the border radius is changed

I haven't checked, but perhaps this would work as a CSS selector: `div[style*='radius']`

carolinan:
Select list before and after:
selectlist-updated
I was not able to make the _fill_ work for the chevron icon from the Gutenberg icon library so a basic svg shape was used instead.

Buttons with border radius, when in focus (front) and when editing:

button-border-radius-focus

The file block download button style was updated to match a filled button.


For the color picker, I can see the selected color if I use Firefox, but it is only 1 pixel high. I can not see the selected color at all in chrome. It does not look anything at all like the examples here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color. I will test this next.

carolinan:
Color input element before and after:
color

melchoyce:
IMO this is good to merge, and we can make any follow-up changes in subsequent PRs.

PR #145, opened 2020-09-24
Small fixes

acosmin:
- The `display` argument will apply `esc_html` on the value, so it shouldn't be escaped. 
- `show-avatars` is a static string, it shouldn't be escaped.

### DIFF
--- a/template-parts/post/author-bio.php
+++ b/template-parts/post/author-bio.php
@@ -8,7 +8,7 @@
  */
 
 ?>
-<div class="author-bio <?php echo get_option( 'show_avatars' ) ? esc_attr( 'show-avatars' ) : ''; ?>">
+<div class="author-bio <?php echo get_option( 'show_avatars' ) ? 'show-avatars' : ''; ?>">
 	<?php echo get_avatar( get_the_author_meta( 'ID' ), '85' ); ?>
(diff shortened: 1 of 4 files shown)


template-parts/post/author-bio.php
```diff
@@ -8,7 +8,7 @@
  */
 
 ?>
-<div class="author-bio <?php echo get_option( 'show_avatars' ) ? esc_attr( 'show-avatars' ) : ''; ?>">
+<div class="author-bio <?php echo get_option( 'show_avatars' ) ? 'show-avatars' : ''; ?>">
```
carolinan, line 11: Good catch, mind fixing https://github.com/WordPress/twentytwentyone/blob/trunk/comments.php#L25 too?

acosmin: Sure, I'll do it first thing tomorrow :) mobile right now

acosmin: Done

acosmin:
@carolinan there are also these two lines :) I think they need to be updated: https://github.com/WordPress/twentytwentyone/blob/9c22fe37620c7bbb064b7336f2a1d5c94c34ed6e/classes/class-twenty-twenty-one-customize-color-control.php#L6-L7

PR #146, opened 2020-09-24
Specify grid column start value for tags links.

kellylawrence:
This should place the tags links container in the last grid column when an edit link isn't being output.

An alternative idea might be to always output a `<span>` or `<div>` where the edit link would go (even if a link wasn't actually output), so the tags links would automatically be pushed into the last grid column/row.

WordPress.org username is [kellylawrence](https://profiles.wordpress.org/kellylawrence/).


carolinan:
Looks good to me, tested in various browsers, thank you.

PR #140, opened 2020-09-24
Add the "Two Images Showcase" block pattern

kjellr:
Addresses the first checkbox for #50. 

As part of implementation, this adds a "Image Frame" block style. It also adds the images to the  `assets/images/` directory, along with license information in `style.css`. 

## Screenshots

Screen Shot 2020-09-24 at 11 29 14 AM

Screen Shot 2020-09-24 at 11 09 46 AM

Issue #50, opened 2020-09-17
Add initial block patterns


melchoyce:
Twenty Twenty One will come packaged with several block patterns. Here's a couple to get us started.

---

- [x] **Two images showcase**

![image](https://user-images.githubusercontent.com/2846578/93505065-5f1c1b00-f8e8-11ea-8907-6f3754d54697.png)

- Full-width block
- Maybe a media and text block, where instead of text, we show a second, smaller, framed image?

---

- [x] **Overlapping gallery**

![image](https://user-images.githubusercontent.com/2846578/93506538-66442880-f8ea-11ea-8592-61172699cd19.png)

- Wide-width block
- Maybe every third image overlaps every second image? Or horizontal images always overlap vertical images? We'll need to play with this in code to determine the best approach.

---

- [ ] **Image underlapping text box**

![image](https://user-images.githubusercontent.com/2846578/93506554-70febd80-f8ea-11ea-9562-0ce3205904ca.png)

- Wide-width block
- Image block
- Overlapped by a paragraph in a container with a border

---

- [x] **Media & Text article title**

![image](https://user-images.githubusercontent.com/2846578/93506644-8c69c880-f8ea-11ea-8720-3e8358d645c0.png)

- Width-width block
- Media and text block with a border. The text includes an H2, short separator block, and a small paragraph

---

- [x] **Portfolio list**

![image](https://user-images.githubusercontent.com/2846578/93506798-c2a74800-f8ea-11ea-9994-8a44ba402e33.png)

- Width-width block
- Separator, H2, tiny image

---

- [x] **Large text**

![image](https://user-images.githubusercontent.com/2846578/93506885-e8345180-f8ea-11ea-8f98-16e470d9020c.png)

- 144px paragraph block, 300 font weight

---

- [ ] **Two overlapping blocks of text**

![image](https://user-images.githubusercontent.com/2846578/93506927-f5e9d700-f8ea-11ea-925b-066fbc741ec0.png)

- Wide-width block
- Extra large paragraph in the first block, large paragraph in the second

---

- [x] **Link CTA**

![image](https://user-images.githubusercontent.com/2846578/93506975-0732e380-f8eb-11ea-8ae1-e61ebf621b0b.png)

- Width-width block
- Group or cover block with a border
- 96px paragraph block, 300 font weight
- Spacer block?
- Columns block
 - Left column: social media links
 - Right column: email link

-----

[...]

PR #150, opened 2020-09-25
Add "Bordered" style and "Media & Text article title" block pattern

kjellr:
This PR tackles another one of the block patterns from #50: 

**Media & Text article title** 

In doing so, it also adds a new "Bordered" block style for the media and text block. I extended this for use in the Group and Cover blocks too, since those will come in handy for other block patterns. #140 added a similar style for the Image block, but it mandates some padding. For completeness, I added this plain, unpadded border style to the image block too. 

## Screenshots 

Block Pattern

Screen Shot 2020-09-25 at 12 42 26 PM

Block Styles

gutenberg test_wp-admin_post php_post=1 action=edit


inc/block-patterns.php
```diff
@@ -36,4 +36,14 @@
...
+			'content'       => '<!-- wp:media-text {"mediaId":1752,"mediaLink":"' . esc_url( get_template_directory_uri() ) . '/assets/images/the_poplars_at_saint_remy.jpg","mediaType":"image","className":"is-style-twentytwentyone-border"} --><div class="wp-block-media-text alignwide is-stacked-on-mobile is-style-twentytwentyone-border"><figure class="wp-block-media-text__media"><img src="' . esc_url( get_template_directory_uri() ) . '/assets/images/the_poplars_at_saint_remy.jpg" alt="' . esc_attr__( '&quot;The Poplars at Saint-Rémy&quot; by Vincent Van Gogh (1889)', 'twentytwentyone' ) . '" class="wp-image-1752 size-full"/></figure><div class="wp-block-media-text__content"><!-- wp:heading {"align":"center"} --><h2 class="has-text-align-center">' . wp_kses_post( __( 'The Poplars at<br>Saint-Rémy', 'twentytwentyone' ) ) . '</h2><!-- /wp:heading --><!-- wp:separator --><hr class="wp-block-separator"/><!-- /wp:separator --><!-- wp:paragraph {"align":"center","fontSize":"small"} --><p class="has-text-align-center has-small-font-size">' . wp_kses_post( __( 'Vincent van Gogh<br>(Dutch, 1853-1890)', 'twentytwentyone' ) ) . '</p><!-- /wp:paragraph --></div></div><!-- /wp:media-text -->',
```
kjellr:
```
wp_kses_post( __( 'The Poplars at<br>Saint-Rémy', 'twentytwentyone' ) )
```

^ Is that the right way to escape + translate this? It's just normal post content.

melchoyce:
Can we add some padding to the image caption here, or should that be a new PR?

melchoyce:
Yeah any padding we could add here would help:

image

(Group block)

kjellr:
@melchoyce Should be all set in b4e6027. 

Screen Shot 2020-09-25 at 3 56 49 PM

I noticed some weirdness with Group block spacing in the front end too (the container had some extra padding by default), so I took that out in the same commit: 

Before (Editor)|Before (Front end)|After (Front end)
---|---|---
Screen Shot 2020-09-25 at 3 58 58 PM|Screen Shot 2020-09-25 at 3 58 50 PM|Screen Shot 2020-09-25 at 4 00 09 PM

inc/block-patterns.php
carolinan: Yes this works well, since there is HTML inside of it.

carolinan:
When I insert the media and text block pattern, there is a block validation error in the editor.
But If I save and refresh the page (editor), the error is gone.

(Gutenberg is not active)

kjellr:
@carolinan Mind giving this another try? I recreated the pattern without the plugin active, and now it seems to work fine in both contexts.

melchoyce (approved):
LGTM from the design side 👍

carolinan:
The pattern works now, but the border color has reverted back to black.

kjellr:
> The pattern works now, but the border color has reverted back to black.

Just needed to be synced up with `trunk`. 🙂

Thanks, all!

PR #151, opened 2020-09-25
Force #88: Use aria-current attribute on footer homepage link when home is the current page

### LINKED ISSUE #87: Use aria-current attribute on footer homepage link when home is the current page

jffng:
A duplicate of #88 because there is an unknown conflict with that PR.


footer.php
```diff
@@ -24,11 +24,14 @@
...
-					?>
-					<a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a>
-					<?php
-				}
+				if ( ! empty( $blog_info ) ) :
+					if ( is_front_page() || ( is_front_page() && is_home() ) ) :
+						?>
+						<?php bloginfo( 'name' ); ?>
+					<?php else : ?>
+						<a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"<?php echo $aria_current; ?>><?php bloginfo( 'name' ); ?></a>
```
jffng: Looks like `$aria_current` needs to be escaped.

audrasjb (changes requested):
Here is the full code of my failed pull request, FWIW:

```
			<div class="site-name">
				<?php if ( has_custom_logo() ) : ?>
					<?php if ( is_front_page() ) : ?>
						<?php the_custom_logo(); ?>
					<?php else : ?>
						<a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php the_custom_logo(); ?></a>
					<?php endif; ?>
				<?php else :
					$blog_info = get_bloginfo( 'name' );
					if ( ! empty( $blog_info ) ) :
						if ( is_front_page() ) : ?>
							<?php bloginfo( 'name' ); ?>
						<?php else : ?>
							<a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a>
						<?php endif;
					endif;
				endif;
				?>
			</div><!-- .site-name -->
```

footer.php
audrasjb: yep!

audrasjb:
Thank you for the new PR @jffng !

footer.php
audrasjb: ah wait, we don't need this $aria_current here

jffng:
Sure thing @audrasjb! Is this good to go from your end?

audrasjb (approved):
Looks good to me, thanks!

footer.php
```diff
@@ -22,14 +22,26 @@
...
 			<div class="site-name">
-				<?php
-				$blog_info = get_bloginfo( 'name' );
-				if ( ! empty( $blog_info ) ) {
-					?>
-					<a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a>
+				<?php if ( has_custom_logo() ) : ?>
+					<?php if ( is_front_page() ) : ?>
+						<?php the_custom_logo(); ?>
+					<?php else : ?>
```
carolinan: Is this needed? The custom logo already has the theme support for unlinking it on the home page, has it been tested?
See https://github.com/WordPress/twentytwentyone/blob/trunk/functions.php#L91

audrasjb: yeah I tested it but I didn't notice the double link… 🤦 

Capture d’écran 2020-09-26 à 08 32 41

carolinan: Oops

jffng: Apologies, I'm not quite following — what needs to happen here? I understand that the double link should be removed. But is the idea to only show the custom logo if it's the front page?

carolinan: The idea is to show the logo if there is a logo. Show the title if there is no logo.

Neither the title or the logo should be linked on the front page, but linked on the **paged** blog pages.

But the logo already has a custom link built in, and it does not need to be altered because there is already theme support for unlinking the logo.

jffng:
I think https://github.com/WordPress/twentytwentyone/pull/151/commits/9016587f3812bc0d8318b5442108311861c1c40a fixes things. @carolinan @audrasjb would you mind giving this another review? 🙏

carolinan:
This works well. We will need to adjust the footer logo size.

PR #152, opened 2020-09-25
Update border color to dark grey

melchoyce:
Instead of black, borders should use the dark grey color.


kjellr (approved):
LGTM. Thanks @melchoyce!

carolinan:
<del>To be able to change the color together with the customizer option, the primary color would be needed, because it is the primary and not the dark grey directly that is used in the code for the color option.</del>

Ignore that, let's instead add the border color to the customizer color option code, and try to change the opacity only for this variable.

PR #155, opened 2020-09-26
Lower border and divider color opacity on dark backgrounds

### LINKED ISSUE #149: The dividers does not change color when the site background color is changed.
**Describe the bug**
When the background color is changed with the option in the customizer, the dividers do not change color.

**To Reproduce**
1. Go to the customizer
2. Click on the color option
3. Select a black background
4. See error, the dividers are #000 like the background.

**Expected behavior**
I expect the dividers colors to be adjusted.

The easiest solution would be to change the dividers from #000 to the primary or the secondary color, that is already 
used in the code that adjust the contrast.
See 
https://github.com/WordPress/twentytwentyone/blob/trunk/assets/sass/01-settings/global.scss#L86

Border with the current black color to the left, compared to the primary color on the right:
color-comparison

Examples where the color has been changed to white:

color-example
color-example2

Considerations: 
During testing I found a white divider on an all black background very _sharp_, and perhaps a different opacity can be tested here.

carolinan:
Closes https://github.com/WordPress/twentytwentyone/issues/149

If the luminance is 40 or less, reduce the opacity of the border color.

darkgrey


melchoyce:
Does this solve your comment in https://github.com/WordPress/twentytwentyone/pull/152#issuecomment-699285496, or should I make a follow-up issue for that?

carolinan:
> Does this solve your comment in [#152 (comment)](https://github.com/WordPress/twentytwentyone/pull/152#issuecomment-699285496), or should I make a follow-up issue for that?

This solves it.

kjellr (commented):
The PR itself works fine in my testing. I'm not 100% sure it's necessary though — in the default color scheme, the borders are the same color as the text, so this changes the overall feel a little bit. 

Here are a couple screenshots where this kicks in, compared to slightly lighter versions of the background colors, where it does not kick in. 

Lighter Borders|No Change
---|---
Screen Shot 2020-09-29 at 1 19 47 PM|Screen Shot 2020-09-29 at 1 20 04 PM

Lighter Borders|No Change
---|---
Screen Shot 2020-09-29 at 1 29 10 PM|Screen Shot 2020-09-29 at 1 20 41 PM

In both examples, I think the version where the borders are the same color as the text carries the theme's default aesthetic a little more successfully. But I'll defer to @melchoyce for her decision here.

melchoyce:
Are you good with how stark the contrast feels, @kjellr? The dividers feel much bolder to me when light-on-dark.

kjellr:
> Are you good with how stark the contrast feels, @kjellr? The dividers feel much bolder to me when light-on-dark.

Yeah, I am ok with the stark contrast. The theme elements are so minimal to begin with, I think it makes sense for them to not shy away from little hints of boldness.

melchoyce:
Okay, thanks. I'm waffling. @carolinan, any preference here?

carolinan:
They are distracting to me, especially when there is little content, but I am also able to adjust them with CSS myself.
Lets keep the opacity but make sure that the color does change when the background changes.

PR #156, opened 2020-09-26
Submenu background color update on Customizer preview when background color change

### LINKED ISSUE #128: Submenu background color doesn't update in Customizer preview when background color is changed
## Description
The background color used by submenus in the navigation doesn't update in the Customizer preview when the background color setting is modified. For example, if the background color was pastel green and you try setting it to black, submenus will look like this while you're in the Customizer:

image

Once you save and refresh, it updates, and of course the final output on the front-end works as expected. But the appearance in the Customizer should automatically update as you change the background color option.

## Reproduction steps
1. If you haven't already, set the Primary Navigation to a menu that includes a submenu.
2. Open the Customizer.
3. Go to the Colors section.
3. Set a new background color.
4. Hover over the submenu while in the Customizer.

kishanjasani:
Issue Fix: #128 
Update submenu color when background color is change.

### DIFF
--- a/assets/sass/06-components/navigation.scss
+++ b/assets/sass/06-components/navigation.scss
@@ -234,7 +234,7 @@
 
 			@include media(mobile) {
 				margin: 0;
-				background: var(--global--color-background);
+				background: transparent;
 				left: -15px;
 				top: calc(100% - (0.5 * var(--primary-nav--padding)));
 				min-width: max-content;
--- a/style.css
+++ b/style.css
@@ -3533,7 +3533,7 @@ h2.page-title {
 @media only screen and (min-width: 482px) {
 	.primary-navigation > div > .menu-wrapper > li > .sub-menu {
 		margin: 0;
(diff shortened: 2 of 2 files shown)


carolinan:
Hi, please see this link that shows the design of the sub menus, with a solid background color:
https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=487%3A67

What are your thoughts on this solution, where the updated background color in the customizer is applied to all elements that uses the `--global--color-background` variable:  https://github.com/WordPress/twentytwentyone/pull/142/files#diff-c48d4a9be3cdf1755cb64b27c7fd69c6R48

kishanjasani:
[This PR](https://github.com/WordPress/twentytwentyone/pull/142/files#diff-c48d4a9be3cdf1755cb64b27c7fd69c6R48) is really cool:ok_hand: . Let's go with that. You can close this PR.

PR #161, opened 2020-09-28
Simplify the implementation of octed hex colors for borders a bit

aristath:
Simplify and make the use of octet hex colors consistent for border-color.
This is an addition to https://github.com/WordPress/twentytwentyone/pull/155
`#RRGGBBAA` colors are [supported by all browsers that can use css-variables](https://caniuse.com/css-rrggbbaa) so there is no problem in using them consistently.


carolinan:
I am closing this since the opacity will not be changed.

PR #162, opened 2020-09-28
🐛 FIX: Align classic editor images correctly (#160)

### LINKED ISSUE #160: Classic: Images are not correctly aligned
**Describe the bug**
This bug was sent to me via direct message on slack.
Images in the classic block or created with the classic editor are not centered.

**To Reproduce**
1. Import the **[theme unit test](https://github.com/WPTT/theme-unit-test)**
2. On the front, open the post called **Markup: Image Alignment**
3. See error
unit-test-image

nielslange:
Fixes #160 

<table>
<tr>
<td>Before:
<br><br>

#160-before
</td>
<td>After:
<br><br>

#160-after
</td>
</tr>
</table>

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -226,6 +226,11 @@ img {
 	max-width: 100%;
 }
 
+/* Classic editor images */
+.entry-content img {
+	max-width: 100%;
+}
+
 /* Make sure embeds and iframes fit their containers. */
 embed,
 iframe,
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -1749,6 +1749,11 @@ img {
 	max-width: 100%;
 }
 
+/* Classic editor images */
+.entry-content img {
+	max-width: 100%;
+}
+
 /* Make sure embeds and iframes fit their containers. */
 embed,
 iframe,
@@ -5262,10 +5267,6 @@ main * {
 
 footer * {
 	max-width: unset;
-}
-
-main img {
-	max-width: 100%;
 }
 
 html,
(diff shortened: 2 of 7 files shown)


carolinan:
This works, but I am confused about why the IE style is not updated?

nielslange:
> This works, but I am confused about why the IE style is not updated?

I just ran `npm run build` again and now I'm seeing the changes in `ie.css` as well. I might have missed that entry before.

kjellr (changes requested):
This works just fine, but I have just a couple organizational notes: 

If the goal is to ensure we're targeting classic editor images, we should probably narrow the scope of this rule a little bit. Maybe we target something like the `entry-content` class, rather than the `main` element? And we do that, I'd recommend placing the rule either in the `04-elements/media.scss` or `06-components/entry.scss` files instead.

nielslange:
> This works just fine, but I have just a couple organizational notes:
> 
> If the goal is to ensure we're targeting classic editor images, we should probably narrow the scope of this rule a little bit. Maybe we target something like the `entry-content` class, rather than the `main` element? And we do that, I'd recommend placing the rule either in the `04-elements/media.scss` or `06-components/entry.scss` files instead.

Thanks for your review, @kjellr.  As I'm not very familiar with the underlying SCSS structure of the Seedlet theme, which file would you recommend `04-elements/media.scss` or `06-components/entry.scss`? Given that the change affects images, I tend to move the changes to `04-elements/media.scss`.

kjellr:
Yeah, I think `media.scss` makes the most sense. 👍 Thanks @nielslange!

nielslange:
> Yeah, I think `media.scss` makes the most sense. 👍 Thanks @nielslange!

Thank you for your prompt reply, @kjellr. I moved the code as discussed. Feel free to run another review.

kjellr (approved):
Thanks!

PR #165, opened 2020-09-28
Revise link styles

### LINKED ISSUE #110: Update link states to match the design

jffng:
This PR picks up on the work started in #133 to address #110. 

Kapture 2020-09-28 at 13 21 35

(Apologies for another PR, but I chatted with @fabiankaegy and figured this was the fastest as I can't commit to his PR.)


assets/sass/03-generic/reset.scss
```diff
@@ -109,13 +97,6 @@ a {
 	}
 }
 
-// Focus styles
```
jffng, line 112: I removed the focus styles because they don't match the design.

melchoyce:
Looking good! 

Noticed that focus styles get cut off when a link is flush against the edge of the container: 
image
image

(Screenshots in Chrome)

kjellr:
> Noticed that focus styles get cut off when a link is flush against the edge of the container

I think the footer needs to have `overflow: visible;` set in order to get that working. Should be fine — I did so for testing and didn't notice anything weird. 

In Safari, I also noticed that multi-line titles look a little weird: 

Screen Shot 2020-09-29 at 1 08 06 PM

I think that's due the line height, but I definitely wouldn't adjust the line height just to fix that. It's weird that the browser doesn't fill that gap automatically.

jffng:
> I think the footer needs to have overflow: visible; set in order to get that working. Should be fine — I did so for testing and didn't notice anything weird.

Nice — I removed the overflow rule entirely and it seems to have done the trick. @melchoyce can you confirm this works for the use cases you found those errors? 

> In Safari, I also noticed that multi-line titles look a little weird.

This is ... unfortunate. AFAIK using outlines for focus styles is a good practice, so not sure there's a better approach that can solve for that edge case.

kjellr (approved):
> Nice — I removed the overflow rule entirely and it seems to have done the trick.

I can confirm that this takes care of it for me. 

> AFAIK using outlines for focus styles is a good practice, so not sure there's a better approach that can solve for that edge case.

Agreed. I think it's fine to merge without addressing that, but it might be worth opening up another issue to track this one in case someone knows of a solution.

I think this is fine to merge as is. Form elements do not have the correct focus styles, but we can tackle that in a followup PR: 

Screen Shot 2020-09-30 at 7 57 13 AM

marctison75:
firefox


In addition to multi-line titles that look weird on Safari, they also don’t look good on Firefox (Windows). In this example, you can see how the outline can interact with the shape of the letter and decrease the readability of the text.

> AFAIK using outlines for focus styles is a good practice, so not sure there's a better approach that can solve for that edge case.

Why don’t the theme use the default focus style set by the browsers (user agent stylesheet), which is already an outline? (https://www.w3.org/ is using it).

jffng:
> Why don’t the theme use the default focus style set by the browsers (user agent stylesheet), which is already an outline? (https://www.w3.org/ is using it).

Not a bad suggestion — feel free to raise it in a follow on PR or new issue, thanks @marcduluc.

Issue #117, opened 2020-09-22
Add social media icon styling to menu widgets


melchoyce:
**Is your feature request related to a problem? Please describe.**
I often want to add social menus to my site. This can be achieved using the beta widget-block screen, but it would also be nice to support social icons inside menu widgets.

**Describe the solution you'd like**
If you link to a social media account within your menu widget, the link is transformed into a social media icon.

**Describe alternatives you've considered**
Only supporting the social icon block, or not including support for social menus — but neither is ideal, IMO.

**Additional context**
Maybe we could reuse the SVGs and styles packaged with the Gutenberg social icon block?

This is a low-priority task.

carolinan:
Is it a navigation menu widget, or an actual navigation menu, located in the footer, that you assign a navigation menu to? 
If the latter, does it replace one of the grids in the footer widget area?

I have no experience of changing the **_navigation menu widget_**. I do not know how it works.

But for a navigation menu that allows both social and regular links, we can just copy back the menu that was already in Seedlet or one of the default themes, and remove the placeholder "link / chain" icon so that regular text is shown if there is no match to a social media site.

melchoyce:
I was thinking it would be a navigation menu widget (and eventually a navigation block), so there's more versatility around placement.

beafialho:
I think this would be a valuable visual addition.

carolinan:
I do not understand the difference between adding a paragraph with two links and a social icon block in a widget area, which is already possible in 5.6?

How is the custom widget offering more flexibility? Both needs the widget area. 
The menu widget has 0 block style options - it is more limited that way. 

Not arguing, trying to understand what needs to be coded. Or if we just need to document how the user can add their social links.

melchoyce:
It sounds like there isn't any guarantee widget blocks will launch with 5.6, unfortunately 😕 That would solve a lot of problems. 

My general thoughts around this:
- We style the navigation menu widget similarly to the primary menu. Horizontal, not vertical.
- If someone includes a link to a popular social media site, we target that link with CSS and change from a label to an icon, similarly to how a lot of themes style social menus.

Let me know if I can clarify more, I hope I'm making sense 😅

melchoyce:
[...]

PR #170, opened 2020-09-29
Don't load navigation script if no menu

LittleBigThing:
The JS for the primary navigation doesn't need load if no navigation menu is set.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -334,7 +334,9 @@ function twenty_twenty_one_scripts() {
 	}
 
 	// Main navigation scripts.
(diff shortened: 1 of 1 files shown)


LittleBigThing:
Maybe the included `inc/menu-functions.php` shouldn't be loaded either in this case?

carolinan:
I think we should await a decision regarding a social menu before making changes to this, or we will just end up redoing it again? See https://github.com/WordPress/twentytwentyone/issues/117

carolinan:
This JS file would not affect the _menu widget_, so lets merge this.

Issue #74, opened 2020-09-19
Testing and adjusting default WP pages


luminuu:
We need to check that the default pages provided by core are matching the design and adjust them if needed.

- [ ] 404 page
- [x] No content
- [ ] Attachment (Seedlet theme provided an image.php file)

PR #172, opened 2020-09-29
Update the image attachment page

### LINKED ISSUE #79: Accessibility: WCAG 2.4.5 Multiple Ways Level AA
More than one way is available to locate a Web page within a set of Web pages except where the Web Page is the result of, or a step in, a process.

Reference: https://www.w3.org/TR/WCAG21/#multiple-ways

- [x] Menu **-Optional**
- [ ] Next and previous post navigation, links between paginated posts (page breaks), link to parent on attachment page (image.php)
- [x] Search  **Optional with widget or block**

@audrasjb do I understand the criteria correctly that this will fail because both the menu and search is optional?
There is no fall back page menu, so if no menu is set, there is no menu at all.

Even though 5.5 added a sitemap, it is not a solution for this guideline since it is not really human readable nor helpful for screen reader users, since the link name is just the file name and the slug.

carolinan:
Partial fix for 
https://github.com/WordPress/twentytwentyone/issues/74

Closes https://github.com/WordPress/twentytwentyone/issues/79

Updates the header, image alignment, and image meta data including a link to the parent post if one exists.

Before:
image-php-before

After: 
image-php-after


melchoyce:
For some reasons it's not showing the divider lines for me:

image

carolinan:
Cache?

kjellr (approved):
Works well in my testing. I'm able to see the divider line. 👍

melchoyce:
Clearing cache doesn't fix it, but it's got to be somewhere weird on my end if no one else can reproduce. Screenshots look good 👍

PR #173, opened 2020-09-29
Update single page header treatment

### LINKED ISSUE #37: Pages - How to implement two page heading styles

kjellr:
Should close #37 by ensuring that single pages use the same header treatment that's currently present on single posts. 

Before: 
Screen Shot 2020-09-29 at 2 11 57 PM

After: 
Screen Shot 2020-09-29 at 2 11 42 PM


kjellr:
Question for @melchoyce: Should we replicate this header treatment in the editor too? 

Current|Proposed
---|---
Screen Shot 2020-09-29 at 2 16 16 PM|Screen Shot 2020-09-29 at 2 15 41 PM

melchoyce:
I think yes, though could we add some more padding in the editor stylesheet to make sure there's adequate space on either side, even if the editor viewport is narrow?

kjellr:
> I think yes

I've pushed this change — give it a try and let me know what you think. (TBH, it feels _a little_ weird to me, but it may just be that I need to sit with this new header treatment everywhere). 

> could we add some more padding in the editor stylesheet to make sure there's adequate space on either side, even if the editor viewport is narrow?

Looks like that's only a problem when the sidebar is open. It _does_ retain the correct padding when the sidebar is closed: 

Screen Shot 2020-09-29 at 3 05 02 PM

In any case, this is an issue with _all_ wide elements in the theme, not just the title, so I think it could be punted to another PR.

melchoyce (approved):
I need to get used to it, too, so let's merge and sit with it for a little bit. PR looks good to me 👍

carolinan:
Single pages look good to me, but the border under the page title is also visible in the search results.
Perhaps use a different class than `page`, one that is not a "post class"?

page-header-border

kjellr:
> Single pages look good to me, but the border under the page title is also visible in the search results.
Perhaps use a different class than page, one that is not a "post class"?

@carolinan f82f9de changes that to `singular`, which solves the problem in my testing.

carolinan:
tested the update, LGTM

kjellr:
Thank you!

PR #174, opened 2020-09-29
Add Link CTA block pattern

kjellr:
Implements the "Link CTA" pattern from #50. 

- The spacing in the editor looks just a little weird, but that's because of the default block padding there. It looks right in the front end. 🤷  
- The font-weight for the main title isn't correct. Techniques to achieve the lighter weight here [are being discussed here](https://github.com/WordPress/twentytwentyone/issues/50#issuecomment-698555348), but I don't think that should hold up this PR in the meantime. 

**Editor**

Screen Shot 2020-09-29 at 2 47 45 PM

**Front End**

Screen Shot 2020-09-29 at 2 41 12 PM


melchoyce (approved):
LGTM 👍

jffng (commented):
The mobile view could use some consideration: 

Screen Shot 2020-09-29 at 1 17 49 PM

Is there a way to implement this that makes use of the theme's responsive type styles?

melchoyce:
We might want to address these on the theme side — for example, I think we'll likely want to style the columns block so everything aligns well on mobile, and implement some responsive type styles for headings so it scales down instead of breaking. Maybe progressively enhance using `clamp()`? https://css-tricks.com/simplified-fluid-typography/

inc/block-patterns.php
```diff
@@ -45,4 +45,14 @@
...
...tyle="font-size:96px;line-height:1.2">' . wp_kses_post( __( 'Let&#8217;s Connect.', 'twentytwentyone' ) ) . '</p><!-- /wp:paragraph --><!-- wp:spacer {"height":75} --><div style="height:75px" aria-hidden="true" class="wp-block-spacer"></div><!-- /wp:spacer --><!-- wp:columns --><div class="wp-block-columns"><!-- wp:column --><div class="wp-block-column"><!-- wp:paragraph --><p><a href="#" data-type="URL">' . wp_kses_post( __( 'Twitter', 'twentytwentyone' ) ) . '</a> / <a href="#" data-type="URL">' . wp_kses_post( __( 'Instagram', 'twentytwentyone' ) ) . '</a> / <a href="#" data-type="URL">' . wp_kses_post( __( 'Dribbble', 'twentytwentyone' ) ) . '</a></p><!-- /wp:paragraph --></div><!-- /wp:column --><!-- wp:column --><div class="wp-block-column"><!-- wp:paragraph {"align":"right"} --><p class="has-text-align-right"><a href="#">' . wp_kses_post( __( 'Contact@WordPress.org', 'twentytwentyone' ) ) . '</a></p><!-- /wp:paragraph --></div><!-- /wp:column --></div><!-- /wp:columns --><!-- wp:spacer {"height":20} --><div style="height:20px" aria-hidden="true" class="wp-block-spacer"></div><!-- /wp:spacer --></div></div><!-- /wp:cover --><!-- wp:paragraph --><p></p><!-- /wp:paragraph -->',
```
carolinan: Is this an actual email address? `Contact@WordPress.org` It is going to get spammed, I always recommend using example.com

```suggestion
			'content'       => '<!-- wp:cover {"overlayColor":"green","contentPosition":"center center","align":"wide","className":"is-style-twentytwentyone-border"} --><div class="wp-block-cover alignwide has-green-background-color has-background-dim is-style-twentytwentyone-border"><div class="wp-block-cover__inner-container"><!-- wp:spacer {"height":20} --><div style="height:20px" aria-hidden="true" class="wp-block-spacer"></div><!-- /wp:spacer --><!-- wp:paragraph {"style":{"typography":{"fontSize":96,"lineHeight":"1.2"}}} --><p style="font-size:96px;line-height:1.2">' . wp_kses_post( __( 'Let&#8217;s Connect.', 'twentytwentyone' ) ) . '</p><!-- /wp:paragraph --><!-- wp:spacer {"height":75} --><div style="height:75px" aria-hidden="true" class="wp-block-spacer"></div><!-- /wp:spacer --><!-- wp:columns --><div class="wp-block-columns"><!-- wp:column --><div class="wp-block-column"><!-- wp:paragraph --><p><a href="#" data-type="URL">' . wp_kses_post( __( 'Twitter', 'twentytwentyone' ) ) . '</a> / <a href="#" data-type="URL">' . wp_kses_post( __( 'Instagram', 'twentytwentyone' ) ) . '</a> / <a href="#" data-type="URL">' . wp_kses_post( __( 'Dribbble', 'twentytwentyone' ) ) . '</a></p><!-- /wp:paragraph --></div><!-- /wp:column --><!-- wp:column --><div class="wp-block-column"><!-- wp:paragraph {"align":"right"} --><p class="has-text-align-right"><a href="#">' . wp_kses_post( __( 'example@example.com', 'twentytwentyone' ) ) . '</a></p><!-- /wp:paragraph --></div><!-- /wp:column --></div><!-- /wp:columns --><!-- wp:spacer {"height":20} --><div style="height:20px" aria-hidden="true" class="wp-block-spacer"></div><!-- /wp:spacer --></div></div><!-- /wp:cover --><!-- wp:paragraph --><p></p><!-- /wp:paragraph -->',
```

inc/block-patterns.php
```diff
@@ -45,4 +45,14 @@
...
 		)
 	);
+
+	register_block_pattern(
+		'twentytwentyone/link-cta',
+		array(
+			'title'         => __( 'Link CTA', 'twentytwentyone' ),
```
carolinan: What is a CTA? I understand that you mean call to action, but can this title be improved?
Or a description added.

kjellr: Yeah, I just grabbed this from the issue. @melchoyce do you have a different suggestion?

inc/block-patterns.php
kjellr: Yeah, I grabbed that from the issue too. Might make sense to have something that's clearly fake? Not sure what a good option would be.

kjellr:
> Is there a way to implement this that makes use of the theme's responsive type styles?

I think it _would_ accept the responsive type styles if we were using a standard type size here. Instead, it's a hardcoded custom size. Maybe we need a "Gigantic/Largest" font size that matches this 96px size. @melchoyce, what do you think?

inc/block-patterns.php
carolinan: example@example.com

inc/block-patterns.php
kjellr: In https://github.com/WordPress/twentytwentyone/pull/174/commits/844eb51486c174fb73f6a83902cee763bf803c7a, I renamed it to "Links Area" for now.

inc/block-patterns.php
kjellr: Cool, that seems like a reasonable idea, [given the provenance behind that URL](https://en.wikipedia.org/wiki/Example.com). I switched to `example@example.com` in https://github.com/WordPress/twentytwentyone/pull/174/commits/cf1ad265ee9a02cbe300bf9618a280d76b41161d.

carolinan:
Does the text size name fit in the editor? =D

kjellr:
> Maybe we need a "Gigantic/Largest" font size that matches this 96px size.

I just pushed a small change that tackles this, and it works _mostly_ well. It doesn't work at all in the editor, because Gutenberg inlines the font size there. But I'm not sure there's anything we can do about that in the theme. 

The font size picker is also weird now, but I'm not sure if it's possible to fix that?

Screen Shot 2020-09-30 at 10 20 10 AM

kjellr:
I also pushed c0fe27b, which sets the font weight for the "Gigantic" preset so that it matches the heading's lighter weight. This is a sort of sneaky way to address [the issue outlined here](https://github.com/WordPress/twentytwentyone/issues/50#issuecomment-698555348), but it _sort of_ makes sense to me? After all, this new size is meant to mimic the page titles, and those are a thinner weight. Interested in opinions on this... we can always just wait and see if https://github.com/WordPress/gutenberg/pull/24978 lands in time instead.

kjellr:
Here's the current state on desktop: 

Screen Shot 2020-09-30 at 10 30 55 AM

And on mobile: 

Screen Shot 2020-09-30 at 10 31 23 AM

It does still trim oddly in some sizes: 

Screen Shot 2020-09-30 at 10 31 14 AM

... but I think we should address that separately in a global PR, since that effects [all large text on the site](https://cloudup.com/cAgXd0MtgJ0).

melchoyce:
Do we have any sort of standard for pattern names yet? Wondering how long is acceptable. "Connect Call to Action" could work?

kjellr:
> Do we have any sort of standard for pattern names yet? Wondering how long is acceptable. "Connect Call to Action" could work?

We don't yet, but I'm not sure that should hold up this PR. I think it might make sense to go through #50 and rename them all in a standardized way. Then we can do a sweep of any already-developed patterns and update them all.

melchoyce:
WFM

PR #179, opened 2020-09-30
Set the background color to the CSS variable in the customizer

### LINKED ISSUE #128: Submenu background color doesn't update in Customizer preview when background color is changed

carolinan:
Sets the color from the background color option to the `--global--color-background` CSS variable in the customizer live preview.
Also replaces spaces with tabs.

Fixes https://github.com/WordPress/twentytwentyone/issues/128

### DIFF
--- a/assets/js/customize-preview.js
+++ b/assets/js/customize-preview.js
@@ -1,49 +1,50 @@
 ( function( api, _ ) {
 
-    /**
-     * Get luminance from a HEX color.
-     *
-     * @param {string} hex - The hex color.
-     *
-     * @return {number}
-     */
-    function twentytwentyoneGetHexLum( hex ) {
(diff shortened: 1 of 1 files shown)


aristath (approved):
LGTM :+1:

PR #180, opened 2020-09-30
Keep focus inside the menu modal

carolinan:
This is an attempt to make sure that the focus is kept inside the menu modal when it is open.

Before:
When you reach the last menu item, you can no longer tab forwards.
When tabbing backwards, focus can move past the close button, all the way around the page, and the visible focus is lost since 
the focus is now on content that is hidden by the open menu.

After:
Focus is kept inside the menu modal, looped, until the menu is closed.


carolinan:
If the toggle button is simplified to one instead of two, this update should no longer be needed.

PR #181, opened 2020-09-30
Adjust widget title size and widget area spacing

### LINKED ISSUE #77: Testing and adjusting footer widget area
There's only one widget area given in Twenty Twenty One and each widget is added to the next available spot. It consists of three columns on wide screens.

As the new block based widget screen got out of experimental status in Gutenberg 8.9.0, it's most likely to land in 5.6, so we need to have this area working for both the old-school widgets and the blocks.

In my recent test the social icons block did not align right due to the given setup (float is added but as it's a flex item float doesn't work, it needs to be justify-content: flex-end). It's also not possible to style the icons with the dark grey color unless you add additional CSS via the customizer.

carolinan:
Reduces the font size for the widget title.
Reduces the spacing below the widget area to better match the design.

Closes https://github.com/WordPress/twentytwentyone/issues/77

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -5100,6 +5100,7 @@ h2.page-title {
 
 .widget-area {
 	font-size: 1rem;
+	margin-bottom: 0;
 }
 
 @media only screen and (min-width: 822px) {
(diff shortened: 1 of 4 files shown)


jffng (approved):
Looks good, thanks @carolinan !

PR #182, opened 2020-09-30
Use only one button to toggle the mobile menu

acosmin:


header.php
```diff
@@ -32,18 +32,16 @@
...
 							<span class="dropdown-icon open"><?php esc_html_e( 'Menu', 'twentytwentyone' ); ?>
 								<?php echo twenty_twenty_one_get_icon_svg( 'menu' ); // phpcs:ignore WordPress.Security.EscapeOutput ?>
 							</span>
-							<span class="hide-visually expanded-text"><?php esc_html_e( 'expanded', 'twentytwentyone' ); ?></span>
-						</button><!-- #primary-open-menu -->
-						<button id="primary-close-menu" class="button close" aria-haspopup="true" aria-controls="primary-navigation" aria-expanded="true">
 							<span class="dropdown-icon close"><?php esc_html_e( 'Close', 'twentytwentyone' ); ?>
 								<?php echo twenty_twenty_one_get_icon_svg( 'close' ); // phpcs:ignore WordPress.Security.EscapeOutput ?>
 							</span>
+							<span class="hide-visually expanded-text"><?php esc_html_e( 'expanded', 'twentytwentyone' ); ?></span>
```
carolinan: @aristath Should we not use aria-expanded instead of the expanded and collapsed text? But this can be a separate PR.

carolinan:
**_Not yet tested with a screen reader_**

This works well and has the "side effect" that the focus is once again kept inside the menu modal.

The position of the close button needs to be adjusted but there are several elements like the site title and logo that also needs to be inside this area so it does not need to be perfect in this PR.

header.php
aristath: Yes, using `aria-expanded` on the `button` element would be preferable... And if I'm not mistaken the "expanded" text won't be needed, as it is announced by screen-readers :+1:

carolinan:
Did a one minute test with the NVDA screen reader and the "expanded" text is read together with the word menu. When the menu is closed it reads MENUExpanded
So it is both without pause in between, and the incorrect text.

carolinan:
This works well with the screen reader, the aria expanded is announced correctly for the mobile menu toggle.

PR #192, opened 2020-09-30
Add the large text block pattern

jffng:
This PR starts to add the Large Text block pattern, as a part of #50. 

To-do:
- [x] Solve for font-weight
- [x] Refine column widths
- [x] Make it translation-ready

Kapture 2020-10-01 at 17 14 13


assets/sass/05-blocks/columns/_editor.scss
```diff
@@ -0,0 +1,6 @@
+.wp-block-columns .wp-block,
+.wp-block-columns .wp-block-column {
+	// Allow the width of a block that lives inside the columns block to be
+	// determined by the parent columns block
+	max-width: none;
```
jffng: I'm not sure about this. Should the width of a block inside a columns block be unconstrained? 

It overrides this rule below which is quite broad — limiting the width of any block (included nested blocks) in the editor to the default width, unless it has a wide or full alignment:

https://github.com/WordPress/twentytwentyone/blob/9e4f5213502ad6007c2c5267b8ad08ed4c5e53d7/assets/sass/05-blocks/utilities/_editor.scss#L139-L143

melchoyce: We might want to see if anything happens with https://github.com/WordPress/gutenberg/issues/24055 or https://github.com/WordPress/gutenberg/issues/25655. Alternately, @mtias, do you have any thoughts here?

kjellr: Yeah, I feel like we should wait for a natively-supported solution to this problem. Otherwise when folks switch themes, their text will end up looking really narrow all of a sudden, while the rest of the page stays the same.

aristath:
Pushed a commit merging master in this branch and resolving the merge conflicts.

assets/sass/05-blocks/columns/_editor.scss
mtias: Agreed on not coding around this issue and to improve the affordances we offer in Gutenberg as a whole. I'm also curious to know when you would use a single column block vs a group block set to wide?

carolinan: Can't we make it a block style for a heading?

carolinan:
Can't we make this a block style for a heading instead, to avoid delays about the columns or weights?

assets/sass/05-blocks/columns/_editor.scss
kjellr: I think the only use case for a single column (versus a Group block) would be to get the text to expand the full-width. But that's just a hack, really. I think if Gutenberg were to allow you to set a headings block to wide width, that would be the cleanest, simplest solution for now.

kjellr: ~FWIW, I don't think that https://github.com/WordPress/gutenberg/pull/25857 would solve this for us either, since we'd still need to override the default max-width imposed on all blocks.~

Nevermind. Looks like Gutenberg does provide a `max-width: none` on its own here, and the theme overrides it with the rule that @jffng noted above. 🙃 We should just set the new rule in this PR to `inherit`.

kjellr:
> Can't we make this a block style for a heading instead, to avoid delays about the columns or weights?

Technically yes, but I'd like to avoid coding around the issue if possible. I'll ask around in the Gutenberg repo to see if someone can help us prioritize built-in solutions. 

Even if one of those solutions does land though, I'm not totally sure that it'll work nicely in older GB versions. 🤔

kjellr:
I pushed a few updates, and I think this is at a good merge point for now. 

- I set the max-width of columns to `inherit` rather than the `none` that was originally in this PR. That allows them to just pull the `none` value from Gutenberg without us having to overwrite it. The fact that we were overwriting that in the first place feels like a bug, since it didn't match the front-end appearance. 
- I revised the bundled font-size names:
    - Huge => Extra Large
    - Gigantic => Huge
- I added a new "Gigantic" size to match this font size. This lets us specify a lighter weight for it like we did for the other "Gigantic" font size that came before this one. 

Aside from that font-weight thing (which we're already doing), I think this seems relatively hack-free. Once there's either a wide-width option for the heading block, or a one-column option available for the columns block, we can update the pattern to match.

melchoyce (approved):
LGTM 👍  Just needs a devreview.

jffng (commented):
I think the latest approach is sound. 

Noting the editor preview for these text sizes is cut off and will be unless patched in Gutenberg. I will open an issue.

Screen Shot 2020-10-06 at 1 57 07 PM

aristath (approved):

PR #196, opened 2020-09-30
Fix: Color contrast is too low in striped table blocks (#185)

### LINKED ISSUE #185: Color contrast is too low in striped table blocks.
**Describe the bug**
This is the only issue I could find when testing the current table block styles.

On dark backgrounds:
When a table block is added and the **striped** style is used, 
the text color is white over a light colored background and the color contrast is too low.
This happens both in the editor and the front.

**To Reproduce**
1. Go to the customizer and change the body background color to black
2. Create or open a post or page.
3. Add a table block with some content.
4. Select the striped block style
5. See that the text color is white and difficult to read.

The table block does not have a block sidebar option to change the text color, but can be changed inline.

**Screenshots**
table-front

2ndkauboy:
The current implementation of the swapped colors on a dark mode made it a bit hard to implement the fix. I've tried to do it as closely to the current code as possible. But maybe functions like `custom_get_readable_color` could be rewritten a bit, so they don't return a color hex value but a keyword such as `light` or `dark` and the functions calling them can then set a color based on that result. This would make the code a bit easier to read and extend.

As there was no global color already defined that would match the Core background color for the "stripes" style, I've added a "light-gray" color, using the same value as in the Core block styles. I've also introduced another CSS custom property so it's easier to overwrite those in child themes.

Fixes #185

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -27,6 +27,7 @@
 	/* Comments */
 	/* Footer */
 	/* Block: Pull quote */
+	/* Block: Table */
 }
 
 /* Button extends */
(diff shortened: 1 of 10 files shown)


aristath:
Thank you @2ndkauboy :)
I reviewed the PR, resolved merge conflicts and merged it :+1:

PR #210, opened 2020-10-01
🐛 FIX: Adjust READ MORE button (#206)

### LINKED ISSUE #206: The read more block is not contained inside the content area
**Describe the bug**
When adding a read more block (.more-link), it is not aligned together with the rest of the content,
instead it is positioned all the way to the left.

Also reported by @LittleBigThing in https://github.com/WordPress/twentytwentyone/issues/129

**To Reproduce**
1. Add a read more block to a post 
2. View a post archive

**Expected behavior**
I expect it the text to be positioned after the content.

**Screenshots**

readmore

nielslange:
Fixes: #206 

<table>
<tr>
<td>Before:
<br><br>

#206-before
</td>
<td>After:
<br><br>

#206-after
</td>
</tr>
</table>

 **Note**

I removed `&hellip;` from the template as the ellipses was defined outside the READ MORE element which make positioning impossible and as it can easily be added by overwriting the translation, if needed.


melchoyce:
Thanks Niels, can you update this to match the [mockup in Figma](https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=363%3A0)?

image

nielslange:
> Thanks Niels, can you update this to match the [mockup in Figma](https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=363%3A0)?

Sure. I'll do that, @melchoyce. Thanks for attaching the correct design.

nielslange:
<table>
<tr>
<td>Before:
<br><br>

#206-before
</td>
<td>After:
<br><br>

#206-after
</td>
</tr>
</table>

@melchoyce & @carolinan To append the `READ MORE` link to the post, I had to use JS. Given that this `READ MORE` comes from Gutenberg, which requires JS to be activated, I consider the risk very low. What are your thoughts here?

carolinan:
Is it possible to do by filtering `the_content_more_link` ?
https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentytwenty/functions.php#L571

nielslange:
> Is it possible to do by filtering `the_content_more_link` ?
> https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentytwenty/functions.php#L571

Looks like it's possible. Thanks for pointing out that link, @carolinan. While my JS-based implementation works, I'm not 100% happy with it.

luminuu:
Just throwing my 2 cents in here, I'm not sure if we should go with placing items via JavaScript to another position just to match the design. If we want the theme to be a learning place for unexperienced designers and developers, I'd rather prefer the design to be slightly different than using JS in here. I'd go with Nils first iteration of the fix.

melchoyce:
If we can position it using the function @carolinan mentioned, that would be awesome. If not, I agree with @luminuu — I'd rather the theme be straight-forward and easy to understand above all else.

nielslange:
> Just throwing my 2 cents in here, I'm not sure if we should go with placing items via JavaScript to another position just to match the design. If we want the theme to be a learning place for unexperienced designers and developers, I'd rather prefer the design to be slightly different than using JS in here. I'd go with Nils first iteration of the fix.

I know exactly what you mean, @luminuu. Let me try the approach @carolinan mentioned in https://github.com/WordPress/twentytwentyone/pull/210#issuecomment-702711256. It might be possible implementing @melchoyce's design without the need of a JS fallback. Otherwise, let's use the first draft of this PR (as @melchoyce already gave her blessing to that, while I was busy crafting this reply). 😀

luminuu:
@nielslange Just noticed while testing, when hovering the read more link, the `text-decoration` is `none` instead of `dotted`, so it doesn't match the default links. Can you also fix that?

assets/js/more-link.js
```diff
@@ -0,0 +1,8 @@
+( function() {
+	var links = document.querySelectorAll('.more');
```
aristath, line 2: Can we follow the WP coding standards here?

```suggestion
	var links = document.querySelectorAll( '.more' );
```

assets/js/more-link.js
```diff
@@ -0,0 +1,8 @@
...
+
+	links.forEach( link => {
+		var post = link.previousElementSibling;
+		post.appendChild(link);
+	} );
```
aristath, line 7: We don't have babel and we need to support IE11

```suggestion
	links.forEach( function( link ) {
		var post = link.previousElementSibling;
		post.appendChild(link);
	} );
```

aristath (changes requested):

carolinan:
Let's try using the filter solution instead of resolving the current JS problems.

nielslange:
> Let's try using the filter solution instead of resolving the current JS problems.

The filter function will not lead to the expected result, as the READ MORE block is not part of the excerpt. Without JS, only the first solution as pointed out in https://github.com/WordPress/twentytwentyone/pull/210#issue-496246647 is possible.

rolf-yoast:
I agree with @nielslange. The filter can only be used to edit the html of the READ MORE block and is already applied on it in inc/template-functions.php

carolinan:
With this, we have been left with the only option which is to not follow the design for this block.

melchoyce:
Read More on a new line is okay with me, since we can't easily append it to the end of the excerpt. I think more important for the theme to be understandable.

nielslange:
@carolinan Can you open this PR again, so that I can push the first version I created?

carolinan:
@nielslange It is already worked on in a different pull request.
It is my fault for not pinging you but because there was no updates for 5 days and because beta 1 is coming closer I did not think we had the time to wait any longer.

I did not mean to make you feel like your time has been wasted.

Issue #193, opened 2020-09-30
Decrease the line height of h1 titles


LittleBigThing:
The line height of the entry- and page-title seems much larger than in the prototype. I think the prototype suggests 105%, which is about 1.09375. Now, it's the default line height for headings (`--global--line-height-heading: 1.3`).

pattonwebz:
Well spotted :) 

As an aside here, and as a matter of personal preference, I think using `1.1` would be a better call than being so hyper specific with long point values like `1.09375`

LittleBigThing:
Yes, 1.1 was looking good. That is about 106%.

![Screenshot 2020-09-30 at 21 30 01](https://user-images.githubusercontent.com/5901923/94734909-3412dc00-036a-11eb-9402-6025886e8a13.png)

I am not sure if that makes a difference code-wise. I was wondering, however, how to implement it. For me, this value looks good over smaller screen sizes (title sizes, only 2 for now), although the font-size might be a bit too large on small screens (see mobile below).

![Screenshot 2020-09-30 at 22 21 24](https://user-images.githubusercontent.com/5901923/94735621-535e3900-036b-11eb-8466-2012b1727d51.png)

(Both images with `line-height: 1.1`)

nielslange:
@LittleBigThing Will you create a PR for this issue or shall I create one?

LittleBigThing:
Ok, I will later today!

carolinan:
Closing this, as the pull request has been merged. Thank you!

PR #212, opened 2020-10-01
Add and adjust global page title line height

LittleBigThing:
The line height of entry and page titles is reduced from 1.3 to 1.1 for all screens. A take to solve #193.

I am not sure if this is the right approach, so please comment.

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -1606,13 +1606,7 @@ pre.wp-block-verse {
 	font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
 	font-weight: 300;
 	font-size: 4rem;
-	line-height: 4rem;
-}
-
-@media only screen and (min-width: 652px){
-	.wp-block.editor-post-title__block .editor-post-title__input{
-	line-height: 6rem;
-	}
+	line-height: 1.1;
 }
 
 @media only screen and (min-width: 652px){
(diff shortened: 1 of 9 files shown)


assets/css/ie-editor.css
```diff
@@ -1606,7 +1606,13 @@ pre.wp-block-verse {
 	font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
 	font-weight: 300;
 	font-size: 4rem;
-	line-height: 1.3;
+	line-height: 4rem;
```
melchoyce: Is this meant to be rems, instead of unitless?

assets/css/ie-editor.css
```diff
@@ -1606,7 +1606,13 @@ pre.wp-block-verse {
...
+}
+
+@media only screen and (min-width: 652px){
+	.wp-block.editor-post-title__block .editor-post-title__input{
+	line-height: 6rem;
```
melchoyce: Same here ^

assets/css/ie-editor.css
pattonwebz: Should be unitless I think

LittleBigThing:
Hm... is that something I added? 🙃 I didn't mean too. I'll check.

melchoyce:
This may have also come in from somewhere else: https://github.com/WordPress/twentytwentyone/pull/212/files#diff-cec66e3dceab2b8961ebb84d2663010bR1749

assets/sass/05-blocks/utilities/_editor.scss
```diff
@@ -15,7 +15,7 @@
 		font-family: var(--heading--font-family);
 		font-weight: var(--heading--font-weight-page-title);
 		font-size: var(--global--font-size-page-title);
-		line-height: var(--heading--line-height);
+		line-height: var(--heading--font-size-h1);
```
carolinan: I think it is because the line height is set to a font size, not the line height.

carolinan:
The change looks correct, but you need to build the css files again.

PR #215, opened 2020-10-01
File block: Added an underline to the download link in the editor.

### LINKED ISSUE #201: File block: Add an underline to the download link in the editor.
**Describe the bug**
At some point, the underline for the download link, `wp-block-file__textlink`, (not the button) was removed from the _editor_.
This needs to be added back so that the editor and front matches.

shamikakumar:
Fixes #201

### DIFF
--- a/assets/sass/05-blocks/file/_editor.scss
+++ b/assets/sass/05-blocks/file/_editor.scss
@@ -1,9 +1,14 @@
 .wp-block-file {
 
 	.wp-block-file__textlink {
-		border-bottom: 1px solid var(--global--color-secondary);
-		color: var(--global--color-primary);
-		text-decoration: none;
+		text-decoration: underline;
+		text-decoration-style: solid;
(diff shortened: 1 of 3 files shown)


assets/sass/05-blocks/file/_editor.scss
```diff
@@ -1,5 +1,11 @@
 .wp-block-file {
 
+	.wp-block-file__textlink {
```
carolinan, line 3: The link style need to be a text decoration, a solid 1 pixel underline to match the themes other link styles.

The :hover style needs to be an underline, but dotted instead of solid.

carolinan: Unless there is a reason for using a border instead of the text decoration in the editor?

melchoyce:
Hey @shamikakumar! What's your WordPress.org username?

shamikakumar:
@melchoyce my username is shamikakumar

Issue #189, opened 2020-09-30
Responsive menu: fix text overlapping and show site branding


jffng:
In the responsive menu, the menu link text text overlaps on scroll, and the site title + logo are hidden. 

The design for the menu is here: (https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=487%3A61) shows. 

@melchoyce I made a quick prototype in the "Expected" column below. The type styles and spacing are not there yet, but can you confirm or clarify the scroll behavior? Also, should the site description be shown in the responsive menu?

Current | Expected
------ | ------
![Kapture 2020-09-30 at 10 03 34](https://user-images.githubusercontent.com/5375500/94696200-a4c6e380-02eb-11eb-9c83-ef7f167b9f5a.gif) | ![Kapture 2020-09-30 at 10 19 42](https://user-images.githubusercontent.com/5375500/94697803-7944f880-02ed-11eb-96a1-412b3fc302c9.gif)

acosmin:
What is the logic behind showing the logo in the mobile menu? It will only add more complexity, you'll have to consider that the title length can also be huge and images sizes differ. I would only keep the close button.

melchoyce:
Yeah, I think my mockup in this case complicated things more than it needed to. I like having the fixed "close," but I think we can nix including the site title/logo in the mobile menu, so we don't need to worry about size and overflow.

jffng:
> I like having the fixed "close,"

Is it okay that the text overlaps with the links? I think so but just want to confirm.

melchoyce:
Can we add a background color behind it? If so, I think that works.

The extra space at the top isn't ideal, but it's workable, IMO. If it bugs me I'll figure out another idea later, lol.

carolinan:
Overlapping them is a bad idea, because both the menu item links and the close button needs to be visible/readable and clickable.

melchoyce:
> Overlapping them is a bad idea, because both the menu item links and the close button needs to be visible/readable and clickable.

I think that can be worked around by scrolling, but mobile navigation is already a pain, and I don't want to make it harder... To mitigate that usability issue, let's go with this version, sans-site title:

![image](https://user-images.githubusercontent.com/2846578/94846792-b7c2dc00-03ef-11eb-8cef-e32bdfe88363.png)

PR #216, opened 2020-10-01
Solve content shifting to the left when mobile menu is open

### LINKED ISSUE #197: The primary menu shifts the content to the left when opened
**Describe the bug**
When the responsive /mobile version of the menu is opened, the content on the page below it, and the toggle button, shifts to the left.

**To Reproduce**
1. Set up a menu in the primary menu location
2. Reduce the browser window width
3. Toggle the menu

**Expected behavior**
There should be no visual shift.

**Screenshots**
Image by Mel:
An animation shows the shift when the menu is toggled.

acosmin:
Fix #197

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -4988,6 +4988,7 @@ h2.page-title {
 
 .primary-navigation-open .menu-button-container #primary-mobile-menu {
 	display: flex;
+	margin-top: 0;
 }
 
 .primary-navigation {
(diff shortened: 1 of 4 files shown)


carolinan:
This looks really nice in Chrome (PC), but when I switch to Firefox, the buttons are in the wrong position and there is also a visible shift of the content.

menu-firefox

melchoyce:
We might be switching this design to the other option @jffng explored: https://github.com/WordPress/twentytwentyone/issues/189#issuecomment-702309364 So let's pause on this for right now.

PR #217, opened 2020-10-01
Use IE11 ponyfill 

jffng:
This PR tries an alternative to using postcss to create IE specific stylesheets. It still makes use of the WP global `$is_IE` (some discussion in #8).

The benefits of this approach are:
- Doesn't require us to bundle IE front-end and editor specific stylesheets with the theme
- Reduces the build time by removing the postcss build process for IE stylesheets  

h/t @maggiecabrera @scruffian for implementing this in Seedlet.

**To test**
- Check out this PR
- Visit the site on IE 
- Verify that the theme styles are basically supported


carolinan:
I think we should discuss dropping support for IE completely.  If we are not dropping it, I'm not sure that this is an improvement for the users.

The build time is not that long -the npm install time got significantly longer when wp env was added.

scruffian:
The advantage this offers over the current implementation is that it makes it easier for people to extend the theme without using the build tools - they can just update the variables without needing to rebuild.

aristath:
If someone is using IE, it's not by choice but by necessity.
The fact that the ponyfill works on on our own machines doesn't mean that it will work fine on a 10-year-old Pentium PC in a bank... because that is the person who will be stuck on IE. 
Unfortunately polyfills are notoriously bad for performance, and however optimized this one is, it's still a bunch of JS that checks everything in the DOM and changes things.

I realize that the build tools make things harder, and I wish we could just dump them all, write plain CSS instead of sass, and only have the linters (btw, there's nothing stopping us from doing that).
But for IE, I'd say it's better if we have separate stylesheets. Yes, it will be slightly more difficult to maintain. But at the same time, it's easier for IE on an old machine to load a stylesheet than it is to run the polyfill (which btw only works for variables defined in `:root`)

If we want to remove the IE-specific stylesheets, I'd rather just drop them completely without a polyfill. Visitors will still be able to read the content, and that is the most important thing.

carolinan:
As mentioned today during the meeting IE11 still needs to be supported.

kjellr:
I gave this a test locally, and it generally worked in IE. I did notice that the editor styles aren't showing up at all when I load this PR. 🤔 It's just Gutenberg's defaults. 

When I changed colors via the Customizer, the text color didn't update, but it looks like that's the case before this PR too.

jffng:
> I did notice that the editor styles aren't showing up at all when I load this PR.

Yeah this is an issue — the editor would also need to load the polyfill via a new `add_action('enqueue_block_editor_assets')` or something. 

> for IE, I'd say it's better if we have separate stylesheets. 

I agree — the support is better when we compile IE specific front-end and editor stylesheets. 

Since IE11 still needs to be supported, and we're comfortable with the build process as is and the associated costs of maintenance + extensibility, shall we close this PR?

aristath:
> Since IE11 still needs to be supported, and we're comfortable with the build process as is and the associated costs of maintenance + extensibility, shall we close this PR?

Makes sense to close it for now :+1:

PR #227, opened 2020-10-01
FIX: fixing tables with a background-color on a dark page background

### LINKED ISSUE #226: Color contrast to low for tables with a background color set
**Describe the bug**

This is a similar issue to #185

All background color that can be chosen from the table block are light. Currently with a dark background the theme is using a light color for the text. This results in unreadable text with almost any available background color.

There is a second issue with tables using the stripes style and having a background color set, showing an additional border with the color of the background on the bottom of the last row and on the right side of the table. This also happens on light page backgrounds.

**To Reproduce**

1. Go to the customizer and change the body background color to black
2. Create or open a post or page.
3. Add a table block with some content.
4. Select a background color for the table.
5. See that the text color is white on a light background and difficult to read.

**Screenshots**

table-dark-background-before
table-light-background-before

2ndkauboy:
This will set an alternating text color for tables using the stripes style and a dark text color for tables using the regular style. It will also fix the second issue with the borders on bottom and right.

**Screenshots**

| Before | After| 
| --------|------| 
| table-dark-background-before | table-dark-background-after | 

| Before | After |
|--------|------|
| table-light-background-before | table-light-background-after |

Fixes #226

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -1688,11 +1688,27 @@ table.is-style-regular .has-background {
 	color: #28303d;
 }
 
+table.is-style-stripes .has-background thead tr {
+	color: #28303d;
+}
+
+table.is-style-stripes .has-background tfoot tr {
+	color: #28303d;
+}
+
 table.is-style-stripes .has-background tbody tr:nth-child(even) {
 	color: #28303d;
 }
 
 .wp-block-table.is-style-regular .has-background {
+	color: #28303d;
+}
+
+.wp-block-table.is-style-stripes .has-background thead tr {
+	color: #28303d;
+}
+
+.wp-block-table.is-style-stripes .has-background tfoot tr {
 	color: #28303d;
 }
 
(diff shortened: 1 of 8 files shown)


carolinan:
Hi, thank you for this PR, I tested it and the text color in the cells works well. 
The color of the text in the optional table header and footer still change color depending on the body background and not the table background.

I am not sure if something can be done for this for all blocks, I mean, it is going to be a lot of work to adjust individual blocks with background options, and we already see this problem with the buttons. And we have to consider future blocks.  What do you think @aristath ?

luminuu:
@2ndkauboy Can you please check on Carolina's feedback? I also noticed that with choosing one of the default colors for the stripes on the dark color options, the stripes won't work if the second dark color option is chosen.

2ndkauboy:
I've fixed the table header and footer sections as well.

As for the issue with the color of the second dark option: I've recognized that on many element. It is fixed in #252 and we coudl potentially use the same light color here as well. I just don't know, if that would fit into a dark color scheme. Another option would be using a new dark color for the stripes, that is not one of the color options.

aristath:
Since this issue is for the table block and not all other blocks, I believe we can merge it.
It fixes the issue it was created to solve.

As for a more abstract approach that would encompass all blocks, it's possible to do it but I'd suggest we discuss that on another ticket or PR.

Issue #183, opened 2020-09-30
Implement the design for the Pullquote block


carolinan:
@melchoyce is this design correct? 
There is one design for a quote, but I can't tell if it is for the quote block or pull quote block?

**This is the style inherited from Seedlet.**

### Front:

![pullquote_front](https://user-images.githubusercontent.com/7422055/94669057-0f3e4a80-0311-11eb-9d3e-65b83f14981a.png)

### Editor spacing problem:

![pq_editor1](https://user-images.githubusercontent.com/7422055/94669252-4876ba80-0311-11eb-9ed2-e6058e63dff4.png)

### Editor left and right aligned for comparison:
![pq_editor2](https://user-images.githubusercontent.com/7422055/94669741-e9657580-0311-11eb-8151-5619faee239f.png)

Sample content:
https://gist.github.com/carolinan/f6528e0839ccd7a2a45fc935722591f5

Code: 
https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/pullquote

carolinan:
Pullquotes in footer widgets:

With the current design, the wide pull quote is actually wider than the full width pull quote.
![pq_widget](https://user-images.githubusercontent.com/7422055/94677176-c68c8e80-031c-11eb-95c1-d1f3187f9371.png)

melchoyce:
How would these designs work for quote and pullquote blocks?

![image](https://user-images.githubusercontent.com/2846578/94748377-2865ec00-034f-11eb-823e-d63502c6eb35.png)
[Figma](https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=1254%3A0)

-----

![image](https://user-images.githubusercontent.com/2846578/94748466-65ca7980-034f-11eb-9b6e-37ede4059f1f.png)
[Figma](https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=1254%3A64)

luminuu:
@melchoyce Do we get the quote icon as SVG? Using the system font, I get this (sneaked the code from Twenty Twenty):

![grafik](https://user-images.githubusercontent.com/7585600/94859945-a56b9100-0435-11eb-8c84-4ca01c88139e.png)

melchoyce:
[...]

PR #229, opened 2020-10-01
New pullquote block design

luminuu:
Referring to issue #183 about the new pullquote block designs.

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -1372,6 +1372,7 @@ p.has-background {
 	border-bottom-style: solid;
 	border-bottom-width: 3px;
 	color: #28303d;
+	position: relative;
 }
 
 .wp-block-pullquote::before {
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -1364,18 +1364,30 @@ p.has-background {
 	padding: 40px 0;
 	margin-left: 0;
 	margin-right: 0;
-	text-align: left;
-	border-top-color: transparent;
-	border-top-width: 0;
-	border-bottom-color: transparent;
-	border-bottom-width: 0;
+	text-align: center;
+	border-top-color: #28303d;
+	border-top-style: solid;
+	border-top-width: 3px;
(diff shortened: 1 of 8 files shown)


luminuu:
Updated the smart quote as Mel suggested, now it's ready for review!

assets/css/ie-editor.css
```diff
@@ -1364,26 +1364,38 @@ p.has-background {
...
+	border-top-width: 3px;
+	border-bottom-color: #28303d;
+	border-bottom-style: solid;
+	border-bottom-width: 3px;
 	color: #28303d;
 }
 
+.wp-block-pullquote::before {
+	color: currentColor;
+	content: "“";
```
pattonwebz: Generally unicode chars are used for CSS `content` if any non-standard chars are added there.
```suggestion
	content: "\201C";
```

luminuu: Thank you! Changed that one.

melchoyce (changes requested):
Let's decrease the padding on the solid style on narrower screens, right now the text gets pretty cramped:



Otherwise, looks good to me!

melchoyce (approved):
LGTM

carolinan:
Do we want to limit the width of the left and right aligned pullquote to 50%? When the content is long enough, the quotes are no longer aligned but looks like they are normal width without alignments.

carolinan:
In the editor, left and right aligned pull quotes needs spacing to the right and left of the block respectively.

It is less visible on the right:
pull quotes in the editor is missing spacing

carolinan:
Related: https://github.com/WordPress/twentytwentyone/pull/250

melchoyce:
> Do we want to limit the width of the left and right aligned pullquote to 50%? When the content is long enough, the quotes are no longer aligned but looks like they are normal width without alignments.

Let's explore this in a new issue, I'll want to try out some ideas.

melchoyce:
@carolinan Moved your feedback to https://github.com/WordPress/twentytwentyone/issues/282

PR #230, opened 2020-10-02
Update audio focus styles

### LINKED ISSUE #207: Audio block: Improve focus style
**Describe the bug**
When the audio block is added, it looks different in different browsers as expected, but the outer focus style in all browsers needs to be more visible.

In Firefox, there is no spacing between the element and the thin dotted outline so the outline is difficult to see.
In chrome, the focus is thicker, but there needs to be more spacing, see screenshot below.

**To Reproduce**
1. Add an audio block. There is an audio file in the [Theme Unit Test](https://github.com/WPTT/theme-unit-test) if needed
2. View the front
3. Place focus on the block

**Screenshots**
Chrome at the top, Firefox below
Focus styles for the audio block are not visible enough

carolinan:
Updates the audio focus styles.
The primary color is used to make sure that the focus is visible on all backgrounds.

Removes the style for the width since the same width is already in the block library styles. The width of the aligned items was tested in multiple combinations with other blocks.
See 
https://github.com/WordPress/gutenberg/commit/567875eaa8dfddd863dd972e7e96ec4306798801#diff-cc1a89df338f7b42f8fd932ff37e9407

Closes https://github.com/WordPress/twentytwentyone/issues/207

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -2088,12 +2088,9 @@ a:hover {
 }
 
 /* Category 05 is all about adjusting the default block styles to the given layout. I only added three blocks as examples. */
-.wp-block-audio {
-	min-width: inherit;
-}
-
-.wp-block-audio.alignleft, .wp-block-audio.alignright {
-	min-width: 300px;
+.wp-block-audio audio:focus {
+	outline-offset: 5px;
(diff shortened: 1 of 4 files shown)


melchoyce:
LGTM 👍

PR #239, opened 2020-10-02
Set PHP version to 5.6 to match PHPCS

kraftbj:
In setting up the demo site, which runs on a server using the WP minimum of PHP 5.6, the theme would not activate.

In examining the `phpcs.xml.dist`, we are testing for PHP 5.6 compat and, by manually removing the line from `style.css`, the theme appears to activate and work without issue.

This PR would update the minimum PHP version to the WP current minimum of 5.6.

### DIFF
--- a/style.css
+++ b/style.css
@@ -7,7 +7,7 @@ Author URI: https://wordpress.org/
 Description:
 Requires at least: 5.3
 Tested up to: 5.5
-Requires PHP: 7.2
+Requires PHP: 5.6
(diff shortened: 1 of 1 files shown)


pattonwebz (approved):
Thus far there has been no code included which does not work on PHP 5.6 so I think this is a safe change to make. 5.6 is also the current min supported version of PHP which WordPress supports and that doesn't seem like it's changing in the next WP release.

Issue #57, opened 2020-09-18
Update custom logo settings, including title and tagline


carolinan:
In addition to https://github.com/WordPress/twentytwentyone/pull/56/,

- [x] the `unlink-homepage-logo` needs to be added to the theme support for custom logo.
See https://make.wordpress.org/core/2020/07/28/themes-changes-related-to-get_custom_logo-in-wordpress-5-5/

<del>- [ ] Apply the different design and positions of the logo and site title
https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=468%3A328
https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=487%3A17</del>

- [x] A decision is needed regarding the logo size limit, currently set to 240x240 with no option to change the size.
See https://github.com/WordPress/twentytwentyone/blob/trunk/functions.php#L84
https://github.com/WordPress/twentytwentyone/blob/trunk/assets/sass/01-settings/global.scss#L160

- [x] Add option for retina logo
See https://themes.trac.wordpress.org/browser/twentytwenty/1.5/classes/class-twentytwenty-customize.php#L65
https://themes.trac.wordpress.org/browser/twentytwenty/1.5/functions.php#L66

- [x] Display logo in site footer

- [ ] Adjust footer logo size.

carolinan:
> I'm thinking a max of 300px wide, or 100px tall?

 The option only has one regular size, and one for the retina that we will add. It does not accept min or max size values, but I'll do my best and we can test and alter the values.

LittleBigThing:
Why don't we make the ‘retina logo‘ the default and drop the (theme) option? And simply recommend the higher (‘retina‘) image size. Or am I missing something?

briceduclos:
Currently, the retina option scales the logo to half its uploaded size (x2). However, more and more devices now have higher screen density (x3.5). So a x2 logo would appear blurry and not sharp on them. As it’s not possible to upload svg, a solution might be to replace the “Retina logo” checkbox with a “Logo width” input number.

aristath:
[...]

PR #244, opened 2020-10-02
Allow for flexible logo sizes

ryelle:
A larger image can be uploaded and displayed, and the CSS will take care of making it the expected size on the frontend. Using this method, we can remove the `retina_logo` option, because larger image sizes will be supported.

See #57 

To test:

- Open the customizer and choose a large (1000px+) image for your site logo
- You should be able to skip the cropping step, or crop it to the 3:1 aspect ratio
- When saved, you see the image in the header & footer 👍

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -4139,6 +4139,7 @@ nav a {
 	max-width: 96px;
 	max-height: 96px;
 	height: auto;
+	width: auto;
 }
 
 @media only screen and (min-width: 482px) {
@@ -4146,6 +4147,7 @@ nav a {
 		max-width: 300px;
 		max-height: 100px;
 		height: auto;
+		width: auto;
 	}
 }
 
--- a/assets/sass/06-components/header.scss
+++ b/assets/sass/06-components/header.scss
@@ -89,6 +89,7 @@ nav a {
 		max-width: var(--branding--logo--max-width-mobile);
 		max-height: var(--branding--logo--max-height-mobile);
 		height: auto;
+		width: auto;
 	}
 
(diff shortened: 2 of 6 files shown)


melchoyce:
Just tested this. Header works great! I'm seeing the logo at full size in the footer, though:

image

LittleBigThing:
I used an image with a different aspect ration (than 3:1) and it does not seem to look as expected when the logo is set without resizing the image:

Screenshot 2020-10-02 at 21 32 41

It is due to the max-width and max-height defined [here](https://github.com/WordPress/twentytwentyone/blob/118e3d585da225fe600d2ee6c77bc955a62e42cf/assets/sass/06-components/header.scss#L88), based on the fixed width and height.

We should either set a `max-height` or a `max-width` and set the other to auto?
In Twenty Twenty, a `max-height` is set and the `width` is auto, this is probably the way to go except when the logo can be centered and then it could be higher? But that is going to make things more complicated.

And indeed, the full-size image visible in the footer.

ryelle:
I've added the same wrapper to the footer logo as is in the header, so now it's contained using the same CSS. I've also added `width:auto` so the image isn't skewed when it's not the right aspect ratio - this way the image should always be in the 300x100 container, but will scale down rather than skew.

jffng (approved):
I like this approach.

Could be addressed in a separate PR but should the `.site-logo` margins be adjusted so the logo is left-aligned with the site title?

Screen Shot 2020-10-05 at 10 16 27 AM

melchoyce:
Margins fixed in #246.

Issue #73, opened 2020-09-19
Search / search results page


luminuu:
The are designs given for three search results related views: https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=269%3A0

- [x] Search results with multiple entries
- [x] Search result page if no matching posts were found

~~Right now the theme does not feature any search icon in the header or footer areas, so we're not sure from where the first design should be reached.~~

The search results page has their own template file and may be changed if the default archive file changes.

The third design already has an issue: #31

melchoyce:
> Right now the theme does not feature any search icon in the header or footer areas, so we're not sure from where the first design should be reached.

You can reach the results pages by using a search block or widget.

luminuu:
Yup! That sentence referred to the initial search results page with just the search input itself, which we don't have anymore. I've striked it through.

melchoyce:
Ahh cool, okay 👍

carolinan:
I believe these two pages are completed and that this can be closed?

luminuu:
![Bildschirmfoto am 2020-10-02 um 21 15 21](https://user-images.githubusercontent.com/7585600/94961170-6cdebc80-04f4-11eb-8b9d-fbc7ea7ed345.png)

So I've been just testing the mobile version of the search results screen, I think we do not want to have the search results headline break like this?

The code which causes this can be found here: https://github.com/WordPress/twentytwentyone/blob/trunk/assets/sass/06-components/archives.scss#L3

All other headlines on blog posts and regular pages look normal, it seems only the archive-pages have the `break-word` property.

melchoyce:
Good catch, let's get rid of that property.

PR #245, opened 2020-10-02
Remove unnecessary CSS property break-word

luminuu:
Fixes the issue mentioned in #73.


carolinan:
but the CSS files are not updated with this PR?

luminuu:
Good catch! Compiled files added.

PR #246, opened 2020-10-02
Remove horizontal margin on the site logo

melchoyce:
Currently, the logo has a left and right margin, which means the logo does not align to the side of the container. This PR removes that margin.

Before | After
----- | -----
image | image


carolinan:
Looks good to me, do you mind refreshing, to solve the problem with the PHP version in style.css?

PR #249, opened 2020-10-03
Fix fullwidth blocks

### LINKED ISSUE #236: Media & Text: Full width issues
**Describe the bug**

**1)** When a stacked full width media and text block with the border style is added,
the left most border is cut off on smaller screens.

**2)** When a stacked full width media and text block without a border is added, there is no spacing between the text and the edge of the browser.
-I am not sure if this is intended since it is a full width block, but for me the text is more difficult to read.
Both the wide and normal width block has some spacing for the text when the block has the stacked option enabled.

**3)** In Firefox, there is a horizontal scroll on smaller screens when a full width media and text block is added: 
- With and without border, 
- With and without the stacked option enabled.
- With and without a cropped image.
- Independent of the image size.
 

**To Reproduce**
1. Add media and text blocks as described above.
2. View the front
3. Reduce the browser window width

media

### LINKED ISSUE #240: Image block: full width issues.
**Describe the bug**
The image block has similar problems as described in https://github.com/WordPress/twentytwentyone/issues/236
but I only see the issues in Firefox.

There is a horizontal scroll on smaller screens when a full width image block is added,
with and without border or framed border.

The left border is not visible on smaller screens when a full width image block with a border or framed border is added.

image-border

ct-martin:
Devices below the small (482px) breakpoint have full-width blocks set to `calc(100% + 50px)` for `width` and `max-width`, and a left margin of `-25px`. This makes fullwidth blocks overflow the screen.

This removes that behavior and uses the maximum width allowed.

Fixes #236 
Fixes #240

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -925,23 +925,23 @@ hr.wp-block-separator.is-style-wide {
 }
 
 .full-max-width {
-	max-width: calc(100% + 50px);
-	width: calc(100% + 50px);
-	margin-left: -25px;
+	max-width: 100%;
+	width: 100%;
+	margin-left: auto;
 	margin-right: auto;
 }
 
 .wp-block-group .wp-block-group__inner-container > *.alignfull {
-	max-width: calc(100% + 50px);
-	width: calc(100% + 50px);
-	margin-left: -25px;
+	max-width: 100%;
+	width: 100%;
+	margin-left: auto;
 	margin-right: auto;
 }
 
 .alignfull {
-	max-width: calc(100% + 50px);
-	width: calc(100% + 50px);
-	margin-left: -25px;
+	max-width: 100%;
+	width: 100%;
+	margin-left: auto;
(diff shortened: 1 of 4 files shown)


ct-martin:
Note with respect to #236, fullwidth text not having padding seems to be the intended behavior. With grouping I'm unsure if it makes sense to add padding to fullwidth paragraph blocks. On one hand, the user may not want extra padding for some reason. On the other, I definitely agree that it's hard to read. Another option would be to have a fullwidth group with a fullwidth paragraph inside that? Open to discussion, but might make sense to have in another issue

luminuu:
Looks good to me! Tested:
- full width group block with paragraph inside
- full width media and text
- full width image

All with and without border style.

carolinan:
Looks good to me, the border is visible and the scroll is no longer visible.

aristath (approved):
confirmed, LGTM :+1:

PR #250, opened 2020-10-03
Fix left & right block align

### LINKED ISSUE #237: Left and right aligned gallery blocks are not aligned on the front
**Describe the bug**
When a gallery block is added and aligned to the right or left,
it does not look the same in the editor and on the front.

On the front, the gallery keeps the normal width and is not aligned.
The images are not reduced in size in the same way they are presented in the editor.

**To Reproduce**
1. Add a gallery block to a post or page
2. Align it to the left or right, with or without content next to it.
3. Save and view the front.

**Expected behavior**
I expected the galleries to be aligned to the left or right and the image sizes to be adjusted.

**Screenshots**
gallery with missing alignments on the front

ct-martin:
The underlying cause for #237 ended up being that the theme has default behaviors for left- & right-aligned content, which does not actually left- or right-align block elements.

Here's a breakdown of what was done in this PR:
* The gallery block has `.alignleft` & `.alignright` styles, but they get overridden by the global `.alignleft, .alignright` styles so I removed them (as they were redundant)
* Fixed `max-width` to use `calc(50% - var(--responsive-align[left/right]-margin]))`
  * (Note: the article content is technically full-width and has a margin; this takes that in to account since just `50%` will be oversized)
* **The default behavior for `.alignleft` & `.alignright` is now to be one-column**
  * The small breakpoint is 482px and the default behavior covers sizes below that
    * To have two-columns below this would mean a max size of 240px-ish, which is hard to see (particularly for media)

**This PR could use extra testing.** I've tried to test it the best I could, but it would be good to have extra eyes on given how this could change behavior. In particular, if anyone knows of somewhere outside of `.entry-content` where `.alignleft`/`.alignright` are used, that should be tested.

Fixes #237

### DIFF
--- a/assets/sass/05-blocks/utilities/_style.scss
+++ b/assets/sass/05-blocks/utilities/_style.scss
@@ -82,7 +82,7 @@
 		margin-left: var(--global--spacing-horizontal);
 	}
 
-	.entry-content > .alignleft {
+	.entry-content > .alignright {
 		max-width: calc(50% - var(--responsive--alignright-margin));
 	}
 }
--- a/assets/sass/05-blocks/utilities/_style.scss
+++ b/assets/sass/05-blocks/utilities/_style.scss
@@ -16,21 +16,31 @@
 	/*rtl:ignore*/
 	text-align: left;
 
-	/*rtl:ignore*/
-	float: left;
 	margin-top: 0;
-
-	/*rtl:ignore*/
-	margin-right: var(--global--spacing-horizontal);
-	margin-bottom: var(--global--spacing-vertical);
 }
 
 // Targeting the .entry-content class is necessary to ensure these styles
 // only apply when the block isn't nested.
 .entry-content > .alignleft {
 	max-width: var(--responsive--aligndefault-width);
-	@extend %responsive-alignleft-mobile;
 	@extend %responsive-alignleft;
+}
+
+@include media(mobile) {
+
+	.alignleft {
+
+		/*rtl:ignore*/
+		float: left;
+
+		/*rtl:ignore*/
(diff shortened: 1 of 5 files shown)


luminuu:
Looks good to me, tested multiple blocks that can be aligned left/right.

assets/sass/05-blocks/utilities/_style.scss
```diff
@@ -50,23 +60,33 @@
...
+	.alignright {
+
+		/*rtl:ignore*/
+		float: right;
+
+		/*rtl:ignore*/
+		margin-left: var(--global--spacing-horizontal);
+	}
+
+	.entry-content > .alignleft {
```
carolinan: Is this meant to be `.alignright?`

carolinan: tested and confirmed that it only works if this class is replaced.
CSS files needs to be rebuilt after this change.

```suggestion
	.entry-content > .alignright {
```

ct-martin: @carolinan Oops! Thanks for catching, I thought I had done that :sweat_smile:

carolinan (approved):

jffng (approved):

jffng:
As a follow up, I wonder if we should remove [these mixins](https://github.com/WordPress/twentytwentyone/blob/trunk/assets/sass/03-generic/breakpoints.scss#L166-L204) and rename `--responsive--alignleft-margin` to `responsive--alignleft-width`.

melchoyce:
@jffng Can I have you make a new issue for that when you have a chance? Going to merge this now.

PR #254, opened 2020-10-03
Update button styles

carolinan:
This is a follow up on [a comment about reducing the editor button styles and using the extend.](https://github.com/WordPress/twentytwentyone/pull/142#discussion_r496164506)

It also:
Fixes a problem where a clicked button (the active state) would have the incorrect text color.
Fixes a problem where the text was not visible when the outlined button was selected in the editor.

The styles were tested with Gutenberg version 9.1, and the 
```
div[data-type="core/button"] {	
	display: block;	
}
```
in the editor style was removed because it caused the buttons block to display vertically not horizontally.

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -31,6 +31,19 @@
 }
 
 /* Button extends */
+.wp-block-button__link {
+	line-height: 1.5;
+	color: #d1e4dd;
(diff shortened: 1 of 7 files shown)


kjellr (approved):
Works in my testing!

jffng (approved):

Issue #80, opened 2020-09-20
Accessibility: WCAG 2.4.8 Location Level AAA 


carolinan:
Information about the user’s location within a set of Web pages is available.
Reference:
https://www.w3.org/TR/WCAG21/#location

- [x] aria-current in pagination for page breaks within content
- [x] aria-current in archive pagination
- [x] aria-current in comment pagination **-Depends on the final design, currently only "newer comments" and "older comments" is used for this navigation.**
- [x] aria-current in widgets, from WP 5.3 (Widgets are optional).
- [x] aria-current in the primary menu  **-We need to confirm that the navigation menu includes aria-current for all the WP versions that the theme supports.**

@audrasjb Would the above be enough, or is a bread crumb required?
If not required, is adding a breadcrumb an improvement that we should still consider?

carolinan:
Closing this since the comments pagination has been merged.

PR #256, opened 2020-10-03
Update comments pagination

### LINKED ISSUE #177: Comment navigation is missing styles
**Describe the bug**
A decision is needed regarding the pagination/ navigation for comments that are split into multiple pages. 

Either a paged pagination that shows the number of pages, or, only links to the next and previous comments pages.

What width would this navigation have, and should it be placed above or below comments, or both?

carolinan:
Adds styling to the comment pagination.
Removes the comment pagination that was above the comments.
Removes some link border styles that are no longer used since we are using an underline.

comment

Closes https://github.com/WordPress/twentytwentyone/issues/177 and https://github.com/WordPress/twentytwentyone/issues/80

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -924,6 +924,24 @@ hr.wp-block-separator.is-style-wide {
 	}
 }
 
+.comments-pagination {
+	max-width: calc(100vw - 30px);
+	margin-left: auto;
+	margin-right: auto;
+}
+
+@media only screen and (min-width: 482px){
+	.comments-pagination{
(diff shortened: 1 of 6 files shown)


kjellr (approved):
This looks good to me. We might want to fix up the display of this on mobile: 

Screen Shot 2020-10-05 at 2 37 39 PM

... but that could be a separate PR.

melchoyce:
re: overflow on mobile, if we can't do this:

image

Could we do something this?

image

(I am cool with us doing this in a separate PR, don't want to hold this one up)

PR #257, opened 2020-10-03
Update social links alignment in widgets

### LINKED ISSUE #205: The social icon block does not align to the right in the widget area
**Describe the bug**
When the social icon block is added inside a widget area, the right alignment works on the widget screen, but not on the front.
Center and left aligns work correctly.

From @luminuu 
> In my recent test the social icons block did not align right due to the given setup (float is added but as it's a flex item float doesn't work, it needs to be justify-content: flex-end). It's also not possible to style the icons with the dark grey color unless you add additional CSS via the customizer.

**To Reproduce**
1. Install and activate Gutenberg to enable blocks in the widget screen
2. Add a social icon block
3. Align the block to the right
4. Go to the front

**Expected behavior**
I expect the icon to align correctly on the front.

carolinan:
Corrects alignments for the social links block when placed inside a widget area.

Closes https://github.com/WordPress/twentytwentyone/issues/205

-I went back and forward on if this should be placed in widgets, or blocks?

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -5438,6 +5438,15 @@ h1.page-title {
 	}
 }
 
+.widget-area .wp-block-social-links.alignright {
+	margin-top: 30px;
+	justify-content: flex-end;
+}
+
+.widget-area .wp-block-social-links.alignleft {
+	margin-top: 30px;
+}
+
 .widget-title {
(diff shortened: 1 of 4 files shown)


luminuu:
Looks good to me!

PR #258, opened 2020-10-03
Adding example gradient presets

### LINKED ISSUE #46: Add custom gradients
The theme currently has no custom gradients, but a placeholder for editor-gradient-presets has been left in functions.php
to make them easier to include. 
If no custom gradients are added, this code needs to be removed in time for beta 1.

LittleBigThing:
Some gradient presets as a start/example, including color-stop gradients inspired by the [Seedlet theme](https://github.com/Automattic/themes/tree/master/seedlet).

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -2099,18 +2099,10 @@ a {
 	background: linear-gradient(to bottom right, #d1e4dd 49.9%, #d1dfe4 50%);
 }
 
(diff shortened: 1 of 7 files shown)


melchoyce:
Now that I can see them in practice (thank you for adding them here!) I don't think the hard-stop gradients work so much with the theme's softer aesthetic. I think we should probably stick to smooth color transitions.

Issue #247, opened 2020-10-02
Hide homepage title


melchoyce:
Let's remove the whole `.entry-header` on the front page. The site title already serves as an `h1`, so we shouldn't need to worry about any a11y repercussions, I believe?

PR #260, opened 2020-10-03
Fix serach results page, removing home.php and template-parts/content…

### LINKED ISSUE #73: Search / search results page

luminuu:
…/blog.php

Fixes #73.


luminuu:
Fixes #247 as well.

PR #264, opened 2020-10-04
Fix text color on light backgrounds #261

### LINKED ISSUE #261: Group block: dark color schemes don't work with light background
**Describe the bug**
When a dark color scheme is chosen, any group / media and text / cover blocks have a light background but also a light font color.

**To Reproduce**
1. Open Customizer
2. Choose one of the dark color schemes
3. Create post with blocks using background colors (group, media/text, cover)
4. Font color is also light

**Expected behavior**
Make sure text is readable with a dark color scheme and blocks providing a different background color.

**Screenshots**
grafik

- WordPress version: 5.5.1
- Does the website has Gutenberg plugin installed, or is it using the block editor that comes by default? Gutenberg plugin active
- If the Gutenberg plugin is installed, which version is it? 9.1

 - OS: macOS Catalina
 - Browser: Firefox 81

sudoshreyansh:
The dark background classes had their text color defined, but the light background classes didn't.
This PR fixes that and closes #261 

My WordPress.org username is sudoshreyansh


carolinan:
Hi, this does not work when the user changes the text color.
If the user changes the text color, it still shows as black on the front.

Black is not used as a text color elsewhere in the theme, I think it would be better to use the dark gray color` 	--global--color-dark-gray: #28303d;`   so that it matches the other text.

sudoshreyansh:
Hi @carolinan 
I was able to reproduce the error, and I have fixed that. Also I have, according to your suggestion replaced `--global--color-black` with `--global--color-dark-gray`

SResok (commented):
CR Done, looks good to me! 🔎

SResok:
Acceptance done 👍🏼

carolinan:
I will merge this and open an issue for the pull quote which is the only block I see that the text color does not work for.

PR #265, opened 2020-10-04
Tools: Try a workflow for building CSS

ryelle:
This creates a workflow using the [Add & Commit](https://github.com/marketplace/actions/add-commit) action to run the build process and update any changed files on pushes to `trunk`. This can ensure that anyone checking out (or downloading a zip) from github gets the latest CSS without having to run anything, and should prevent unexpected local changes when viewing trunk.

This doesn't run on branches (though it could), because IMO that could be confusing for a contributor to have a bot pushing to their branch. It would run once their branch is merged though.

You can see this on my fork:
- Commit some changed SCSS: https://github.com/ryelle/twentytwentyone/commit/4729f87c9afcf4d430f84bddd972b6990caf6618
- Wait for [the actions to run](https://github.com/ryelle/twentytwentyone/runs/1206021747)
- The changes are detected and committed back to trunk: https://github.com/ryelle/twentytwentyone/commit/a02d7fb54103c1d3cfad22a11aa8f0af7cb4e69b

I also removed the `style.css.map` file because it was only generated once in the initial commit, hasn't been updated since, and `build` does not create sourcemap files.


carolinan:
I'm afraid I don't know how to test it, but as a side note,  the themes team does not allow .map files to be included in themes in the directory either.

ryelle:
@carolinan Unfortunately I don't think you can, since it only runs on trunk — but that's why I linked to it running in my fork.

Issue #8, opened 2020-09-15
Discussion:  Methods for including IE specific CSS


carolinan:
Currently, two separate style sheets are created for Internet Explorer by running the final style sheets through postcss:
https://github.com/WordPress/twentytwentyone/blob/trunk/assets/css/ie-editor.css
https://github.com/WordPress/twentytwentyone/blob/trunk/assets/css/ie.css

These files are only loaded when the `$is_IE` WordPress global is set to true.  This global is created by checking the `$_SERVER['HTTP_USER_AGENT`.

-Is this the preferred method or should the theme include fall backs for IE in the main style.css and editor styles?
Are there other methods that should be considered?

+User Agent is not completely reliable
+Less files

-Larger files
-Unused styles will always be loaded

kjellr:
If that user agent check is good enough for core, I imagine it's good enough for the theme. If it's working as is, I'm not sure it makes sense to invest in a different solution.

aristath:
I agree with @kjellr 
Though the `$is_IE` global is far from perfect (has lots issues with aggressive caching plugins), for the sake of simplicity it's fine to use it here.

> The other approach could be feature detection, but I'm not sure it's allowed to load the theme's stylesheet via client side JS?

It's allowed and works fine. Surprisingly it's also extremely easy to do... We can simply add this in `functions.php`:
```php
add_action( 'wp_footer', function() {
	?>
	<script>
	if ( ! window.CSS || ! CSS.supports( 'color', 'var(--fake-var)' ) ) {
		var style = document.createElement( 'link' );

		style.rel = 'stylesheet';
		style.type = 'text/css';
		style.href = '<?php echo esc_url( get_theme_file_uri( 'assets/css/ie.css' ) ); ?>';
		document.head.append( style );
	}
	</script>
	<?php
} );
```
The reason I'm suggesting it be added inline is because there's no reason to enqueue yet another script file - especially when it's that small and we need to get the URL via PHP, gzipped that's < 200 bytes.
It's equally possible to add it using `wp_enqueue_script()` and use `wp_localize_script()` to pass the file-url as a var from there, but it just adds useless overhead and will force the browser to download yet another asset for the 1-2% of users that will need it.

carolinan:
But do we need and or want to support IE?

aristath:
_Need_ is a strong word...
As I've said before, the fact that we "support" IE doesn't mean that IE should have the same styles as other browsers. As long as the content is readable in IE I don't think we need to do something.

PR #266, opened 2020-10-05
Use `$wp_styles` conditional instead of `$is_IE` to load IE-specific stylesheets

aristath:
This PR changes the way the IE-specific stylesheet is loaded on the frontend. Instead of using `$is_IE` in PHP to detect if the browser is IE, it uses `$wp_styles->add_data()`.

The IE-specific stylesheet will be wrapped in `<!--[if IE]> ... <![endif]-->`

The reason for this change is caching: Relying on PHP variables for client detection is bad practice. If the site is behind a reverse proxy or is using some form of aggressive caching, then the 1st time someone visits the site is what will be cached. If the 1st visit is not IE, nobody will get the IE styles regardless of what browser they are using.
If on the other hand the 1st visit is from IE, then everyone will get the IE styles.

With this PR we don't use the PHP var and everyone gets what they need.
Note that this change was only made for the frontend stylesheet. For the editor styles it's OK to use the PHP implementation since the dashboard is not cached.

The downside: IE will get the full stylesheet twice (1st the normal one and then the IE-specific one). But I think we can live with that...

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -326,16 +326,11 @@ add_action( 'after_setup_theme', 'twenty_twenty_one_content_width', 0 );
  * Enqueue scripts and styles.
  */
 function twenty_twenty_one_scripts() {
-	// Note, the is_IE global variable is defined by WordPress and is used
-	// to detect if the current browser is internet explorer.
-	global $is_IE;
-	if ( $is_IE ) {
-		// If IE 11 or below, use a flattened stylesheet with static values replacing CSS Variables.
-		wp_enqueue_style( 'twenty-twenty-one-style', get_template_directory_uri() . '/assets/css/ie.css', array(), wp_get_theme()->get( 'Version' ) );
-	} else {
-		// If not IE, use the standard stylesheet.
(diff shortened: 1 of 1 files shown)


aristath:
Related: https://github.com/WordPress/twentytwentyone/issues/8 (possibly fixes that issue?)

carolinan:
I don't know enough about cache to review it.

But if IE gets both stylesheets, why don't we put IE safe styles in the standard one...?
That idea has already been turned down though.

aristath:
No need to know anything about cache to review it...
If the stylesheet gets applied on IE then it's OK :+1:

carolinan:
I only have IE11, where the IE styles are not loaded

It says here that  `<!--[if IE]>` only works on versions below 10

https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/compatibility/hh801214(v=vs.85)?redirectedfrom=MSDN

aristath:
Hmmmm that's too bad...
In that case I'll go ahead and close this PR.
Thank you for checking!

PR #267, opened 2020-10-05
Add a screen reader text with the site title to the front page

### LINKED ISSUE #251: Site title and tagline are removed completely when display option unchecked
When the option to display the site title and tagline is unchecked in the Customizer, the site branding is removed completely. The template part is [not included](https://github.com/WordPress/twentytwentyone/blob/691ae72442f425beceec347088ea4c2c3d6f5474/template-parts/header/site-branding.php#L10) anymore, even when there is no custom logo added.

Shouldn't we rather make them visually hidden for accessibility (and search engines), similar to how we do it for screen reader texts?

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/251

### DIFF
--- a/template-parts/header/site-branding.php
+++ b/template-parts/header/site-branding.php
@@ -6,12 +6,6 @@
  * @subpackage Twenty_Twenty_One
  * @since 1.0.0
  */
-
-// If there is neither a title, tagline or logo, we don't want to print the empty site branding tag.
-if ( get_theme_mod( 'display_title_and_tagline', true ) === false && ! has_custom_logo() ) {
-	return;
-}
-
 ?>
 <div class="site-branding">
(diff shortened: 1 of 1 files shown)


carolinan:
@LittleBigThing Is this how you intended it? It does not add the site title to pages that already has an H1.

It also does not solve the missing H1 if the title field is empty, but that is very much an edge case.

ryelle:
This works for "latest posts" as home, but not for pages — it's now showing two headings on the homepage & none on the posts page.

carolinan:
Should we just move the page titles down one heading level and always use the site title as the H1?

aristath (approved):
LGTM :+1:

PR #268, opened 2020-10-05
Hide the site title in the footer when the display option is unchecked

carolinan:
If the Display Site Title & Tagline option is unchecked, hide the site title in the footer.

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -27,7 +27,7 @@
 					<?php
 					else :
 						$blog_info = get_bloginfo( 'name' );
-						if ( ! empty( $blog_info ) ) :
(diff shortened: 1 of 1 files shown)


footer.php
```diff
@@ -27,7 +27,7 @@
 					<?php
 					else :
 						$blog_info = get_bloginfo( 'name' );
-						if ( ! empty( $blog_info ) ) :
+						if ( ! empty( $blog_info ) && get_theme_mod( 'display_title_and_tagline', true ) === true ) :
```
aristath: We don't need to check if `===true`, we can simplify this check as per the suggestion below
```suggestion
						if ( ! empty( $blog_info ) && get_theme_mod( 'display_title_and_tagline', true ) ) :
```

aristath (changes requested):

PR #269, opened 2020-10-05
ESLint configuration & fixes

aristath:
Adds an eslint config and applies the fixes.

Once this is done we can check out how to do the auto-linting on PRs.


carolinan:
I can't get this to work on Windows, I have never used @wordpress/eslint-plugin, but I think we need both eslint and the plugin?

package.json
```diff
@@ -15,6 +15,7 @@
 	"devDependencies": {
 		"@wordpress/browserslist-config": "^2.2.2",
 		"@wordpress/env": "^2.0.1",
+		"@wordpress/eslint-plugin": "^7.2.1",
```
carolinan, line 18:
```suggestion
		"@wordpress/eslint-plugin": "^7.2.1",
		"eslint": "^7.10.0",
```

carolinan:
besides the suggestion above, add `"lint:js": "eslint assets/js",` to scripts (no ending comma if it is the last item)

With these changes, I confirmed it is working (for  me ;) ) on Windows.

PR #273, opened 2020-10-05
fixes #209

### LINKED ISSUE #209: Default button style on frontend is broken
**Describe the bug**

The default button style on frontend is currently broken.

**To Reproduce**
1. Import the [Theme Unit Test](https://github.com/WPTT/theme-unit-test)
2. Go to the homepage
3. Scroll down to post **Block category: Layout Elements**

**Expected behavior**

The button should look like this as seen in https://github.com/WordPress/twentytwentyone/issues/36#issuecomment-694324773:

Screenshot 2020-10-01 at 19 17 47

**Current behavior**

The button looks like this:

Screenshot 2020-10-01 at 19 18 15

**Note**

This bug only affects the frontend styles, not the editor styles. It also does not affect the `:focus` and `:hover` state.

aristath:


carolinan:
This solves the button text color, but also highlights a second problem with the button background color, which will need a separate issue.

carolinan (approved):

PR #277, opened 2020-10-05
Try: Overlapping Gallery Block Pattern

kjellr:
Implements the "Overlapping Gallery" pattern from #50. 

Due to the wide range of options possible here (number of images, number of columns, etc), this is a rough one to get right. This is the closest I got so far. 

- For now, it enforces two columns, no matter what the user selects in the sidebar. It's possible (but complicated) to support additional columned layouts, but it might help to see what that should look like. @melchoyce, is that something you can mock up? Gutenberg supports 1-8 columns.  
- I discovered a bug in Gutenberg, where if you point to theme assets (which don't come with an ID) for your gallery pattern, then the media library doesn't work. 😄 I'm going to file a bug for that shortly, but in the meantime, if you want to add/subtract images in one of these galleries, you should create a fresh one and apply the "Overlapping Gallery" block style to it. 
- Because the overlap can very easily make captions unusable, I hid them for now. 😕
- In general this could use lots of testing. 

## Screenshots

Editor Desktop|Editor Mobile
---|---
gutenberg test_wp-admin_post php_post=1963 action=edit (1)|gutenberg test_wp-admin_post php_post=1963 action=edit

Front End Desktop|Front End Mobile
---|---
gutenberg test__p=1963|gutenberg test__p=1963 (1)

Lots of Images:
gutenberg test__p=1963 (2)


melchoyce:
> For now, it enforces two columns, no matter what the user selects in the sidebar. It's possible (but complicated) to support additional columned layouts, but it might help to see what that should look like. @melchoyce, is that something you can mock up? Gutenberg supports 1-8 columns.

Can do 👍

melchoyce:
Tried three and four columns, and it's already starting to get a little chaotic: 

image

kjellr:
😄 Yeah... maybe for 3+ columns, the grid is a little more regular, but every now and then an image overlaps?

carolinan:
I think a pattern with 3 image blocks would be easier to manage

kjellr:
> I think a pattern with 3 image blocks would be easier to manage

It will include just three, but since it's just a regular gallery block, it'll need to support folks adding any number of items.

kjellr:
> I think a pattern with 3 image blocks would be easier to manage

Sorry — I see that I misunderstood your suggestion. 

Technically, the main benefit of using a gallery block is that we have full control of the grid container within that single block. Otherwise, we'd need some sort of other wrapper (columns block presumably), that would position the interior blocks. But since folks will be able to move and replace those blocks with other types of blocks, there's more potential for things to break. With a Gallery block, we always know what is going to be inside of it. 

Also, if someone did want to include 4 or 5 images, it's way easier for them to just add to the gallery block, versus more image blocks manually and trying to achieve the same overlapping layout on their own.

melchoyce:
Tried a version without the overlap, to see if we could simplify:

galleries

Definitely not nearly as compelling as the overlapped version.

I really wish we could get the flow more like https://www.marvinschwaibold.com/, http://www.danimorales-style.com/editorial, or https://bttrhlf.com/. Animation aside, the positioning and overlap on all of these feel very smooth and natural.

ItsJonQ:
Haii! I took a stab at attempting this masonry-like layout. I've needed to do something like this in the past (with the exception of the overlaps). This brought back memories...

Anywho! This task is really tricky with pure CSS 😭 (I'd argue, currently impossible with the current specs).

### 🍿 Demo
The closest solution I've seen is using the `column-count` trick.

I've assembled a Codepen here:
https://codepen.io/itsjonq-the-lessful/pen/WNxbxKB

I've added some variables at the top can be adjusted (for dev purposes) to try various columns/gaps.

Interestingly enough, it looks like it was something @jasmussen attempted in 2018.

### 😱 Issues

The issue with this is that it's ordered in a top->bottom kind of way. Like this:

```
[1] [4]
[2] [5]
[3]
```

Whatever column the image appears in is unpredictable, depending on browser and image dimensions.

Because of this, it may not always respect the column count value. For example, if there are many short images, and the column count is set to `5`, it may only ever expand 3 columns.

### 📐 CSS Grid?

CSS Grid can't fit (and wrap) content of varying sizes without manual intervention. By that, I mean you have to know (roughly) a head of time where you went inner elements to be placed.

### 💪 CSS Flexbox?

Flexbox has it's own issues, specifically with height. With varying image sizes, a flex solution will produce large gaps (in an undesirable kind of way).

### 😭 Why is this so hard?

It's mostly an issue of unpredictability (image sizes), with a hint of subjectivity on what "looks good".
There are some things that CSS can't do (bummer). One of them is knowing the size of things (which is key for making this work).

Popular solutions like Masonry only work because it calculates the sizes of every item (using JS), and applies dimension/layout adjustments as needed (with probably with **MATH**)

---

You folks can give the `column-count` solution a shot... but be aware that it may not be reliable.

jffng:
I tried out this PR as well as #370.

My thoughts:

- With this PR, it's really nice out of the box but falls apart immediately when you start to replace it with your own images or use the gallery block settings.
- IMO, what this block pattern is trying to achieve from would make more sense as a custom block, given what has been said about needing to know the number of images and dimensions, and locking down some control over the block in order to guarantee it works as expected.
- Despite some hiccups, #370 is easier to work with — it's clearer what the block style variation is doing, and despite some editor quirks, you can change the number of columns and it works e.g.

Screen Shot 2020-10-08 at 7 09 54 PM

(Please don't judge my Gutenberg design skills, I made this in ~20 seconds, do you recognize the images 😅 )

jasmussen:
2021 looks amazing, this PR with the overlapping gallery looks beautiful. The here looks pretty solid too (and impressive at that).

I would offer one perspective on the effort, which has nothing to do about the visuals (because they're nailed) and is purely about post-launch theme support. The more complicated the CSS is, the more likely it is to break due to either the refactoring of the gallery block to use innerblocks, the arrival of an iframe, or just annoying caption issues. My only advice there is: do favors to your future selves to keep things as simple as you can. Even if that means limiting the gorgeous overlapping effect only to the 3 column version, that might be worth it.

melchoyce:
Thanks all for your perspectives. I think we'll close this PR in favor of exploring #370.

mtias:
I think this is also one where the theme could get a nice update once a block or a pattern is capable of handling this layout well.

PR #279, opened 2020-10-05
Remove retina option for logo

ryelle:
Following up on #244, See #57.

This removes the retina option from the customizer, removes the conditional from the logo settings, and removes the now-unnecessary filter on `get_custom_logo`.

To test

- Add & remove the logo via Customizer
- It should say "Suggested image dimensions: 300 by 100 pixels."
- Use a logo larger than 300px x 100px
- The preview should immediately update with the new (or removed) logo
- Check that there is no retina setting
- Check that the logo is output correctly


melchoyce:
LGTM 👍 Thanks for the detailed testing instructions.

PR #281, opened 2020-10-05
Fix editor custom colors bug

jffng:
This PR fixes a small bug in the conditional for when to run custom color generation in the editor.

Previously, if you change the default color of your site, and then back to the default color, the script to generate and load custom colors would still run in the editor — causing slightly differences to the background color in the editor and front-end:

**Before**
Screen Shot 2020-10-05 at 12 58 59 PM

**After**
Screen Shot 2020-10-05 at 12 59 31 PM

To test:
- Change the background color of your site via the customizer to something from the default, then change it back to the default.
- Add a button to your site
- Verify that it uses the correct button background color

### DIFF
--- a/classes/class-twenty-twenty-one-custom-colors.php
+++ b/classes/class-twenty-twenty-one-custom-colors.php
@@ -105,7 +105,7 @@ class Twenty_Twenty_One_Custom_Colors {
 			array(),
 			(string) filemtime( get_theme_file_path( 'assets/css/custom-color-overrides.css' ) )
 		);
-		if ( 'D1E4DD' !== get_theme_mod( 'background_color', 'D1E4DD' ) ) {
+		if ( 'd1e4dd' !== strtolower( get_theme_mod( 'background_color', 'D1E4DD' ) ) ) {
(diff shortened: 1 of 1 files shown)


melchoyce:
Works for me 👍

classes/class-twenty-twenty-one-custom-colors.php
```diff
@@ -105,7 +105,7 @@ public function editor_custom_color_variables() {
 			array(),
 			(string) filemtime( get_theme_file_path( 'assets/css/custom-color-overrides.css' ) )
 		);
-		if ( 'D1E4DD' !== get_theme_mod( 'background_color', 'D1E4DD' ) ) {
+		if ( 'd1e4dd' !== strtolower( get_theme_mod( 'background_color', 'D1E4DD' ) ) ) {
```
melchoyce, line 108: Any reason the first hex code is lowercase now?

aristath: > Any reason the first hex code is lowercase now?

Since users can manually enter any color they want, they may enter lowercase, uppercase or mixed. Using lowercase here makes sense because we're passing the theme-mod through `strtolower` which will convert it to lowercase. This way the case used by users won't matter. It's more robust this way :+1:

PR #283, opened 2020-10-05
Pagination: Wrap pagination links on smaller screens

ryelle:
See #256 — Comment & post pagination links wrap around awkwardly on smaller screens. This PR fixes the wrapping issue by putting the "Older" & "Newer" links on their own lines. I've set the breakpoint to `tablet` because I was still seeing wrapping above the mobile size. Additionally, I noticed that the next/prev links are mislabelled — the "next page" is older content, and previous is newer.

Comments, small screen:
Screen Shot 2020-10-05 at 5 32 11 PM

Posts, small screen:
Screen Shot 2020-10-05 at 5 30 13 PM

Posts, large screen:
Screen Shot 2020-10-05 at 5 30 25 PM

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -227,19 +227,16 @@ if ( ! function_exists( 'twenty_twenty_one_the_posts_navigation' ) ) {
 			array(
 				'mid_size'  => 2,
 				'prev_text' => sprintf(
-					/* Translators: Left arrow */
 					'%s <span class="nav-prev-text">%s</span>',
 					is_rtl() ? twenty_twenty_one_get_icon_svg( 'arrow_right' ) : twenty_twenty_one_get_icon_svg( 'arrow_left' ),
-					__( 'Older posts', 'twentytwentyone' )
+					__( 'Newer posts', 'twentytwentyone' )
 				),
 				'next_text' => sprintf(
-					/* Translators: Right arrow */
 					'<span class="nav-next-text">%s</span> %s',
-					__( 'Newer posts', 'twentytwentyone' ),
+					__( 'Older posts', 'twentytwentyone' ),
 					is_rtl() ? twenty_twenty_one_get_icon_svg( 'arrow_left' ) : twenty_twenty_one_get_icon_svg( 'arrow_right' )
 				),
 			)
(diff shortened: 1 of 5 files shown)


inc/template-tags.php
```diff
@@ -227,19 +227,16 @@ function twenty_twenty_one_the_posts_navigation() {
 			array(
 				'mid_size'  => 2,
 				'prev_text' => sprintf(
-					/* Translators: Left arrow */
 					'%s <span class="nav-prev-text">%s</span>',
```
carolinan, line 230: Good catch on the link text! Aren't the comments link text in the wrong position too then?

Why was the translators comment removed?

ryelle: Because it doesn't actually apply to anything here— the placeholder is not translated, only the text is.

kjellr:
Thanks, @ryelle! 

I noticed that the hover state looks a little weird now that these are block elements. Can we can shorten that but still retain the break? 

Screen Shot 2020-10-06 at 10 23 12 AM

ryelle:
Yeah, I saw that with the focus states too — unfortunately these links are all just inline `a`s, so to make it large enough to wrap I've set it to be full width. Maybe some `:before` & `:after` magic can help… 🤔

ryelle:
Okay, that was an adventure of trying various CSS hacks (nope), to seeing if we could filter these links (we can't), to finally looking for a core ticket about changing the pagination links output… which lead me to [this same problem in Twenty Seventeen,](https://core.trac.wordpress.org/changeset/38996) which was solved by hiding the page numbers. So that's what I've got here now.

Posts:
Screen Shot 2020-10-06 at 7 43 02 PM
Screen Shot 2020-10-06 at 7 42 55 PM
Screen Shot 2020-10-06 at 7 42 40 PM

Comments:
Screen Shot 2020-10-06 at 7 44 42 PM
Screen Shot 2020-10-06 at 7 43 37 PM

kjellr (approved):
Thanks for sticking with this! This seems like a solid solution to me, and I like that there's precedent in 2017. 👍

carolinan (approved):

PR #286, opened 2020-10-06
Primary Menu: Fix alignment and scrolling on mobile

### LINKED ISSUE #189: Responsive menu: fix text overlapping and show site branding

ryelle:
Fixes #189 — The menu/close button now stays fixed at the top of the screen, while long menu items can be scrolled through. Menu items move under the button container, so the text doesn't overlap anymore. This also fixes the content shift described in #216. 

|           | Logged in | Not logged in |
|-----------|-----------|---------------|
| initial   | Screen Shot 2020-10-05 at 8 20 21 PM | Screen Shot 2020-10-05 at 8 16 21 PM |
| menu open | Screen Shot 2020-10-05 at 8 20 55 PM | Screen Shot 2020-10-05 at 8 16 30 PM |
| scrolled  | Screen Shot 2020-10-05 at 8 21 02 PM | Screen Shot 2020-10-05 at 8 17 52 PM |

⚠️  I couldn't get my test site to work in Firefox or Safari, so this has _only been tested in Chrome/Mac._  I'll try to fix my setup, but until then other folks testing in other browsers or actually on mobile devices would be a big help.

### DIFF
--- a/assets/sass/06-components/navigation.scss
+++ b/assets/sass/06-components/navigation.scss
@@ -399,5 +399,6 @@
 	.lock-scrolling .site {
 		position: fixed;
 		max-width: 100%;
+		width: 100%;
 	}
 }
--- a/assets/sass/06-components/navigation.scss
+++ b/assets/sass/06-components/navigation.scss
@@ -107,11 +107,10 @@
 		@include media(mobile-only) {
 			height: 100vh;
 			z-index: 499;
-		}
 
-		.admin-bar & {
-			height: calc(100vh - 46px);
-			padding-top: calc(3.5 * var(--global--spacing-unit) - 46px);
+			.admin-bar & {
+				height: calc(100vh - 46px);
+			}
 		}
 	}
 
--- a/assets/sass/06-components/navigation.scss
+++ b/assets/sass/06-components/navigation.scss
@@ -97,7 +97,11 @@
 		right: 0;
 		bottom: 0;
 		left: 0;
-		padding: calc(3.5 * var(--global--spacing-unit)) var(--global--spacing-unit) var(--global--spacing-horizontal);
+		// Height of the menu-button-container using font size, line height, and total padding, plus 5px so the focus of the first item is visible.
+		padding-top: calc(var(--button--line-height) * var(--primary-nav--font-size-button) + 42px + 5px);
+		padding-left: var(--global--spacing-unit);
+		padding-right: var(--global--spacing-unit);
+		padding-bottom: var(--global--spacing-horizontal);
 		background-color: var(--global--color-background);
 		overflow-x: scroll;
 		overflow-y: auto;
@@ -131,9 +135,7 @@
 	}
 
 	// Adjust positions when logged-in
-	.admin-bar &,
-	.admin-bar & > .primary-menu-container,
-	.admin-bar.lock-scrolling & .button {
+	.admin-bar & {
 		top: 46px;
 	}
 
--- a/assets/sass/06-components/navigation.scss
+++ b/assets/sass/06-components/navigation.scss
@@ -7,26 +7,27 @@
 	justify-content: space-between;
 	position: absolute;
 	right: 0;
-	padding-top: calc(1.4 * var(--global--spacing-vertical));
+	padding-top: calc(0.5 * var(--global--spacing-vertical));
 	padding-bottom: calc(0.25 * var(--global--spacing-vertical));
 
 	#primary-mobile-menu {
 		margin-left: auto;
 		padding: calc(var(--button--padding-vertical) - (0.25 * var(--global--spacing-unit))) calc(var(--button--padding-horizontal) - (0.25 * var(--global--spacing-unit)));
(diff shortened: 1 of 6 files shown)


carolinan:
I still see a visible shift where the content moves to the left when the button is clicked.
In this screenshot, I have removed the background color of the menu just to confirm:

The content below the menu shifts to the left

I am not sure if the devtools are affecting the menu and keyboard testing here in any way. It makes me wish that the responsive menu appeared on a wider width.

The scroll works well for me.
In my tests, a part of the focus outline on the first menu item is cut off, and it is cut off in various degrees in different browsers:

Chrome
menu item focus outline is cut off in Chrome

Firefox:
menu item focus outline is cut off in Firefox

assets/sass/06-components/navigation.scss
```diff
@@ -89,34 +88,41 @@
...
 		overflow-y: auto;
 		transition: all .15s ease-in-out;
 		transform: translateY(var(--global--spacing-vertical));
 
 		@include media(mobile-only) {
+			height: 100vh;
 			z-index: 499;
 		}
+
+		.admin-bar & {
```
carolinan, line 112: This is affecting the wide menu as well.

assets/sass/06-components/navigation.scss
```diff
@@ -89,34 +88,41 @@
...
+			height: calc(100vh - 46px);
```
carolinan: This is height is affecting the wide menu as well.

carolinan:
In Firefox on PC, an extra tab is needed to reach the first menu item.
I do not see it in any other browsers,  and not without this PR applied.

1) Tab to the menu button and open the menu
2) Press tab: - Focus is lost or is on an item that is not visible.
3) Press tab again: Focus is on the first menu item.

The same thing happens when you tab backwards between the first menu item and the close button.

aristath:
> In Firefox on PC, an extra tab is needed to reach the first menu item.

I can confirm this (Firefox on Ubuntu)

assets/sass/06-components/navigation.scss
ryelle: ah, thanks for catching that. Fixed in https://github.com/WordPress/twentytwentyone/pull/286/commits/32c420deb097acd8ea1b26db43c7c6b3ae5336ff

ryelle:
I've fixed the menu alignment so the top focus outline should not be cut off anymore. I had also misunderstood the alignment issue (I wasn't watching the disappearing content), but the transparent menu made that clear — so that's also fixed in https://github.com/WordPress/twentytwentyone/pull/286/commits/8a4eaea412657e05cad11e1f10440b9698eeb5d7.

carolinan:
Looks good to me, I feel that this can be merged, I am no closer to finding why Firefox has an extra tab, perhaps it is JS related.

ryelle:
Good news: I fixed my local environment & was able to test in Firefox
Bad news: The tab stop issue exists on trunk too, and appears to be [behavior specific to Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=616594) when there's `overflow: auto` with scrolling content:

> it's the only way for a keyboard-only user to scroll it (by tabbing to it and then using arrow keys and the like).

(The element focused is `primary-menu-container`)

We might want to add an outline style to the menu just to make the active element more obvious, but that can be another issue.

carolinan:
what about reverting the overflow:auto?

ryelle:
It looks like this happens with any overflow property, which we can't remove because it's needed to allow scrolling with a mouse.

jffng:
I tried this out on my mobile device (iPhone 11) and here's what I'm getting:
 
Chrome | Safari
------- | -------
IMG_AF7AA954D077-1 | IMG_4FDC6A9E6A88-1

The menus won't open in either browser — I'm not sure if this is related to my tunnel, but I can upload the PR to a different test site and check.

carolinan:
Tested the PR on my Huawei phone in both orientations on Firefox, Chrome, Opera mini, Edge.
On ipad in Chrome, Safari. Edge and Firefox.  The menu and the buttons work as expected. 

I'd like to see it paired with https://github.com/WordPress/twentytwentyone/pull/290, either on trunk or if trunk can be merged back into this PR, if you want to do more testing.

ryelle:
I've rebased the branch so it should have that change— in a quick look I don't think there would be any interaction, but if there is I can fix it in (my) morning.

carolinan:
Let's merge this and make a note to retest on Iphone.

PR #292, opened 2020-10-06
Solve dotted underline issue in Chrome

carolinan:
Props to @pattonwebz 

Before and after:
dot

dot-after

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -377,6 +377,7 @@ a {
 
 a:hover {
 	text-decoration-style: dotted;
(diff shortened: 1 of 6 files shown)


pattonwebz (approved):
LGTM :+1: :shipit:

For reference here is the [blink render bug](https://bugs.chromium.org/p/chromium/issues/detail?id=808603) that causes the problem. It is marked as _fixed_ but who knows when it's going to be rolled out in chrome.

melchoyce:
I think Chromium 87: https://twitter.com/argyleink/status/1309172888243712002?s=20

Issue #232, opened 2020-10-02
Option to hide the author’s biography


briceduclos:
With Twenty Twenty One, the author's biography with its avatar always appears on single posts. Some users might want to hide it.
Twenty Twenty has an option inside the Customizer to hide the biography while Seedlet & Twenty Twenty Nineteen only displays it if the user has entered a description:
`if ( (bool) get_the_author_meta( 'description' ) ) :`

melchoyce:
> Seedlet & Twenty Twenty Nineteen only displays it if the user has entered a description

This approach makes the most sense to me — let's add this in.

pattonwebz:
I can take on the change needed for this then unless @briceduclos wants to take care of it in a PR before me :)

carolinan:
With the profile removed, the post has no other author name, I don't think that is ideal for example when there is more than one author.

Design guidance is needed if we want to place the name else where, in the footer meta information for example.

melchoyce:
I'll mock something up tomorrow 👍

melchoyce:
Two ideas:

![image](https://user-images.githubusercontent.com/2846578/95123808-86644c00-0720-11eb-954d-4d7a58ccc693.png)

carolinan:
👋 @briceduclos what is your WordPress.org username, for the contributors list?

briceduclos:
@carolinan My WordPress.org username is bduclos.

PR #294, opened 2020-10-06
Add author name to entry footer

carolinan:
Follow up on https://github.com/WordPress/twentytwentyone/issues/232

Adds author name and link to the entry footer if the author has no biography.

A new div is added so that the left area of the entry footer can span two grid rows, preventing the tags from being pushed down, causing a gap.

Edge case with many tags or categories:
entry-footer

The CSS for the entry footer is not ideal and could use a complete rewrite by someone who is better at sass, when there is time.

Figma:
https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=1394%3A58


inc/template-tags.php
```diff
@@ -40,13 +40,15 @@ function twenty_twenty_one_posted_on() {
...
-			/* translators: 1: post author, only visible to screen readers. 2: author link. 3: author name*/
-			'<span class="byline"><span class="screen-reader-text">%1$s</span><span class="author vcard"><a class="url fn n" href="%2$s">%3$s</a></span></span>',
-			esc_html__( 'Posted by', 'twentytwentyone' ),
-			esc_url( get_author_posts_url( get_the_author_meta( 'ID' ) ) ),
-			esc_html( get_the_author() )
-		);
+		if ( ! get_the_author_meta( 'description' ) ) {
+			printf(
+				/* translators: 2: author link. 3: author name*/
+				'<span class="byline">%1$s <span class="author vcard"><a class="url fn n" href="%2$s">%3$s</a></span></span>',
```
ryelle: This translator comment is not necessary, since the string with placeholders isn't translated.

inc/template-tags.php
```diff
@@ -40,13 +40,15 @@ function twenty_twenty_one_posted_on() {
...
+				esc_html__( 'By', 'twentytwentyone' ),
```
ryelle: It would be more i18n-friendly to use a placeholder here, like this snippet from twentytwenty:
```php
printf(
	/* translators: %s: Author name. */
	__( 'By %s', 'twentytwenty' ),
	'<a href="' . esc_url( get_author_posts_url( get_the_author_meta( 'ID' ) ) ) . '">' . esc_html( get_the_author_meta( 'display_name' ) ) . '</a>'
);
```

ryelle: Here, this works:

```php
echo '<span class="byline">';
printf(
	/* translators: %s author name. */
	esc_html__( 'By %s', 'twentytwentyone' ),
	'<span class="author vcard"><a class="url fn n" href="' . esc_url( get_author_posts_url( get_the_author_meta( 'ID' ) ) ) . '">' . esc_html( get_the_author() ) . '</a></span>'
);
echo '</span>';
```

ryelle (changes requested):
If you want to use grid without the wrapper, you could assign each item to the specific row/column (IIRC you'll need to do this for IE11 grid support, unless there's another fallback). That leaves a bunch of gaps though:

Screen Shot 2020-10-07 at 11 56 02 AM

Screen Shot 2020-10-07 at 11 57 34 AM

I think the CSS solution here works 👍

SResok (commented):
CR Done, nice usage of the coding standards 👍🏼

inc/template-tags.php
```diff
@@ -115,42 +117,40 @@ function twenty_twenty_one_entry_meta_footer() {
...
-					/* translators: %1$s: posted in label, %2$s: list of tags. */
-					'<span class="tags-links">%1$s %2$s</span>',
-					esc_html__( 'Tagged', 'twentytwentyone' ),
-					$tags_list // phpcs:ignore WordPress.Security.EscapeOutput
-				);
-			}
-		} elseif ( is_single() ) {
-			// Posted on.
-			twenty_twenty_one_posted_on();
+				echo '<div class="post-taxonomies">';
```
ryelle, line 117: Wrapping this changed the design a bit, so that categories & tags are now on separate lines in archive views, is that intentional?

| Trunk | This branch |
|-------|-------------|
| Screen Shot 2020-10-09 at 10 39 15 AM | Screen Shot 2020-10-09 at 10 31 13 AM |

ryelle (commented):
This is working well across the posts/pages I checked, I just had one note about the design change.

I also noticed this doesn't handle small screens, but that can be done in a follow-up PR.

footer-small

inc/template-tags.php
carolinan: They should be on the same line.

GuidoOffermans:
code looks good! This still has to be acceptance-tested.

carolinan:
The spacing needs improvement:
Archive: There is no space between the last category and the "Tagged:" text.  
There is a space in the linked post date, so the link underline looks longer than it should be.
Single: There is no space between the last category and the "Tagged:" text.

In the archive, the categories and tags are meant to be on the same line, but in the single view, they are meant to be on separate lines.

SResok:
@carolinan the last couple of commits should've fixed the spacing problems. I created a separate `<span>` element with the 'Updated' date

inc/template-tags.php
```diff
@@ -17,9 +17,6 @@
 	 */
 	function twenty_twenty_one_posted_on() {
 		$time_string = '<time class="entry-date published updated" datetime="%1$s">%2$s</time>';
-		if ( get_the_time( 'U' ) !== get_the_modified_time( 'U' ) ) {
-			$time_string = '<time class="entry-date published" datetime="%1$s">%2$s</time> <time class="updated" datetime="%3$s">%4$s</time>';
-		}
 
 		$time_string = sprintf(
```
ryelle, line 21: Since the second half of `$time_string` is not used anymore, you don't need to pass the last two args here (`get_the_modified_date`) - there will always only be two placeholders.

inc/template-tags.php
```diff
@@ -115,39 +131,40 @@ function twenty_twenty_one_entry_meta_footer() {
...
+
+				/* translators: used between list items, there is a space after the comma. */
+				$categories_list = get_the_category_list( __( ', ', 'twentytwentyone' ) );
+				if ( $categories_list ) {
+					printf(
+						/* translators: 1: posted in label, 2: list of categories. */
+						'<span class="cat-links">%1$s %2$s. </span>',
+						esc_html__( 'Categorized as', 'twentytwentyone' ),
+						$categories_list // phpcs:ignore WordPress.Security.EscapeOutput
+					);
```
ryelle: It looks like there were some merge issues here, and the change from #423 was lost.

```suggestion
					printf(
						/* translators: %s: list of categories. */
						'<span class="cat-links">' . esc_html__( 'Categorized as %s', 'twentytwentyone' ) . '</span>',
						$categories_list // phpcs:ignore WordPress.Security.EscapeOutput
					);
```

inc/template-tags.php
```diff
@@ -115,39 +131,40 @@ function twenty_twenty_one_entry_meta_footer() {
...
-				);
+				/* translators: used between list items, there is a space after the comma. */
+				$tags_list = get_the_tag_list( '', __( ', ', 'twentytwentyone' ) );
+				if ( $tags_list ) {
+					printf(
+						/* translators: %1$s: posted in label, %2$s: list of tags. */
+						'<span class="tags-links">%1$s %2$s.</span>',
+						esc_html__( 'Tagged', 'twentytwentyone' ),
+						$tags_list // phpcs:ignore WordPress.Security.EscapeOutput
+					);
```
ryelle: Same as above, this lost the string change. It should be 

```suggestion
					printf(
						/* translators: %s: list of tags. */
						'<span class="tags-links">' . esc_html__( 'Tagged %s', 'twentytwentyone' ) . '</span>',
						$tags_list // phpcs:ignore WordPress.Security.EscapeOutput
					);
```

inc/template-tags.php
```diff
@@ -165,15 +182,34 @@ function twenty_twenty_one_entry_meta_footer() {
...
-				);
+				/* translators: used between list items, there is a space after the comma. */
+				$tags_list = get_the_tag_list( '', __( ', ', 'twentytwentyone' ) );
+				if ( $tags_list ) {
+					printf(
+						/* translators: 1: posted in label, 2: list of tags. */
+						'<span class="tags-links">%1$s %2$s</span>',
+						esc_html__( 'Tagged', 'twentytwentyone' ),
+						$tags_list // phpcs:ignore WordPress.Security.EscapeOutput
+					);
```
ryelle: Same as above

```suggestion
					printf(
						/* translators: %s: list of tags. */
						'<span class="tags-links">' . esc_html__( 'Tagged %s', 'twentytwentyone' ) . '</span>',
						$tags_list // phpcs:ignore WordPress.Security.EscapeOutput
					);
```

inc/template-tags.php
```diff
@@ -165,15 +182,34 @@ function twenty_twenty_one_entry_meta_footer() {
...
+
+				/* translators: used between list items, there is a space after the comma. */
+				$categories_list = get_the_category_list( __( ', ', 'twentytwentyone' ) );
+				if ( $categories_list ) {
+					printf(
+						/* translators: 1: posted in label, 2: list of categories. */
+						'<span class="cat-links">%1$s %2$s </span>',
+						esc_html__( 'Categorized as', 'twentytwentyone' ),
+						$categories_list // phpcs:ignore WordPress.Security.EscapeOutput
+					);
```
ryelle: Same as above

```suggestion
					printf(
						/* translators: %s: list of categories. */
						'<span class="cat-links">' . esc_html__( 'Categorized as %s', 'twentytwentyone' ) . '</span>',
						$tags_list // phpcs:ignore WordPress.Security.EscapeOutput
					);
```

ryelle (changes requested):
Visually looks good on posts by authors without a bio
Screen Shot 2020-10-13 at 5 30 53 PM

But something strange is happening on authors with bios — the date & edit link have no space.

Screen Shot 2020-10-13 at 5 31 32 PM

melchoyce:
Also noticed that this PR adds the "Updated..." meta. Can we remove that?

SResok:
> Also noticed that this PR adds the "Updated..." meta. Can we remove that?

@melchoyce Removed it! 

> Since the second half of `$time_string` is not used anymore, you don't need to pass the last two args here (`get_the_modified_date`) - there will always only be two placeholders.

@ryelle Updated the time_string and implemented the suggested changes 👍🏼

ryelle (approved):
Looks good now! 👍

PR #295, opened 2020-10-06
Adds styles on the editor for alignleft/alignright. Fixes #282

### LINKED ISSUE #282: Update design of floated pullquotes
Moving this over from #229:

> Do we want to limit the width of the left and right aligned pullquote to 50%? When the content is long enough, the quotes are no longer aligned but looks like they are normal width without alignments.
>
> In the editor, left and right aligned pull quotes needs spacing to the right and left of the block respectively.
>
> It is less visible on the right:

image

aristath:
Fixes #282 - Adds styles on the editor for alignleft/alignright

### DIFF
--- a/assets/sass/05-blocks/utilities/_editor.scss
+++ b/assets/sass/05-blocks/utilities/_editor.scss
@@ -179,3 +179,15 @@
 	margin: 0;
 	margin-left: var(--global--spacing-horizontal);
 }
+
+@media only screen and (min-width: 482px) {
+	.wp-block[data-align="left"] > * {
+		max-width: 290px;
+		margin-right: var(--global--spacing-horizontal);
+	}
+
+	.wp-block[data-align="right"] > * {
+		max-width: 290px;
+		margin-left: var(--global--spacing-horizontal);
+	}
+}
--- a/assets/sass/05-blocks/utilities/_editor.scss
+++ b/assets/sass/05-blocks/utilities/_editor.scss
@@ -181,6 +181,7 @@
 }
 
 @media only screen and (min-width: 482px) {
+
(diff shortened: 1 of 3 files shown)


assets/sass/05-blocks/utilities/_editor.scss
```diff
@@ -179,3 +179,16 @@
 	margin: 0;
 	margin-left: var(--global--spacing-horizontal);
 }
+
+@media only screen and (min-width: 482px) {
+
+	.wp-block[data-align="left"] > * {
+		max-width: 290px;
```
kjellr, line 186: Can we use a variable here, rather than hardcoding this width?

carolinan:
Looks good to me, tested several block types.

carolinan (approved):

assets/sass/05-blocks/utilities/_editor.scss
aristath: I tried the same same calc we use on the front-end but there were lots of issues with it. Hardcoding a value seemed like a reasonable compromise.
We don't have any related vars that I could see, and adding a new one just for the purpose of using it here didn't seem right

kjellr: Maybe we use something based on the [sass variable](https://github.com/WordPress/twentytwentyone/blob/07eab795ee6c80c662293d800081af22c39facac/assets/sass/03-generic/breakpoints.scss#L9) then? Something like `#{0.5 * $default_width}`.

carolinan:
Because we need the alignments for testing the blocks, I am going to merge this without the CSS variable change.

Issue #270, opened 2020-10-05
Columns block spacing


carolinan:
**Describe the bug**
Compare the screenshots below, the first is a columns block with mixed content in the editor, 
and the second is the same content on the front.

On the front, there is a left and right padding added to the columns block, is that intended?

The group block with the black background, seen in the center, is full width.
This seems to remove the blocks vertical spacing in the editor, but not on the front.

![columns-editor](https://user-images.githubusercontent.com/7422055/95071281-afc5ad80-0709-11eb-9d81-8f7bb23b4d2f.png)

![columns-front](https://user-images.githubusercontent.com/7422055/95071762-76da0880-070a-11eb-9400-caac90dd26f7.png)

carolinan:
After a conversation on Slack, it was decided that the left and right padding on the outside of the full width columns block should be removed.

carolinan:
As a follow up to 297, paragraphs and headings inside the first and last column needs to have padding so that there is space between the text and the edge of the browser window.

Are there any other blocks that needs this padding?

JustinyAhin:
> As a follow up to 297, paragraphs and headings inside the first and last column needs to have padding so that there is space between the text and the edge of the browser window.
> 
> Are there any other blocks that needs this padding?

I think all blocks except list block (which already have a padding) would need this padding.

I'm going to test all of them and create PRs if necessary.

JustinyAhin:
> I think all blocks except list block (which already have a padding) would need this padding.

I finally think all the blocks should have this padding. As this should not really depend on whether they have padding or not.

PR according to that coming in a couple of minutes.

carolinan:
But for example an image or cover block should not have spacing between the block and the edge of the browser.
Paragraph blocks with background colors already have padding, it should not be duplicated.

JustinyAhin:
Make sense.

JustinyAhin:
So here are the blocks that should not receive this padding:

- Paragraph with background
- List
- Image cover
- Image
- Media and text

All good @carolinan?

carolinan:
looks correct to me but it has been a long day =D

JustinyAhin:
> looks correct to me but it has been a long day =D

Agree. And it looks more complex actually. With blocks as pullquote or blockquote where the padding should also apply to the borders, etc...

I'll continue working on it tomorrow.

PR #297, opened 2020-10-06
Remove padding from full-width column block in frontend

JustinyAhin:
This PR fix the issue #270.

w.org: justinahinon


kjellr (approved):
Thank you!

PR #298, opened 2020-10-06
Add clearing after the widget area

### LINKED ISSUE #271: Add clearing after the last block in the widget area
**Describe the bug**
If the last item in the widget area has a left or right alignment, it overlaps the next element.
Admittedly, I only tested with image and the social icon block. The quick and easy solution would be to add clear: both to the site info.

widget-clear

**To Reproduce**
1. Enable Gutenberg 9.1. Add a block in the widget screen and set the alignment to left or right.
2. View the front.
3. You may need to reduce the window width to see the problem.

**Expected behavior**
I expect the items to not overlap.

JustinyAhin:
Fix #271.

This will need to be discussed to decide if it is an acceptable solution to the issue (see https://github.com/WordPress/twentytwentyone/issues/271#issuecomment-704381075)

w.org: justinahinon.

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -5563,6 +5563,12 @@ h1.page-title {
 	margin-top: 30px;
 }
 
+.widget-area:after {
+	content: "";
+	display: table;
+	clear: both;
+}
+
 .widget-title {
 	font-size: 1.125rem;
 	font-weight: 700;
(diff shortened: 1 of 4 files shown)


JustinyAhin:
A failing test on .scss file. Fix coming in a few moments.

aristath (approved):
LGTM :+1:

Issue #278, opened 2020-10-05
Improving the Separator Block - Gutenberg issues/PRs


paaljoachim:
**Is your feature request related to a problem? Please describe.**
It would be nice to get improvements/options into the Separator Block as it could help improve the design.

![Screen Shot 2020-10-05 at 20 44 05](https://user-images.githubusercontent.com/5323259/95119001-8f1e4780-074b-11eb-9353-88c136aaaebb.png)

Existing Gutenberg issues/PR's:
**Separator Block: Add support for wide/full-width versions**
https://github.com/WordPress/gutenberg/issues/25080 -> https://github.com/WordPress/gutenberg/pull/25147
(The above has just been merged into Gutenberg.)

**Separator: allow changing the thickness / height**
https://github.com/WordPress/gutenberg/issues/20758 -> https://github.com/WordPress/gutenberg/pull/23403

**Separator: move color tools to the toolbar**
https://github.com/WordPress/gutenberg/issues/20757

**Blocks: Separator: Background style is too specific**
https://github.com/WordPress/gutenberg/issues/19539

**Add better variations for Separator block**
https://github.com/WordPress/gutenberg/issues/16483

Associated:
**Add Shape Divider block**
https://github.com/WordPress/gutenberg/issues/16351

melchoyce:
![image](https://user-images.githubusercontent.com/2846578/95121713-6bdca380-071d-11eb-8038-455b8724c86a.png)

[Figma mockup](https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=1394%3A0)

Unsure about the dots — right now it's just using the • • • characters. There's more potential for an interesting divider here. @ellenbauer do you have any ideas?

paaljoachim:
I have made two new Separator block issues in the Gutenberg repo:
Adding styles such as waves, jagged and loops.
https://github.com/WordPress/gutenberg/issues/25991

Controlling the height of the Separator block
https://github.com/WordPress/gutenberg/issues/25989
(As one would not need to add as many spacer blocks if one can control the height of the Separator block.)

carolinan:
With the pattern merged, are there any more actionable items here or can this be closed?

paaljoachim:
Since I have made Gutenberg issues. Which I have listed. This issue might as well be closed.

PR #299, opened 2020-10-06
Try: Adding "Portfolio List" block pattern + modifying "wide" separator style

kjellr:
Would address part of #50. 

This is a first pass at achieving the "Portfolio List" block pattern, using just what's available today in Gutenberg. 

Original Mockup|Current State
---|---
93506798-c2a74800-f8ea-11ea-9994-8a44ba402e33-1|Screen Shot 2020-10-06 at 1 54 16 PM

A few caveats:

- Using default Gutenberg tools, It's not possible today to easily inline an image next to the text. There is an inline image option, but it's not suitable for use in a Block Pattern because it does not include any way to replace the image. Instead, I've placed the images at the right for now. 
- I've modified the "Wide" separator style so that it includes the thicker stroke. I'm unable to make it wide-width for now though, because of the structure of the block in the editor (wide alignment styles would need to be applied to the parent of the separator block, and CSS isn't good about that).  We'll need to wait for https://github.com/WordPress/gutenberg/pull/25147. In the meantime, this pattern is the default width instead of the wide width. 😕
- While we work through this, I used only images that we've added to the repo already. 
- Mobile views of this look _ok_, but not great.

### Screenshots

Editor

Screen Shot 2020-10-06 at 1 41 47 PM

Desktop|Mobile
---|---
Screen Shot 2020-10-06 at 1 40 27 PM|Screen Shot 2020-10-06 at 1 41 03 PM


melchoyce:
Man, we are kind of striking out with these patterns. 

The separators + the link underlines make this feel really busy. What if the top and bottom separators were 3px, but the in-between ones were 1px instead?

image

It's not perfect, but it helps. Hopefully we'll be able to do try wide-width separators soon as well, because I think that would also help.

Since we can't use inline images, it might also help to float the images all the way to the right, and then cap all their heights at 67px:

image

Since they have different aspect ratios, this makes for a nice ragged edge.

kjellr:
> What if the top and bottom separators were 3px, but the in-between ones were 1px instead?

Good suggestion — I can try that out tomorrow. 

> it might also help to float the images all the way to the right

I tried that initially actually, but changed them back because it led to this on mobile. 😕

95241300-8d9c6000-07db-11eb-9259-6d4df6874275

melchoyce:
Can we wrap in a group block with a custom class, and tweak the mobile positioning?

kjellr:
We _can_, but I'm not sure that we should. Using custom classnames in these seems like a hack to me. For simplicity's sake, I consider patterns to be something that the user should be able to just create themselves without having to open the "Advanced" panel and type in a custom classname. I could be being too conservative about them though.

kjellr:
> The separators + the link underlines make this feel really busy. What if the top and bottom separators were 3px, but the in-between ones were 1px instead?

I pushed an update to take care of this. In order to make that happen, I made the default separator 100% wide and 1px tall. That's more or less in line with the comps [here](https://github.com/WordPress/twentytwentyone/issues/278#issuecomment-703834149) anyway. 

Screen Shot 2020-10-07 at 9 45 19 AM

> it might also help to float the images all the way to the right, and then cap all their heights at 67px:

Mobile issues aside, I tried this too. But it highlighted a couple other issues. 

- The floated images mess with the equal spacing. 😕
- Also, it seems like our alignleft/alignright styles are imposing some sort of (minimum?) enforced width, so I can't get that ragged edge working. 

Screen Shot 2020-10-07 at 9 50 15 AM

melchoyce:
Weird, the screenshot from above is from my editor, so recent updates must have changed things. Would #295 by @aristath fix the width issue?

kjellr:
I got the right alignment + rag sorted out. I had to explicitly set both the width and height of each image.

melchoyce:
Do we want to merge this and make a new issue to handle the responsive styles?

kjellr:
I think that's reasonable, as long as folks are cool with the changes to the separator block included here. Here's a quick reference: 


Before:

Screen Shot 2020-10-07 at 1 24 35 PM

After:

Screen Shot 2020-10-07 at 1 25 01 PM

melchoyce:
Good with me, let's just keep an eye on https://github.com/WordPress/gutenberg/pull/25147 and maybe revisit 👍

carolinan (approved):

PR #301, opened 2020-10-06
Ignore changes to .css files in diffs

jdevalk:
This change should make GitHub ignore the changes to CSS files when you're looking at pull requests and also make it ignore CSS for code stats purposes.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,7 +2,7 @@
 
 *.php text eol=lf
 *.js text eol=lf
-*.css text eol=lf
+*.css text eol=lf linguist-generated
 *.scss text eol=lf
 *.json text eol=lf
 


ryelle:
It looks like `linguist-generated` is a property [used by github to ignore these files,](https://github.com/github/linguist#generated-code) so this would only apply to the PRs on github.

My initial question is whether someone could make a change to the CSS directly & we wouldn't see it in the PR, but even if they do, the Build action would run & overwrite the CSS file with the correct generated file. So I think this is a good idea to add.

Does this change how the files show up in the github file browser? Ex: https://github.com/WordPress/twentytwentyone/blob/trunk/style.css

aristath:
> Does this change how the files show up in the github file browser? Ex: https://github.com/WordPress/twentytwentyone/blob/trunk/style.css

No, it won't affect that. But it will definitely improve PR reviews a lot!

aristath (approved):

PR #303, opened 2020-10-06
Update the left arrow SVG

ryelle:
The previous SVG was slightly but noticeably different, so this takes the right-arrow & flips it.

Before:
Screen Shot 2020-10-06 at 7 51 20 PM

After:
Screen Shot 2020-10-06 at 7 51 33 PM


melchoyce:
Thanks for fixing this! Can you pull the left arrow up 1px and push the right arrow down 1px?

carolinan:
I'm glad someone else noticed this too, it was starting to bug me 😂

Do we need to report this upstream to Gutenberg? That’s where the icons are from.

PR #306, opened 2020-10-07
remove margin-right on item-has-children navigation.scss

mager19:
Added margin-right:0, remove margin right - .menu-item-has-children - Navigation.scss


assets/sass/06-components/navigation.scss
```diff
@@ -375,6 +375,7 @@
 
 	// Show top-level sub-menu indicators above mobile-breakpoint-only
 	.menu-item-has-children {
+		margin-right: 0 !important;
```
carolinan, line 378: It is best practice to avoid using !important.

I suggest using specificity instead, and placing

`			&.menu-item-has-children {
				margin-right: 0;
			}`

Inside this code starting on line 283:
```
.primary-menu-container > ul > .menu-item {
		@include media(mobile) {
			....
		}
	}
```

(Github removes the formatting here, follow the coding standards :) )

carolinan:
@mager19  A different solution will be used to solve this issue. But I will of course add you to the contributors list.

PR #317, opened 2020-10-07
Firefox: Indicate focus on the scrollable area of the responsive menu

carolinan:
Follow up to
https://github.com/WordPress/twentytwentyone/pull/286#issuecomment-704345280

Firefox issue: _overflow:auto causes DIV to get tab focus even if it has no scrollbars_
https://bugzilla.mozilla.org/show_bug.cgi?id=616594

This PR adds a one sided border to the primary-menu-container of the mobile menu when it is in focus.

I tested using an outline with different offsets, but because the outline was overlapping the scrollbars, I thought that a less distracting one sided border would work better.

When the focus is moved one step to the button or the menu item, the border is no longer visible.

focus-indicator-firefox


aristath (approved):
LGTM :+1: 

Code-wise it makes sense.
The only thing that troubles me here is the `:focus` on a `div`.
Can a `div` even have `focus`? I believe that `:focus` styles are only applied to links, buttons and other actionable elements.
Should we maybe add `:focus-within` ? :thinking:

carolinan:
I am not sure how to interpret the firefox bug :(

ryelle:
> Can a div even have focus?

Yep, in this case it can— this is a special case in Firefox where the div is scrollable, so FF adds a tab stop/focus abilities to it. `:focus-within` actually would not work, because the focus is on the div itself.

melchoyce:
Can we do something more like this?

image

(5px inset on each side)

carolinan:
I was not able to get such an inset to work because the line overlaps both the vertical and horisontal scrollbar

carolinan:
Outline offset -4:
offset4

Outline offset:-14
offset14


Outline offset -18
offset18

carolinan:
Box shadow:
`box-shadow: 0 0 0 1px var(--global--color-primary) inset, -18px -18px 0 1px var(--global--color-primary) inset;`

boxshadow

melchoyce:
Browsers 😩 😩 

Let's forget about the offset, then, and have the border flush with the edge of the screen:

image

Will this work? I'm okay with a little overlap on the very edge of the scrollbar, it's just really weird when it's like, in the middle of the scrollbar.

carolinan:
Lets wait with this until the header part of the menu is finalized again.

carolinan:
End result:

result

carolinan:
Since this design is already approved, I will merge it.

PR #325, opened 2020-10-07
Add new action before closing header

### LINKED ISSUE #320: Consider adding a hook inside the header
It would be nice to have a hook just before the end of the site header so developers can add additional elements (for example, the [Seedlet theme](https://github.com/Automattic/themes/blob/master/seedlet/header.php#L68) adds a WooCommerce cart and a socials menu inside the header):
```
<?php do_action( 'twentytwentyone_header' );
</header><!-- #masthead -->.
```

abhijitrakas:
Closes: https://github.com/WordPress/twentytwentyone/issues/320

### DIFF
--- a/header.php
+++ b/header.php
@@ -52,7 +52,15 @@
 					);
 					?>
 				</nav><!-- #site-navigation -->
-				<?php } ?>
+			<?php } ?>
+			<?php
+			/**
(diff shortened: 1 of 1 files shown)


carolinan:
During Friday's bug scrub it was decided that the theme would not have any extra hooks.

PR #329, opened 2020-10-07
Refactor Navigation spacing. Fixes #302

### LINKED ISSUE #302: Navigation Menu: Remove extra right margin on dropdown items
When a navigation menu item has a dropdown, there's extra space added for the + sign, in addition to the `margin-right` on the list item:

image

Could we remove that margin for all navigation items with dropdowns (`.menu-item-has-children`), so the spacing feels a little more even?

image

aristath:
In this PR:

* Removed the padding from `> li` elements and added that padding to `> li > a` elements. This has the benefit that there's a larger hover/click area which improves a11y.
* Removed both margins from `li`
* Added negative margin to the nav wrapper to offset the constant padding in `a` elements for the beginning and the end of menu's alignments.
* repositioned the dropwdown buttons a bit so they're not too far from their menu.


carolinan:
Happy to see the padding removed from the container because it was adding a horizontal scroll on mobile. 
I will not be able to test this until tomorrow.

melchoyce:
Seems good! Two notes:

Any way to keep the focus state on open menu items from getting cut off?
image

There's a little bit of a visual jump between the + and - here:
menu-hover

The plus feels like it's one pixel down and to the right of the minus. Any way to smooth that out a bit so the transition feels more seamless?

aristath:
> Any way to keep the focus state on open menu items from getting cut off?

Good catch! Done in [`aef0b44`](https://github.com/WordPress/twentytwentyone/pull/329/commits/aef0b4459a47e4021a647b8c069363fd02c53f76)

> The plus feels like it's one pixel down and to the right of the minus. Any way to smooth that out a bit so the transition feels more seamless?


I hadn't noticed that :+1: 
The problem seems to have been with the minus icon's SVG path, it was not properly centered.
Changed it from this:
```xml
<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24' width='18' height='18'><path d='M4 9h12v2H4V9z' fill='%2328303d'/></svg>
```
to this:
```xml
<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24' width='18' height='18'><path d='M6 11h12v2H6z' fill='%2328303d'/></svg>
```
(the only difference is the `path`)
Looks OK now:

Peek 2020-10-08 08-42

carolinan:
In Chrome, the outline of the parent menu item is still cut off by one pixel at the bottom when the submenu is open.

aristath:
> In Chrome, the outline of the parent menu item is still cut off by one pixel at the bottom when the submenu is open.

Hmmm I don't see it... :thinking: 
Screenshot from 2020-10-08 11-53-51

assets/sass/06-components/navigation.scss
```diff
@@ -263,7 +263,7 @@
 				margin: 0;
 				min-width: max-content;
 				position: absolute;
-				top: 100%;
+				top: calc(100% + 2px);
```
carolinan:
```suggestion
				top: calc(100% + 3px);
```
Changing this to 3px solves the outline visibility issue in Chrome and I see no unexpected side effects in other browsers.

PR #336, opened 2020-10-07
Try fixing page margins when editor sidebar is open

kjellr:
Twenty Twenty-One's responsive width variables work great in all cases except when the editor sidebar is open. When that's the case, they still work pretty well, except that wide ones tend to expand all the way to the edges of the screen, removing the comfortable margins that are supposed to exist around the content. 

This PR is a quick attempt to solve that problem. In doing so, it creates a new problem but I'm sharing it in case someone else has an idea for how to actually fix this. 😄 

In general: 

- Theoretically, switching to `%` units instead of `vw` should work fine for non-nested blocks in the editor and the front end. It also significantly improves the appearance of the page on desktop views where the sidebar is expanded. 
- Child blocks do need to use `vw`, since they need to refer to the width of the parent container. 
- In general, making this change seems to work fine. Except for the "Mobile" editor preview, in which case it looks terrible. 

## Screenshots

**Editor: Desktop, Sidebar Open** 
🟢 Improved.

Before|After
---|---
 Screen Shot 2020-10-07 at 1 52 48 PM|Screen Shot 2020-10-07 at 1 50 37 PM

---

**Editor: Desktop, Sidebar Closed**
🟡 Almost Identical — This PR makes the wide width slightly less wide. Not a huge deal.

Before|After
---|---
Screen Shot 2020-10-07 at 1 52 54 PM|Screen Shot 2020-10-07 at 1 50 41 PM

---

**Editor: Desktop, Mobile Preview** 
🔴 Definitely not identical. 😞

Before|After
---|---
Screen Shot 2020-10-07 at 1 59 53 PM|Screen Shot 2020-10-07 at 2 00 09 PM

---

**Editor: Mobile**
🟢 Identical

Before|After
---|---
Screen Shot 2020-10-07 at 1 51 06 PM|Screen Shot 2020-10-07 at 1 53 08 PM

---

**Front End: Desktop**
🟢 Identical

Before|After
---|---
Screen Shot 2020-10-07 at 1 53 15 PM|Screen Shot 2020-10-07 at 1 51 31 PM

---

**Front End: Mobile**
🟢 Identical

Before|After
---|---
Screen Shot 2020-10-07 at 1 53 24 PM|Screen Shot 2020-10-07 at 1 52 20 PM


kjellr:
I noticed this does indeed break the width of some child blocks. 😕

Open to suggestions for other solutions. If this solution turns out to be way off or unworkable, we can close.

carolinan:
On chrome on PC, I can not reproduce the error width the wide blocks, shown in the very first screenshot. And because we don't have time to retest nested blocks, I vote for this to be closed.

kjellr:
It's visible (in my testing) in Chrome between 782px and ~1540px breakpoints. I'll keep this open today and see if I can come up with a better working solution. Otherwise, I'll close and just file an issue to track it for later.

kjellr:
I wasn't able to get this working better, so I'm going to close and re-think it. I opened #395 to track the issue.

PR #340, opened 2020-10-08
Resets site-branding margin if last/only child. Fixes #307

### LINKED ISSUE #307: Mobile/responsive site title spans half the width when there is no menu
**Describe the bug**
The primary navigation is optional. When there is no menu, and the browser window is reduced to the size where the button would show, the site title and tagline is still spanning about half the page, instead of the full width.

**To Reproduce**
1. View a page with no menu but with a site title
2. Reduce the window width

**Expected behavior**
I expected the text to span the full width

**Screenshots**

responsive

glauberm:


glauberm:
#307

ryelle (approved):
Looks good! Thanks for fixing this.

ryelle:
Dropping in a screenshot for reference:

Screen Shot 2020-10-08 at 12 02 26 PM

PR #341, opened 2020-10-08
Remove class and use function instead

### LINKED ISSUE #313: Remove the `Twenty_Twenty_One_Non_Latin_Languages` class
Spawned from #312

The `Twenty_Twenty_One_Non_Latin_Languages` object contains a single static method.
It can just as easily be a function instead, there's absolutely no point in keeping it as a class.

abhijitrakas:
Closes: https://github.com/WordPress/twentytwentyone/issues/313


aristath (approved):
As a first pass this looks OK.

I don't know if we'll be able to further simplify it, but this one is OK to merge :+1:

PR #346, opened 2020-10-08
Fatal error: Remove Non-latin language handling file from function.php which is already deleted

kishanjasani:
This included file was deleted but not removed from function.php and is generating a fatal error.


kishanjasani:
@carolinan Please have a look. I think It's urgent to fix this issue.

carolinan:
Thank you, this has already been solved.

PR #348, opened 2020-10-08
Add padding to blocks that are in columns edges

JustinyAhin:
Follow up #270.

This is a first step to resolve the issue mentioned in this comment #270 (comment).


carolinan:
I will test the scss changes, I think trunk needs to be merged into this PR to hide the PHP changes from earlier.

carolinan:
When testing I realized the issue description is not correct, I'm sorry for that.

In the test I added a 3 column block with a paragraph in each.
When the browser window width is reduced, and the column is vertical, the text in the middle columns also needs spacing.

The same blocks that has the added padding on the front, should also have the padding in the editor so that the editor and front matches.


My test content: https://gist.github.com/carolinan/55f91835e9b3b977704a643c71bd0800

JustinyAhin:
> The same blocks that has the added padding on the front, should also have the padding in the editor so that the editor and front matches.

Great. Will look back at it tomorrow.

PR #349, opened 2020-10-08
Update style-rtl.css

aditi572:
update style-rtl.css


aristath (changes requested):
These styles get auto-generated so if any changes have to be made they should be done in the scss files instead.

Also, I don't see anything valid in that PR... the font-weights are wrong, the hex color codes are invalid. Was this PR done by mistake?

aditi572:
I'm new to this so it might have been a mistake.

PR #354, opened 2020-10-08
Excerpt on postspage toggle

### LINKED ISSUE #129: Add option to use excerpts on posts page
This is basically the Twenty Twenty One version of https://github.com/WordPress/twentytwenty/issues/2.

Twenty Twenty has an option to use excerpts on the default posts page. This option is currently absent from Twenty Twenty One. I think it would be a good idea to bring back this option. In fact, I'd prefer if excerpts were used by default. The inability to change this has been a major missing feature in previous default themes, in my opinion. The default behavior of showing the entire content of every post on the home page has never made sense to me.

rolf-yoast:
Fixes https://github.com/WordPress/twentytwentyone/issues/129

## Summary

Add a section named `Theme settings` where you can toggle between excerpt or full text. Excerpt is set as default.

n/a

## Test instructions

1. Make sure you have a posts page set in `Settings` -> `Reading`
1. Make sure you have posts with long enough of a content so that you can see the difference between excerpts and full text.
1. Go to `Appearance` -> `Customize`
1. Go to `Theme settings` and change the `On the Posts page, post show:` settings
1. Check on the Posts page if the content switch between showing excerpts or full text.

Don't forget to test the unhappy-paths!

## Quality assurance

### DIFF
--- a/classes/class-twenty-twenty-one-customize.php
+++ b/classes/class-twenty-twenty-one-customize.php
@@ -97,6 +97,9 @@ if ( ! class_exists( 'Twenty_Twenty_One_Customize' ) ) {
 				array(
 					'capability' => 'edit_theme_options',
 					'default'    => 'excerpt',
+					'sanitize_callback' => function( $value ) {
+						return 'excerpt' === $value || 'full' === $value ? $value : 'excerpt'
+					},
 				)
 			);
 
--- a/classes/class-twenty-twenty-one-customize.php
+++ b/classes/class-twenty-twenty-one-customize.php
@@ -88,7 +88,7 @@ if ( ! class_exists( 'Twenty_Twenty_One_Customize' ) ) {
 			$wp_customize->add_section(
 				'theme_settings',
 				array(
-					'title'    => __( 'Theme settings', 'twentytwentyone' ),
+					'title'    => esc_html__( 'Theme settings', 'twentytwentyone' ),
 					'priority' => 30,
 				)
 			);
--- a/classes/class-twenty-twenty-one-customize.php
+++ b/classes/class-twenty-twenty-one-customize.php
@@ -95,8 +95,8 @@ if ( ! class_exists( 'Twenty_Twenty_One_Customize' ) ) {
 			$wp_customize->add_setting(
 				'display_excerpt_or_fullpost',
 				array(
-					'capability' => 'edit_theme_options',
-					'default'    => 'excerpt',
+					'capability'        => 'edit_theme_options',
+					'default'           => 'excerpt',
 					'sanitize_callback' => function( $value ) {
 						return 'excerpt' === $value || 'full' === $value ? $value : 'excerpt';
 					},
--- a/classes/class-twenty-twenty-one-customize.php
+++ b/classes/class-twenty-twenty-one-customize.php
@@ -85,17 +85,20 @@ if ( ! class_exists( 'Twenty_Twenty_One_Customize' ) ) {
 			 * Add excerpt or full text selector to customizer
 			 */
 
-			$wp_customize->add_section( 'theme_settings' , array(
-				'title'      => __('Theme settings','twentytwentyone'),
-				'priority'   => 30,
-			) );
+			$wp_customize->add_section(
+				'theme_settings',
+				array(
+					'title'    => __( 'Theme settings', 'twentytwentyone' ),
+					'priority' => 30,
+				)
+			);
 
 
 			$wp_customize->add_setting(
 				'display_excerpt_or_fullpost',
 				array(
-					'capability'        => 'edit_theme_options',
-					'default'           => 'excerpt',
+					'capability' => 'edit_theme_options',
+					'default'    => 'excerpt',
 				)
 			);
 
@@ -107,7 +110,7 @@ if ( ! class_exists( 'Twenty_Twenty_One_Customize' ) ) {
 					'label'   => __( 'On the Posts page, post show:', 'twentytwentyone' ),
 					'choices' => array(
 						'excerpt' => 'Excerpt',
-						'full' 	  => 'Full text',
+						'full'    => 'Full text',
 					),
 				)
 			);
--- a/classes/class-twenty-twenty-one-customize.php
+++ b/classes/class-twenty-twenty-one-customize.php
@@ -107,7 +107,7 @@ if ( ! class_exists( 'Twenty_Twenty_One_Customize' ) ) {
 				array(
 					'type'    => 'radio',
 					'section' => 'theme_settings',
-					'label'   => __( 'On the Posts page, post show:', 'twentytwentyone' ),
+					'label'   => esc_html__( 'On the Posts page, post show:', 'twentytwentyone' ),
 					'choices' => array(
 						'excerpt' => 'Excerpt',
 						'full'    => 'Full text',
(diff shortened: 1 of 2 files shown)


classes/class-twenty-twenty-one-customize.php
```diff
@@ -81,6 +81,40 @@ public function register( $wp_customize ) {
...
 			);
 
+			/**
+			 * Add excerpt or full text selector to customizer
+			 */
+
+			$wp_customize->add_section(
+				'theme_settings',
+				array(
+					'title'    => __( 'Theme settings', 'twentytwentyone' ),
```
aristath: Can we use `esc_html__` instead of `__` here?
Since this will be a core theme everything needs to be escaped

```suggestion
					'title'    => esc_html__( 'Theme settings', 'twentytwentyone' ),
```

classes/class-twenty-twenty-one-customize.php
```diff
@@ -81,6 +81,40 @@ public function register( $wp_customize ) {
...
+					'default'    => 'excerpt',
+				)
+			);
+
+			$wp_customize->add_control(
+				'display_excerpt_or_fullpost',
+				array(
+					'type'    => 'radio',
+					'section' => 'theme_settings',
+					'label'   => __( 'On the Posts page, post show:', 'twentytwentyone' ),
```
aristath: `esc_html__` instead of `__` - same as line 91

```suggestion
					'label'   => esc_html__( 'On the Posts page, post show:', 'twentytwentyone' ),
```

index.php
```diff
@@ -21,7 +21,12 @@
 	// Load posts loop.
 	while ( have_posts() ) {
 		the_post();
-		get_template_part( 'template-parts/content/content' );
+
+		if ( get_theme_mod( 'display_excerpt_or_fullpost', 'excerpt' ) === 'excerpt' ) {
```
aristath: Unless the WordPress Coding Standards changed, this should probably use Yoda syntax

```suggestion
		if ( 'excerpt' === get_theme_mod( 'display_excerpt_or_fullpost', 'excerpt' ) ) {
```

aristath (changes requested):

index.php
rolf-yoast: Sorry. Used to non Yoda syntax.

index.php
```diff
@@ -21,7 +21,12 @@
 	// Load posts loop.
 	while ( have_posts() ) {
 		the_post();
-		get_template_part( 'template-parts/content/content' );
+
+		if ( 'excerpt' === get_theme_mod( 'display_excerpt_or_fullpost', 'excerpt' ) ) {
+			get_template_part( 'template-parts/content/content-excerpt' );
+		} else {
+			get_template_part( 'template-parts/content/content' );
+		}
```
aristath, line 29: Hmmm thinking about this one some more, would it perhaps be better written like this instead?

```php
get_template_part(
	'template-parts/content/content-excerpt',
	'excerpt' === get_theme_mod( 'display_excerpt_or_fullpost', 'excerpt' ) ? 'excerpt : ''
);
```

aristath: Scratch that, what you have there is more readable :+1:

aristath (approved):
Tested & confirmed. Code is good, functionality is good. :+1:

classes/class-twenty-twenty-one-customize.php
```diff
@@ -81,6 +81,38 @@ public function register( $wp_customize ) {
...
+			);
+
+			$wp_customize->add_control(
+				'display_excerpt_or_fullpost',
+				array(
+					'type'    => 'radio',
+					'section' => 'theme_settings',
+					'label'   => esc_html__( 'On the Posts page, post show:', 'twentytwentyone' ),
+					'choices' => array(
+						'excerpt' => 'Excerpt',
```
carolinan: The text for the choices needs to be translation ready and escaped.

aristath: Good point, thank you @carolinan for catching that! :+1: 
Note: should use `esc_html__`

classes/class-twenty-twenty-one-customize.php
```diff
@@ -81,6 +81,38 @@ public function register( $wp_customize ) {
...
+			 */
+			$wp_customize->add_section(
+				'theme_settings',
+				array(
+					'title'    => esc_html__( 'Theme settings', 'twentytwentyone' ),
+					'priority' => 30,
+				)
+			);
+
+			$wp_customize->add_setting(
```
carolinan, line 95: The setting needs to have a sanitize_callback. For the radio option, we need to validate the data.

A new function needs to be added, but you can refer to this example: 
https://github.com/WPTT/code-examples/blob/master/customizer/sanitization-callbacks.php#L278

classes/class-twenty-twenty-one-customize.php
aristath:
```suggestion
						'excerpt' => esc_html__( 'Excerpt', 'twentytwentyone' ),
```

classes/class-twenty-twenty-one-customize.php
```diff
@@ -81,6 +81,38 @@ public function register( $wp_customize ) {
...
+
+			$wp_customize->add_control(
+				'display_excerpt_or_fullpost',
+				array(
+					'type'    => 'radio',
+					'section' => 'theme_settings',
+					'label'   => esc_html__( 'On the Posts page, post show:', 'twentytwentyone' ),
+					'choices' => array(
+						'excerpt' => 'Excerpt',
+						'full'    => 'Full text',
```
aristath:
```suggestion
						'full'    => esc_html__( 'Full text', 'twentytwentyone' ),
```

classes/class-twenty-twenty-one-customize.php
```diff
@@ -81,6 +81,38 @@ public function register( $wp_customize ) {
...
+					'title'    => esc_html__( 'Theme settings', 'twentytwentyone' ),
+					'priority' => 30,
+				)
+			);
+
+			$wp_customize->add_setting(
+				'display_excerpt_or_fullpost',
+				array(
+					'capability' => 'edit_theme_options',
+					'default'    => 'excerpt',
```
aristath: For the `sanitize_callback` mentioned by @carolinan above (see https://github.com/WordPress/twentytwentyone/pull/354/files#r501613550), it could be as simple as this:
```suggestion
					'capability' => 'edit_theme_options',
					'default'    => 'excerpt',
					'sanitize_callback' => function( $value ) {
						return 'excerpt' === $value || 'full' === $value ? $value : 'excerpt'
					},
```

aristath (changes requested):

classes/class-twenty-twenty-one-customize.php
```diff
@@ -81,6 +81,41 @@ public function register( $wp_customize ) {
...
+			$wp_customize->add_section(
+				'theme_settings',
+				array(
+					'title'    => esc_html__( 'Theme settings', 'twentytwentyone' ),
+					'priority' => 30,
+				)
+			);
+
+			$wp_customize->add_setting(
+				'display_excerpt_or_fullpost',
```
ZebulanStanphill, line 96: Shouldn't this be `display_excerpt_or_full_post`?

aristath: Thank you @ZebulanStanphill for the review :+1: 
Fixed in https://github.com/WordPress/twentytwentyone/pull/379

PR #356, opened 2020-10-08
Hide page title on frontpage.

### LINKED ISSUE #314: Hide the page title on the front page
In general, default themes are designed to be as little opinionated as possible. Creating more unique websites is becoming easier now having several blocks and more extensive possibilities to style them. But this freedom is still more or less limited to the ‘post content‘. We have not yet a lot of styling possibilities for the page title.

Therefore, could we consider add an option to the Customizer to hide the `.entry-header` **for the front page only**? We should consider whether

1. we hide it visually to have an `h1` or
2. don‘t display it and make the `.site-title` an `h1` on the front page as well (it is now on the home page only).

This would give the user more freedom for designing a unique front page with blocks. See some quick examples below.

Screenshot 2020-10-07 at 09 55 02

Screenshot 2020-10-07 at 09 41 52

Screenshot 2020-10-07 at 09 48 40

I realise that it is rather easy to add custom CSS to hide it (which I end up doing for several websites I create), but that can ‘break’ the page structure if done incorrectly or it is simply not ‘just something everyone is able to do’.

GuidoOffermans:
Fixes #314 

## Summary

This PR hides the `template-parts/header/entry-header` if it's the front page.
- Implemented an if statement that does this check. 
- A bit of Sass that hides the border.

## Test instructions
1. Set your front-page as a static page.
1. Set a post title.
1. Navigate to your front-page and you should not be able to see the title.
Screenshot 2020-10-08 at 11 48 27

Don't forget to test the unhappy-paths!
## Quality assurance

### DIFF
--- a/template-parts/content/content-page.php
+++ b/template-parts/content/content-page.php
@@ -14,8 +14,13 @@
 <article id="post-<?php the_ID(); ?>" <?php post_class(); ?>>
 
 	<header class="entry-header alignwide">
-		<?php get_template_part( 'template-parts/header/entry-header' ); ?>
-		<?php twenty_twenty_one_post_thumbnail(); ?>
+		<?php
+		if ( ! is_front_page() ) {
+			get_template_part( 'template-parts/header/entry-header' );
+		}
+
+			twenty_twenty_one_post_thumbnail();
+		?>
 	</header>
 
 	<div class="entry-content">
--- a/template-parts/content/content-page.php
+++ b/template-parts/content/content-page.php
@@ -14,13 +14,10 @@
 <article id="post-<?php the_ID(); ?>" <?php post_class(); ?>>
 
 	<header class="entry-header alignwide">
-		<?php
-		if ( ! is_front_page() ) {
-			get_template_part( 'template-parts/header/entry-header' );
-		}
-
-			twenty_twenty_one_post_thumbnail();
-		?>
+		<?php if ( ! is_front_page() ) : ?>
+			<?php get_template_part( 'template-parts/header/entry-header' ); ?>
+		<?php endif; ?>
+		<?php twenty_twenty_one_post_thumbnail(); ?>
 	</header>
 
 	<div class="entry-content">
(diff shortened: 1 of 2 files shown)


template-parts/content/content-page.php
```diff
@@ -14,8 +14,13 @@
...
 	<header class="entry-header alignwide">
-		<?php get_template_part( 'template-parts/header/entry-header' ); ?>
-		<?php twenty_twenty_one_post_thumbnail(); ?>
+		<?php
+		if ( ! is_front_page() ) {
+			get_template_part( 'template-parts/header/entry-header' );
+		}
+
+			twenty_twenty_one_post_thumbnail();
+		?>
```
aristath: Templates should use the alternative syntax for conditions (see suggestion below)

```suggestion
		<?php if ( ! is_front_page() ) : ?>
			<?php get_template_part( 'template-parts/header/entry-header' ); ?>
		<?php endif; ?>
		<?php twenty_twenty_one_post_thumbnail(); ?>
```

aristath (changes requested):

template-parts/content/content-page.php
GuidoOffermans: @aristath Thankyou for pointing this out, we were not aware of this. Is there some documentation on this subject?

aristath: Unfortunately no... It's not a hard requirement just a convention & best-practice.
Template files should keep the logic to a minimum, and contain mostly HTML. In case logic is contained (like for example conditions like this one) it's traditional to not do it in PHP blocks but instead using a more HTML-like formatting so that people with no PHP experience don't get lost. That way it's easier for them to copy-paste things in a child theme and tweak things with minimal effort. If on the other hand PHP opening & closing tags are on separate lines, someone who doesn't know PHP can easily skip the closing tag and wonder why their site fails. Keeping things as self-contained as humanly possible makes it simple for non-developers to work on some things.

carolinan:
Tested with and without featured image, looks good to me.

aristath (approved):

Issue #332, opened 2020-10-07
I18N: Read more link might be hard to translate correctly


ryelle:
The "read more" link is built up out of parts to visibly show "Read more", with screen reader text identifying the post, "about [post name]". The way this string is built, each part is translated, but 'read more' and 'about' are translated separately. I think this would create an issue when translating the strings, if the word order is different than English — but I'm not sure, so I wanted to flag this as a possible issue.

https://github.com/WordPress/twentytwentyone/blob/dfd6b2bad2c3f87fd11a47a0482c0ecfd85c1a70/inc/template-functions.php#L141-L145

```
<a class="more-link" href="#">Read more <span class="screen-reader-text">about [Post Name]</span></a>
```

carolinan:
What you bring up makes sense, and why I prefer _continue reading_.

Sorry for all the edits, I am more tired than I'd like =D

ryelle:
This was discussed in [today's bug scrub](https://wordpress.slack.com/archives/C02RP4VMP/p1602261216351500) and it sounds like the best way forward will be to update this string so that it becomes:

```html
<a class="more-link" href="#">Continue reading <span class="screen-reader-text">[Post Name]</span></a>
```

sabernhardt:
Both [Twenty Seventeen](https://themes.trac.wordpress.org/browser/twentyseventeen/2.4/template-parts/post/content.php#L54) and [Twenty Nineteen](https://themes.trac.wordpress.org/browser/twentynineteen/1.7/template-parts/content/content.php#L32) have
`Continue reading<span class="screen-reader-text"> "%s"</span>`

(space inside the span tag)

ryelle:
>Both Twenty Seventeen and Twenty Nineteen have
> `Continue reading<span class="screen-reader-text"> "%s"</span>`
>
> (space inside the span tag)

Apparently the only reason that works is because the post title is wrapped in quotes — NVDA is still trying to read `Continue Reading"Post Name"`. The space before the `span` doesn't seem to be adding any visual awkwardness, so I think it's fine to keep it (and I have, in #433).

PR #361, opened 2020-10-08
Add div around read more block

### LINKED ISSUE #206: The read more block is not contained inside the content area

rolf-yoast:
Fixes https://github.com/WordPress/twentytwentyone/issues/206

## Summary

Places the read more link in the same alignment as the content

## Test instructions

1. Add a read more to a post
1. Check if it aligns with the content

## Quality assurance

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -201,7 +201,7 @@ function twenty_twenty_one_continue_reading_link() {
 			the_title( '<span class="screen-reader-text">' . esc_html__( 'about ', 'twentytwentyone' ), '</span>', false )
 		);
 
-		return '&hellip; <a class="more-link" href="' . esc_url( get_permalink() ) . '">' . $continue_reading . '</a>';
+		return '<div><a class="more-link" href="' . esc_url( get_permalink() ) . '">' . $continue_reading . '</a></div>';
 	}
 }
 
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -208,12 +208,6 @@ function twenty_twenty_one_continue_reading_text() {
 function twenty_twenty_one_continue_reading_link_excerpt() {
 
 	if ( ! is_admin() ) {
-		$continue_reading = sprintf(
-			/* translators: %s: Name of current post. */
-			wp_kses( esc_html__( 'Read more %s', 'twentytwentyone' ), array( 'span' => array( 'class' => array() ) ) ),
-			the_title( '<span class="screen-reader-text">' . esc_html__( 'about ', 'twentytwentyone' ), '</span>', false )
-		);
-
 		return '&hellip; <a class="more-link" href="' . esc_url( get_permalink() ) . '">' . twenty_twenty_one_continue_reading_text() . '</a>';
 	}
 }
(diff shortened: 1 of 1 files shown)


carolinan (approved):
This works well, but I wonder if it is worth adding a CSS class to the div so that it can be customized further. Even if this theme does not use the class, it can be useful for others.

carolinan:
@melchoyce This is the style with this PR:

readmore

inc/template-functions.php
```diff
@@ -206,11 +206,26 @@ function twenty_twenty_one_continue_reading_link() {
...
+function twenty_twenty_one_continue_reading_link() {
 
+	if ( ! is_admin() ) {
+		$continue_reading = sprintf(
+			/* translators: %s: Name of current post. */
+			wp_kses( esc_html__( 'Read more %s', 'twentytwentyone' ), array( 'span' => array( 'class' => array() ) ) ),
+			the_title( '<span class="screen-reader-text">' . esc_html__( 'about ', 'twentytwentyone' ), '</span>', false )
+		);
+
+		return '<div><a class="more-link" href="' . esc_url( get_permalink() ) . '">' . $continue_reading . '</a></div>';
```
ryelle: It would be nice to see the link tag generated in another function, so it can be shared between the two `continue_reading` functions, just to ensure any string changes happen in both places.

carolinan:
Related: https://github.com/WordPress/twentytwentyone/issues/332

inc/template-functions.php
carolinan: I did not consider that and I agree it would be a better solution.

rolf-yoast: I've added a separate function for the text.

carolinan:
So the screen reader test found a bug. There is no pause at all between "read more" and about, so it reads like `read moreabout post title` So an extra space is needed. There already is one space in the translation after the about text.

inc/template-functions.php
```diff
@@ -189,29 +189,45 @@ function twenty_twenty_one_get_avatar_size() {
...
 }
 
+/**
+ * Creates continue reading text
+ */
+function twenty_twenty_one_continue_reading_text() {
+	$continue_reading = sprintf(
+		/* translators: %s: Name of current post. */
+		wp_kses( esc_html__( 'Read more %s', 'twentytwentyone' ), array( 'span' => array( 'class' => array() ) ) ),
+		the_title( '<span class="screen-reader-text">' . esc_html__( 'about ', 'twentytwentyone' ), '</span>', false )
```
carolinan:
```suggestion
		the_title( '<span class="screen-reader-text">&nbsp;' . esc_html__( 'about ', 'twentytwentyone' ), '</span>', false )
```

PR #362, opened 2020-10-08
Fix site title tag and link

### LINKED ISSUE #358: The site title on the blog page is not a link when a static homepage is set
**Describe the bug**
When a static page is set as the homepage, and a page is selected as the blog, the site title is not a link on the blog page

**To Reproduce**
1. Go to the customizer or the homepage setting and set pages to the homepage and blog.
2. Open the blog page, and see that the site title is not a link.

**Expected behavior**
The blog page is not the homepage so there needs to be a link.

LittleBigThing:
Fixes #358 and sets the site title to an `h1` (see #314)

The conditionals were changed in `site-branding.php` to have a site title as follows
- h1 and unlinked on the front page and on the blog page if it is the front page
- h1 and linked on the blog page (if it is not the front page)
- p and linked everywhere else


carolinan:
> @carolinan: Can I fix this by correcting this as well? After fixing #314, we now don't have an h1 title on the front page. I would >add that here, make the site title an unlinked h1 tag on the front page.
>
>A smaller issue is that now we have an empty header on the front page if there is no featured image.

I did not see your comment until now.

Yes, but lets also simplify it by removing the p tag and allow multiple H1's on the same page.

pattonwebz:
> allow multiple H1's on the same page

Do we really want to allow this? Isn't this an anti-pattern?

ryelle:
> Do we really want to allow this? Isn't this an anti-pattern?

It is, we really shouldn't do this.

> A smaller issue is that now we have an empty header on the front page if there is no featured image.

This is a different issue & can be fixed in another PR with some more conditionals or maybe re-working the logic here in general.

ryelle (commented):
This PR is working well, but it removes the home link on paged views (like `/page/2/ ` etc, when showing latest posts or when the front page has multiple pages using the page break block)

Was it an intentional decision to remove those?

aristath:
> Do we really want to allow this? Isn't this an anti-pattern?

> It is, we really shouldn't do this.

It was an anti-pattern pre-HTML5.
That is no longer the case... Multiple H1s are fine when they make sense. In this case it makes perfect sense, the site-name is a title and not a paragraph.
We're used to doing it as a paragraph because that's what we were doing 10 years ago, but screen readers, browsers and even search engines understand it perfectly fine and there are no issues whatsoever with it.

ryelle:
I was under the impression it's still an _accessibility_ anti-pattern. I'm sure search engines are fine with it, but I want to make sure that assorted assistive tech users will have a good experience. If you're saying that's not the case anymore then cool. (have an article about it? I'd love to know when it stopped being an issue)

carolinan:
My first instinct was also to only use one H1. But with the unlinking of the site title and the option to hide the title and tagline, the if statements became a mess that was near unreadable.
@aristath brought it up on Slack during the code review of https://github.com/WordPress/twentytwentyone/pull/267
And it made sense to do what we can to simplify the code.

W3: https://www.w3.org/WAI/tutorials/page-structure/headings/
Quote:
```
Exception for fixed page sections
In fixed sections of the page, for example in sidebars, 
the heading ranks should not change depending on the ranks in the content area. 
In those cases, consistency across pages is more important.
```
I would say the header is a fixed section. -Even if it is possible to not have a title, tagline, logo or menu... (phew)

W3 shares two valid examples, one where the H1 is the document heading and one where the H1 is the main content heading.


The most comprehensive and useful information I found was from https://dequeuniversity.com/ 
but it is locked behind a payed course so I can't link to it. They mention an exception where a blog index page can use H1's for all articles because it is easier to maintain the markup, but the example does not strictly match our scenario.

Free material:
https://dequeuniversity.com/rules/axe/4.0/page-has-heading-one

_With that, lets test the actual pull request._

carolinan:
Found one more link that may be more relevant.

It is a screen reader survey where

- 60% of the respondents says that they prefer: One first level heading that contains the document title. _(not the site name)_
- 33% prefers: Two first level headings, one for the site name and one for the document title.

https://webaim.org/projects/screenreadersurvey7/#heading

aristath:
Also worth noting: In FSE sites by design the site-title will always be a header and won't change to a paragraph depending on the page's context.

carolinan:
### With the title enabled:
Single posts, single pages, attachment page, archives, search, 404 all have one H1.
The site title is a p tag.

#### When the blog is the homepage:
The paged blog (When you click "older posts" and end up on /page/2/) has one H1 but is missing a link on the site title. This link needs to be added.

The homepage has one H1 and no link and this is correct.

#### When a static page is the homepage:
The homepage has one H1 and no site title link and this is correct.
The blog page has one H1 and a site title link and this is correct.
When a page with page breaks is the homepage, the pages (/page/2/) is missing a link on the site title. This link needs to be added.

carolinan:
### With the title hidden by the customizer option

Single pages, single posts, attachment page, archive, search, 404 has one H1, the page title. The site branding div is present but empty.

#### When the blog is the homepage:
The homepage has one H1 which is visually hidden. This is correct.
The paged blog has one H1 which is visually hidden. This is correct.

#### When a static page is the homepage:
The homepage has no H1. This must be added. The site branding div is present but empty.
When a page with page breaks is the homepage, the pages (/page/2/) is missing an H1. This must be added.

The blog page has one H1 which is visually hidden. This is correct.

LittleBigThing:
I am thinking. ;-)

carolinan:
I did not get far with testing this because when **Display Site Title & Tagline** is checked, the title does not show. There is an empty link tag.

template-parts/header/site-branding.php
```diff
@@ -18,20 +18,16 @@
...
-	<?php if ( ! empty( $blog_info ) && get_theme_mod( 'display_title_and_tagline', true ) === true ) : ?>
-		<?php if ( is_home() ) : ?>
-			<?php if ( is_paged() ) : ?>
-				<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></h1>
-			<?php else : ?>
-				<h1 class="site-title"><?php bloginfo( 'name' ); ?></h1>
-			<?php endif; ?>
-		<?php elseif ( is_front_page() ) : ?>
-			<p class="site-title"><?php bloginfo( 'name' ); ?></p>
+	<?php if ( $blog_info && get_theme_mod( 'display_title_and_tagline', true ) === true ) : ?>
```
aristath: This should use a Yoda syntax, see [WP Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#yoda-conditions)

```suggestion
	<?php if ( $blog_info && true === get_theme_mod( 'display_title_and_tagline', true ) ) : ?>
```

template-parts/header/site-branding.php
```diff
@@ -18,20 +18,16 @@
...
+		<?php if ( is_front_page() ) : ?>
+			<h1 class="site-title"><?php $blog_info; ?></h1>
+		<?php elseif ( is_home() ) : ?>
+			<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php $blog_info; ?></a></h1>
 		<?php else : ?>
-			<p class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></p>
+			<p class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php $blog_info; ?></a></p>
```
aristath: Why was this line changed? `<?php $blog_info; ?>` doesn't doanything :thinking: 
I assume we want to echo it

```suggestion
			<p class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php echo esc_html( $blog_info ); ?></a></p>
```

template-parts/header/site-branding.php
```diff
@@ -18,20 +18,16 @@
...
 		<?php endif; ?>
-	<?php elseif ( ! empty( $blog_info ) && is_home() ) : ?>
-		<h1 class="screen-reader-text"><?php bloginfo( 'name' ); ?></h1>
+	<?php elseif ( $blog_info && is_home() ) : ?>
+		<h1 class="screen-reader-text"><?php $blog_info; ?></h1>
```
aristath: Same as line 27

```suggestion
		<h1 class="screen-reader-text"><?php echo esc_html( $blog_info ); ?></h1>
```

template-parts/header/site-branding.php
```diff
@@ -18,20 +18,16 @@
...
-			<?php if ( is_paged() ) : ?>
-				<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php bloginfo( 'name' ); ?></a></h1>
-			<?php else : ?>
-				<h1 class="site-title"><?php bloginfo( 'name' ); ?></h1>
-			<?php endif; ?>
-		<?php elseif ( is_front_page() ) : ?>
-			<p class="site-title"><?php bloginfo( 'name' ); ?></p>
+	<?php if ( $blog_info && get_theme_mod( 'display_title_and_tagline', true ) === true ) : ?>
+		<?php if ( is_front_page() ) : ?>
+			<h1 class="site-title"><?php $blog_info; ?></h1>
```
aristath: Same as line 27
```suggestion
			<h1 class="site-title"><?php echo esc_html( $blog_info ); ?></h1>
```

template-parts/header/site-branding.php
```diff
@@ -18,20 +18,16 @@
...
+		<?php elseif ( is_home() ) : ?>
+			<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php $blog_info; ?></a></h1>
```
aristath: Same as line 27
```suggestion
			<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php echo esc_html( $blog_info ); ?></a></h1>
```

aristath (changes requested):

LittleBigThing:
I think it's fixed. Sorry for the chaos of temporary ‘fixes‘. :-)
Is a screen reader text OK as simple as it is right know (when the site title is hidden), only to have an `h1`?

ryelle (changes requested):
This is working correctly for non-screen-reader views 👍 

> Is a screen reader text OK as simple as it is right know (when the site title is hidden), only to have an h1?

hmm, I think you'd want the same functionality for screen reader & non-SR users, so the fact that they don't have a home link & loose the site-name context on interior pages seems wrong.

What if we just swapped the class when hiding the site title?

```php
$header_class = ( true === get_theme_mod( 'display_title_and_tagline', true ) ) ? 'site-title' : 'screen-reader-text';
```

and then you don't need to check that later, so the block here can be simplified a bit:

```php
<?php if ( $blog_info ) : ?>
	<?php if ( is_front_page() && ! is_paged() ) : ?>
		<h1 class="<?php echo esc_attr( $header_class ); ?>"><?php echo esc_html( $blog_info ); ?></h1>
	<?php elseif ( is_front_page() || is_home() ) : ?>
		<h1 class="<?php echo esc_attr( $header_class ); ?>"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php echo esc_html( $blog_info ); ?></a></h1>
	<?php else : ?>
		<p class="<?php echo esc_attr( $header_class ); ?>"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php echo esc_html( $blog_info ); ?></a></p>
	<?php endif; ?>
<?php endif; ?>
```

LittleBigThing:
Not much left to change, but there is still a little fix for the description in the PR.

melchoyce:
Looks like this was fixed in #435; let's open a new PR if anything else needs updating. Thanks @LittleBigThing!

PR #365, opened 2020-10-08
Fix menu icon fill colors on dark backgrounds

### LINKED ISSUE #360: Plus and minus icon color does not change when the body background color is changed.
**Describe the bug**

When the body background color is changed, the plus and minus icons in the menu does not change color.
The becomes a problem when the icon is dark over dark backgrounds.

**To Reproduce**
1. Set up a menu with sub menu items.
2. Change the body background color.
3. See that the text color changes, but that the icons do not.

**Expected behavior**
I expect them to change color so that the color contrast remains high enough.
(The color should be white when the text color is set to white by the color option.)

**Screenshots**
The icon for the sub menu toggle does not have enough contrast against the background

aristath:
Fixes #360 

## Summary

Since the icons are on the `background` of buttons, the only way to tweak them was to set a different icon when the background-color is dark (can't use CSS `'fill`).
The only way that could happen is if we have a way to differentiate dark from light, so a body-class `is-dark-background` gets added when needed.
This PR adds the class in `<body>` both in PHP and in the customizer when changing the background-color.
After that was done, the CSS was tweaked to accommodate the white icons on dark backgrounds.

## Test instructions

1. Go to the customizer > colors, and select a dark background. You should see the menu icons change color to white.
2. Save, exit the customizer and verify on the frontend.
3. Go back to the customizer and select a light background. You should see the menu icons change color to dark.
4. Save, exit the customizer and verify on the frontend.

## Quality assurance


carolinan (approved):
Looks good to me and I don't see any new unexpected problems with other colors.

PR #370, opened 2020-10-08
Add overlapping images block pattern

kjellr:
This PR adds the "Overlapping Gallery" pattern from #50, but calls it "Overlapping Images" to better represent the blocks used in it. 

It's a way simpler alternative to #277. The exact overlap here is a little different from the comps, so I'm interested in your thoughts, @melchoyce. Feel free to recommend any adjustments. It also loses the overlap completely on mobile, which is a little unfortunate. This is because of the way the columns block stacks its columns by default on mobile. We could change that, but I imagine we'd run into some issues. 

Mobile|760px|1024px|1400px
---|---|---|---
gutenberg test__p=1 (4)|gutenberg test__p=1|gutenberg test__p=1 (1)|gutenberg test__p=1

~^ Note that there looks to be a small bug in the tablet (`760px`) breakpoint. I'm going to dig into that, but its unrelated to this PR, so I think it can be handled separately.~ _EDIT: Fixed in #372_


melchoyce (approved):
This one's good from a design standpoint.

One positive I've realized about this approach is, since it's using individual image blocks, you'll be able to link to pages.

PR #372, opened 2020-10-08
Add flex-grow rule to overlapping columns block style.

kjellr:
Followup from #337. Adds a single `flex-grow: 1` rule to fix an issue where the columns did not take up the full width at tablet-size breakpoints (between 600-781px). 

 

Before|After
---|---
Screen Shot 2020-10-08 at 1 41 26 PM|Screen Shot 2020-10-08 at 1 39 33 PM


melchoyce (approved):
Looks good to me in Chrome, Firefox, and Safari 👍

aristath (approved):
LGTM :+1:

PR #381, opened 2020-10-09
Fix gap between menu and submenu where the submenu disappears. Fixes #375

### LINKED ISSUE #375: Desktop Menu: It's not always possible to access the submenus on hover.
When hovering over top-level parent menu items, it's sometimes difficult to access the children: 

hover

Tested in Safari v14.0.
MacOS 10.15.6.

aristath:
Fixes #375

## Summary
Removed the extra 3px from `top` for submenus and made adjustments to properly reposition everything.

The `top` property in `<ul>` submenus was `calc(100% + 3px)`. That extra 3px is what was causing an issue, so that was removed. Instead of that, the 3px was moved to the `<ul>` as `padding-top`. 
That caused another issue, where the focus outline on the parent-menu was cut-off. So to fix that, the background-color was removed from `<ul>` and moved to `<li>` elements.
That change revealed another issue: The background was only set behind the links, leaving a transparent space between them because `<li>` elements had margins. So to fix that, the margins were removed and converted to paddings. 
That revealed another issue, the links hover/click area was too small. So the paddings were moved once again, from `<li>` to `<a>`. That meant that `<a>` had to change from `display:inline` to `display:block` so it can have paddings.
Having paddings/margins in `<li>` items instead of the links themselves didn't make a lot of sense anyway.

Other than fixing the issue described in #375, this PR has 2 other side-effects:
* It's easier to select submenu items since links now have a padding (improved a11y).
* focus/hover areas in submenus have consistent width (see screenshots below)

| Before | After |
| ----- | ----- |
| Screenshot_2020-10-09 My Blog – Just another WordPress site | Screenshot_2020-10-09 My Blog – Just another WordPress site(1) |

## Test instructions

Try to replicate issue reported on #375. Test both desktop & mobile.

## Quality assurance


carolinan (commented):
With the PR applied, I confirmed that:
When hovering over the parent menu item, the sub-menu items are displayed correctly.
The focus outline for the parent menu item is fully visible.

There is a small gap between the background and the outline?
The text of the body content can be seen through this gap.

submenu-gap1

aristath:
> There is a small gap between the background and the outline?
> The text of the body content can be seen through this gap.

That's weird... I tested on Firefox & Chromium on Ubuntu and I can't replicate :thinking:

SResok:
Problem was
```css
a:-webkit-any-link:focus {
    outline-offset: 1px;
}
```
I committed the [fix](https://github.com/WordPress/twentytwentyone/pull/381/commits/227adca1fb6752a960a65b63dc73fe46bdc412c1) 😄

kjellr (approved):
LGTM! Thanks everyone.

PR #385, opened 2020-10-09
Moved font sizes to own utilities file

### LINKED ISSUE #328: Create a separate file for editor font sizes
**Describe the bug**
Updated issue description:

The editor font sizes, the named font sizes like large and gigantic, added by the theme support here:
https://github.com/WordPress/twentytwentyone/blob/trunk/functions.php#L123
are theme specific features, and placing them in their own file makes it easier to locate them and 
work with them if the theme support is updated to remove or add font sizes.

If possible, use one file to be built into both front end styles _and_ the editor styles.
The reason for this is to reduce the number of files with duplicate content.

Remove unused or incorrectly named font sizes:
```
.is-tiny-text,
.has-tiny-font-size 
```

Related files:
https://github.com/WordPress/twentytwentyone/blob/trunk/assets/sass/05-blocks/utilities/_editor.scss

https://github.com/WordPress/twentytwentyone/blob/trunk/assets/sass/05-blocks/utilities/_style.scss

List of the font sizes:
https://github.com/WordPress/twentytwentyone/blob/trunk/functions.php#L123

rolf-yoast:
Fixes https://github.com/WordPress/twentytwentyone/issues/328

## Summary

Moves font-sizes to one file for easy editing

## Test instructions

1. Check if the font-sizes still show up correctly in front-end and the editor.

## Quality assurance


GuidoOffermans (approved):
Clean and simple. Well done! :D

GuidoOffermans:
Acceptence 👍

carolinan:
@rolf-yoast I understand now why you asked about the line heights and how I misunderstood the question :).

assets/sass/05-blocks/utilities/_editor.scss
```diff
@@ -71,39 +69,19 @@
...
-.has-regular-font-size,
-.has-normal-font-size,
-.has-medium-font-size {
-	font-size: var(--global--font-size-md);
-}
-
 .is-large-text,
 .has-large-font-size {
-	font-size: var(--global--font-size-lg);
 	line-height: var(--global--line-height-heading);
```
carolinan: Please move the line heights with the font-sizes :). I meant that all references to the editor font size, the named font sizes such as large and gigantic, should not be inside `_editor.scss` but inside the new `font_sizes.scss` file.

assets/sass/05-blocks/utilities/_font-sizes.scss
```diff
@@ -0,0 +1,94 @@
+// Gutenberg Font-size utility classes
+:root {
```
carolinan, line 2: Did you learn why or if :root is needed here? I'd like to learn why it can't be removed.

rolf-yoast: It's used by Gutenberg in this way on the front end of the site.

carolinan:
I tried removing the named editor font sizes that was outside the root, lines 55-95, and removing it did not break anything in the editor or the front.

assets/sass/05-blocks/utilities/_font-sizes.scss
```diff
@@ -0,0 +1,113 @@
...
+.has-gigantic-font-size {
+	font-size: var(--global--font-size-xxl);
+	line-height: var(--global--line-height-heading);
+
+	// This size is meant to mimic the page titles, so the font weight is reduced to match.
+	font-weight: var(--heading--font-weight-page-title);
+}
+
+// Post title style
+.wp-block.editor-post-title__block {
```
carolinan: We should move these back too, since they only affect the editor.  

Sorry for being unclear -only intended to move the named editor font sizes from this list: https://github.com/WordPress/twentytwentyone/blob/trunk/functions.php#L125

GuidoOffermans:
Acc-test works! I can still see the font sizes correctly :)

assets/sass/05-blocks/utilities/_font-sizes.scss
ryelle: It's for specificity — gutenberg uses it to make sure the colors take effect over other defaults. It could be anything, like a class (`:root` has the specificity level of a class), but it's guaranteed to be on the page since it refers to the root element (`html` in a website).

carolinan: Yes. To be more specific but most likely out of scope for this PR:
The editor font sizes are the **only** styles besides the CSS variables that are placed in the :root. Why is this needed?
For example, the custom colors are not.

carolinan: Oh well, it is not going to matter for much longer since themes eventually will be using the _--wp--preset--_ variables.

ryelle: I see the colors are using this format `.has-black-color[class]`, which accomplishes the same thing as `:root .has-black-color`.

It's needed to override things like if the heading block has a base font size, the `:root .has-large-font-size` (or .has-large-font-size[class]`) rule will override that without needing to explicitly write out each block class.

PR #387, opened 2020-10-09
Add padding to blocks in columns blocks

### LINKED ISSUE #270: Columns block spacing

JustinyAhin:
Fixes #270.

## Summary
Add padding to blocks in columns blocks.

No.

## Test instructions

1. Create a column block with 3 columns.
2. Add both paragraphs and headings in the three columns.
3. When they don't have background images, these blocks should have a padding, both in the editor and in frontend.

## Quality assurance


aristath (approved):
LGTM :+1: 
Requesting sanity check from @carolinan too just to make sure there are no unforeseen side-effects, though I think we're OK to merge this one.

carolinan (approved):
Tested, looks good to me.

PR #392, opened 2020-10-09
Set the text-decoration-thickness to 1px

### LINKED ISSUE #377: Use text-decoration for all link underlines
Some of our underlines are actually `border-bottom: 1px`, while others are `text-decoration: underline`. We should change all of these (preferably to `text-decoration`, and add while we're at it we should add a `text-decoration-thickness: 1px` rule for browsers that support it. 

Screen Shot 2020-10-08 at 3 51 40 PM

SResok:
Fixes #377 

## Summary
Set the text-decoration-thickness to 1px for browsers that support it.

## Test instructions

1. 
1. 
1. 

## Quality assurance

### DIFF
--- a/assets/sass/03-generic/normalize.scss
+++ b/assets/sass/03-generic/normalize.scss
@@ -75,6 +75,7 @@ pre {
 
 a {
 	background-color: transparent;
+	text-decoration-thickness: 1px;
 }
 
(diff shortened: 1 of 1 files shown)


kjellr (approved):
thanks!

PR #393, opened 2020-10-09
Add "Overlapping Images and Text" block pattern

kjellr:
Adds another pattern from #50. 

One difference from [the original mockup](https://github.com/WordPress/twentytwentyone/issues/50#issuecomment-705767188): I brightened the colors of the image on the right so that it feels more similar to the one on the left. 

## Screenshots

Screen Shot 2020-10-09 at 3 14 37 PM
Screen Shot 2020-10-09 at 3 14 29 PM
Screen Shot 2020-10-09 at 3 14 29 PM


melchoyce (approved):
LGTM 👍 Let's make sure we update the copy when we switch out the imagery.

melchoyce:
Actually, is there anything we can do to smooth out some of the middle-width layouts?

Screen Shot 2020-10-09 at 5 15 18 PM
Screen Shot 2020-10-09 at 5 15 28 PM

(That doesn't need to hold up this PR though)

inc/block-patterns.php
```diff
@@ -85,6 +85,19 @@
...
+		)
+	);
+}
```
carolinan:
```suggestion
```

PR #397, opened 2020-10-10
Fix footer margin-top height.

### LINKED ISSUE #396: incorrect footer margin-top height fix.
when review footer in all kinds of device desktop and mobile the margin-top is too much that should be similar as per the header. 

| Current | Expected |
| ------ | ------ |
|Screenshot_1| Screenshot_2|

On the "Expected" screenshot, note the extra margin-top in the footer. These match the margins available on the header.

web-migrates:
Fixes #396 

- Fixed extra footer margin-top in all devices as expected.


melchoyce:
Hi @web-migrates, this is an intentional design decision. There should be 250px of spacing between the footer and the bottom of the page. Going to close this issue, but please continue testing and bringing up issues as you see them.

PR #400, opened 2020-10-10
Fix search widget

### LINKED ISSUE #399: Search widget search input border radius displayed in iOS safari mobile
A search widget search input shows unnecessary border-radius in the safari browser of iOS mobile device, it should be similar to the other devices. 
| Current | Expected |
| ------ | ------ |
|IMG_9247|IMG_9248|

On the "Expected" screenshot, it looks the same as other devices without border-radius.

web-migrates:
Fixes #399 

- search widget search input border-radius fixed for the iOS Safari browser.
- add a media query for the mobile devices which width smaller than 360px.


carolinan:
I can't properly test the new break point, I don't have an iPhone. <del>The webkit addition does not turn off the rounded corners on the field when I view it my iPad.</del> Cleared the browser cache, the search input looks square and correct to me.

bhautikvirani:
Hello @carolinan 
yes, I have cleared the cache, reviewed on private browsing, and tested with the fresh installation. 
The issue for border-radius is only for the iOS mobile device i.e iPhone 6, iPhone 7, iPhone 8 which screen width is less than 360px.

assets/sass/06-components/widgets.scss
```diff
@@ -96,6 +97,10 @@
 			background-color: var(--button--color-background);
 			color: var(--button--color-text);
 		}
+
+		@include media(mobile-small) {
+			margin-top: 5px;
```
kjellr: This added margin needs to apply to larger screen sizes as well. It isn't kicking in on the iPhone 8 or XS for example: 

Screen Shot 2020-10-12 at 8 58 35 AM

I'd recommend adding a `margin-bottom` instead, and adding it to both the field and to the button. It should also use one of our existing spacing variables. (So perhaps something like `calc(0.5 * var(--global--spacing-vertical))`).

kjellr (changes requested):
👋 Hey @web-migrates. Thanks for the PR. The `-webkit-appearance` rule is great, but the added margin isn't applying the way it needs to. I left a note above.

assets/sass/06-components/widgets.scss
web-migrates: Hey, @kjellr we fixed CSS as you mentioned here.

kjellr:
Thanks, @web-migrates. I don't think we need that new breakpoint anymore since it's unused. Once that's removed and the conflicts are resolved, this should be good to merge.

web-migrates:
@kjellr We removed unused breakpoint.

kishanjasani (approved):
LGTM :+1:

kjellr (approved):
Thank you!

PR #408, opened 2020-10-12
Add customizer option for footer-copyright

### LINKED ISSUE #401: Copyright date is incorrect
**Describe the bug**
This was sent to me from @pixolin:

The footer displays a copyright information using function date_i18n(), which seems to retrieve the current year. 
This seems legally tricky. 
According to [https://stackoverflow.com/a/2391555/12490449](https://stackoverflow.com/a/2391555/12490449)

The copyright notice on a work establishes a claim to copyright. The date on the notice establishes how far back the claim is made. 
This means if you update the date, you are no longer claiming the copyright for the original date and that means if somebody has copied the work in the meantime and they claim its theirs on the ground that their publishing the copy was before your claim, then it will be difficult to establish who is the originator of the work

aristath:
Fixes #401 

## Summary
Adds a new customizer option for the footer-copyright text and removes the copyright disclaimer & date.

I opted for a new function to get the footer-copyright content. This way we can use that function both to render the content, and as a callback for the partial-refresh in the customizer setting.

## Test instructions

1. Go to the customizer > Theme Settings
2. Change the footer copyright text
3. Save and confirm everything works both in the customizer and on the frontent.

## Quality assurance

### DIFF
--- a/classes/class-twenty-twenty-one-customize.php
+++ b/classes/class-twenty-twenty-one-customize.php
@@ -116,6 +116,35 @@ if ( ! class_exists( 'Twenty_Twenty_One_Customize' ) ) {
 				)
 			);
 
+			$wp_customize->add_setting(
+				'footer_text',
+				array(
+					'capability'        => 'edit_theme_options',
+					'default'           => '<a href="' . esc_url( __( 'https://wordpress.org/', 'twentytwentyone' ) ) . '" class="imprint">' . esc_html__( 'Proudly powered by WordPress.', 'twentytwentyone' ) . '</a>',
+					'sanitize_callback' => 'wp_kses_post',
+					'transport'         => 'postMessage',
+				)
+			);
+
+			$wp_customize->add_control(
+				'footer_text',
+				array(
+					'type'    => 'textarea',
+					'section' => 'theme_settings',
+					'label'   => esc_html__( 'Footer text and copyright:', 'twentytwentyone' ),
+				)
+			);
+
+			// Add partial for footer_text.
+			$wp_customize->selective_refresh->add_partial(
+				'footer_text',
+				array(
+					'selector'            => '#footer-copyright',
+					'render_callback'     => 'twenty_twenty_one_the_footer_copyright',
+					'container_inclusive' => true,
+				)
+			);
+
 			// Background color.
 			// Include the custom control class.
 			include_once get_theme_file_path( 'classes/class-twenty-twenty-one-customize-color-control.php' ); // phpcs:ignore WPThemeReview.CoreFunctionality.FileInclude.FileIncludeFound
--- a/footer.php
+++ b/footer.php
@@ -34,21 +34,7 @@
 					<?php endif; ?>
 				<?php endif; ?>
 			</div><!-- .site-name -->
-			<div class="copyright">
-				<?php
-				/* translators: 1: Copyright date format, see https://www.php.net/manual/datetime.format.php, 2: Site name */
-				printf(
-					/* Translators: %1$s: Copyright date. %2$s: Site name. */
-					esc_html__( '&copy; %1$s %2$s', 'twentytwentyone' ),
-					esc_html( date_i18n( _x( 'Y', 'copyright date format', 'twentytwentyone' ) ) ),
-					esc_html( get_bloginfo( 'name' ) . '.' )
-				);
-				?>
-				<a href="<?php echo esc_url( __( 'https://wordpress.org/', 'twentytwentyone' ) ); ?>"  class="imprint">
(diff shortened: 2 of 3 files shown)


carolinan:
Code looks OK, but I don't want to make the decision alone, we can bring this up in todays meeting, because as a squad we have tried to use as few customizer options as possible.

aristath:
It could also be a PHP filter instead of an option...

melchoyce:
Yeah, I'd prefer if this wasn't an a Customizer option — especially not one that requires someone to enter HTML into a textarea. I think we should automate this:

> ©[Year] [Site Name]. Proudly Powered by [WordPress](https://wordpress.org/).

melchoyce (changes requested):
Let's remove the Customizer option.

carolinan:
But the year is the root of the problem and it was agreed on during the meeting that we would use the customizer option?

melchoyce:
I had a quick chat with @carolinan and it sounds like we've had a misunderstanding. Let's talk more about this, so I can understand better. What's the problem we're having with displaying the year?

pattonwebz:
Most often we see something like `2002-2020` where people indicate the age and copyright timeframe of the site and it's contents. Having it autogenerate the copyright with just the current year is probably not sufficient for a good portion of the potential user; hence the need for an input to set it.

aristath:
> especially not one that requires someone to enter HTML into a textarea

I agree that HTML in a textarea looks a bit funky. But there is no other way to actually have links... If we want `WordPress` to link to w.org then they need to be able to add the link somehow, which is why in this PR I did it with a textarea/HTML

carolinan:
What if we left the org link (imprint) out of it (left it as it is right now without this setting), and only had a simple text field for the users copyright text.
The field would _allow_ basic HTML like a link or bold text, but it would not be _needed._

If they want to write "Copyright 2002-2020" in plain text they can, if they want to write "This site uses cookies", or link to the privacy policy, they can do that too.

Another thing this solves is the duplication of the site title in the footer:
footer-titles

aristath:
We could also just replace `WordPress` in the text with `<a href="https://w.org">WordPress</a>`...

melchoyce:
Just had a quick chat with @kjellr about this. What if we instead recommended that folks use a text widget to add copyright information, if they need it (we can demo this on https://2021.wordpress.net/), and then had a static "Proudly powered by [WordPress](https://wordpress.org/)" line that developers building sites for clients could filter out or whatever if needed? That would match Twenty Nineteen, Seventeen, etc. 

This way, if folks need formatting, the text widget will allow them to add this without HTML. If they need something really custom, they can use an HTML widget, which has some basic validation that will alert them to things like missing closing brackets.

aristath:
Hmmm So basically add a new widget-area for the footer-right, and if there are no widgets in it then show the hardcoded "powered by" string?
It could work...

melchoyce:
I'm thinking we don't even add anything new, just recommend people using the existing widget area in the footer. (So essentially, just get rid of the copyright as an explicit theme feature and leave it to site builders to include it if they want)

aristath:
I don't understand how that would work...

Currently the footer looks like this:

Screenshot_2020-10-13 My Blog(5)


So we've got the widget-area, followed by the footer nav, then a horizontal line, then below that we've got the logo and the copyright.
If we don't add a new widget area, then the copyright will be above the horizontal line & footer-nav :thinking:

melchoyce:
Yes, but it's an optional feature. If we can't automate it, we may as well take the simplest approach, IMO.

aristath:
So should we also remove the footer logo etc? Basically everything below the horizontal line goes bye-bye? Or should we leave it in and just center the logo?

kjellr:
Yeah, I think we should just look to what the other twenty- themes do here: Allow the the footer credit note to be filterable, and then let it be. No need to make this complicated. 

I think what Mel's suggesting is that folks can just use a text widget in the normal footer widgets area if they need to add a copyright (and can't filter the footer credit line). It's not the 100% ideal solution for this use case, but it's alright. 

In general, I personally think there are more pressing tasks to work on, and I don't think we should spend too much time figuring this out at the moment.

melchoyce:
Quick visual example:

image

aristath:
@melchoyce so if I understand correctly, I'm closing this PR and just changing the string in the copyright notice (remove the copyright date)

melchoyce:
Yup! Sound okay?

aristath:
Yes, sounds good. I'll take care of it tomorrow my morning :+1:

aristath:
Closing this one, replaced by https://github.com/WordPress/twentytwentyone/pull/459

PR #416, opened 2020-10-12
improve the way template-parts are included

aristath:
Improves the way template-parts for the content are included in post-archives and the index.

## `index.php`:
We had this condition:
```php
if ( 'excerpt' === get_theme_mod( 'display_excerpt_or_full_post', 'excerpt' ) ) {
	get_template_part( 'template-parts/content/content-excerpt' );
} else {
	get_template_part( 'template-parts/content/content' );
}
```
That can be simplified as a single-liner:
```php
get_template_part( 'template-parts/content/content', get_theme_mod( 'display_excerpt_or_full_post', 'excerpt' ) );
```

## `archive.php`

~~We were always getting the excerpt. This PR changes it to respect the user's selection from the `display_excerpt_or_full_post` setting so they will get what they want.~~

EDIT: Changes in archive.php were reverted (see comments below)

## Quality assurance


carolinan:
I thought the option was only meant for the blog, not all archives?
_On the Posts page, post show:_

aristath:
Hmmmm so for example a category archive will force-display excerpts and if the user wants to show full posts they need to create a child-theme? I hadn't thought of it that way, I thought the options should be global :thinking:

aristath:
@carolinan I reverted the changes in `archive.php`, this one now only contains that change in index.php :+1:

rolf-yoast (approved):
CR & Acceptance 👍

PR #417, opened 2020-10-12
Reduce gallery caption size

### LINKED ISSUE #238: Gallery block: Small gallery images with captions and links are not clickable
**Describe the bug**
When a gallery image is very small and has both a link and a caption, the caption covers the clickable area of the image.

**To Reproduce**
1. Add a gallery block to a post or page
2. Add a caption to the image
3. Select one of the link options in the block sidebar
4. Set the gallery so that the images are fairly small
5. save and view the error on the front.

**Expected behavior**
The link needs to be clickable

**Screenshots**
gallery-link

JustinyAhin:
Fixes #238

## Summary
Reduce size of gallery caption.

No

## Test instructions

1. Add a gallery block to a post or page
2. Add a caption to the image
3. Select one of the link options in the block sidebar
4. Set the gallery so that the images are fairly small
5. For galleries that have three column and more the font size of the caption is now defined to `--global--font-size-xxs` which value is 0.75rem.

## Quality assurance


kjellr (commented):
I'm a little confused why this PR includes two size rules, and introduces a new font size variable. Can you explain the reasoning there? From the issue, it sounds like the caption size should just be changed to `--global--font-size-xs`, which is an existing variable. 

Also — when changing the font size here, be sure to change it in the editor styles as well.

melchoyce:
> From the issue, it sounds like the caption size should just be changed to `--global--font-size-xs`, which is an existing variable.

Yes, we should go with this instead of introducing anything new.

JustinyAhin:
@kjellr @melchoyce  I introduced the new font size because with `--global--font-size-xs` which is 1rem, the caption still covers all the image.

@aristath mentionned to me that a size under 16px could introduce accessibility issues. So I reverted and use the `--global--font-size-xs`.

kjellr (approved):
Thanks, I think this works fine for now. 👍

PR #419, opened 2020-10-12
Fix customizer text color on background change

### LINKED ISSUE #405: When the body background is reset to default the text color is black
**Describe the bug**
If you change the body background color in the customizer, and then use the option to reset the color to the default,
the text color is still black #000 in the cusomizer, and not the original text color.

The text on the front is the correct color.

**To Reproduce**
1. Go to the customizer
2. Change the background color and save. See the text color change.
3. Reset the background color to the default color.
4. See that the text color is #000.

**Expected behavior**
When the background is the default color, I expect the default text (and border) color to be used.

rolf-yoast:
Fixes https://github.com/WordPress/twentytwentyone/issues/405

## Summary

Set variables instead of hard coded colors for text in the customizer.

## Test instructions

1. Change the background color.
1. Check that the color uses the set variables for color.

## Quality assurance

### DIFF
--- a/assets/js/customize-preview.js
+++ b/assets/js/customize-preview.js
@@ -1,4 +1,4 @@
-(function (api) {
+( function( api ) {
 	/**
 	 * Get luminance from a HEX color.
 	 *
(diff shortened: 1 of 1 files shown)


kishanjasani (approved):
LGTM :+1:

aristath (approved):
LGTM :+1:

PR #420, opened 2020-10-12
Add a z-index value to the overlapping columns block style

kjellr:
This adds a `z-index` value to the overlapping columns block style, in order to ensure that the foremost column appears in front of blocks that have a z-index value assigned (like the Cover block). 

Before: 

Screen Shot 2020-10-12 at 9 54 26 AM

After:

Screen Shot 2020-10-12 at 9 54 16 AM


kjellr:
@melchoyce In addition to the cover block, wasn't there another core block that we saw this issue with? I can't remember which one, but I want to make sure the z-index value here overrides that one too.

aristath (approved):
After running `npm run build` this looks good :+1:

melchoyce:
The other overlapping block was Gallery, but this PR fixes it 👍 

Before | After
----- | -----
image | image

melchoyce (approved):
LGTM 👍

PR #422, opened 2020-10-12
Update starter content

### LINKED ISSUE #1: Add starter content
Starter content makes it easier for users to get started with editing their front page.
It populates menus, widget areas, and can set the front page to be static page. It is also possible to assign featured images.

The starter content front page should reflect the front page of the figma design.
We are able to use markup for blocks inside the content itself and this would be a good opportunity to showcase block patterns.
Images placed directly in the starter content will not be in the media library but in the theme folder.

Because any text content needs to be translated, the earlier we can decide the text, the better. 
The amount of text should be limited for the same reason.

**What kind of images do we want to use for the front page?**
Art? Photography? 
Neutral "block themed" images as in 2020?
-Are there any female artists in the community that we want to highlight?

carolinan:
Fixes #1

## Summary
Adds block patterns and text to the starter content.

No
(The social links in the footer menu can not be tested until the menu has been merged.)

## Test instructions
Starter content can only be tested on fresh sites.

1. Reset your WordPress install or set up a new fresh install
2. Install and activate the theme
3. Go to the customizer and view the starter content.

## Quality assurance

## Screenshot
start


carolinan:
There probably needs to be a spacer block between the large text pattern and the overlapping images

melchoyce:
Do you mind if I take a stab at polishing this in-editor? I can copy/paste the HTML output here after.

carolinan:
I don't mind

carolinan:
I think it also needs a third widget to fill that space on the right side.

melchoyce:
Added a proposal to the issue: https://github.com/WordPress/twentytwentyone/issues/1#issuecomment-709484176

kjellr:
Just pushed a few commits to include @melchoyce's latest content. 

Screenshot: 

ready-lemming jurassic ninja_

carolinan:
There is too much text content: If nothing else, people will not have enough patience to read all 3 blocks. I would go as far as saying reduce it to one text block.

pattonwebz:
The less text we include here the easier it is for the translators to tackle it.

carolinan:
It showcases the borders and overlaps beautifully

carolinan:
Maybe if there was a little more spacing or perhaps a dash of bold inside the text 3 columns so that the block of text breaks up a little.

kjellr:
Here's a brief rewrite to shorten those sections. I'm not sure we can get shorter than this without either removing a column or removing explanations for how to use these features. Let me know what you think:

Screen Shot 2020-10-16 at 10 14 38 AM

carolinan:
I like this :)

melchoyce:
Works for me 👍

melchoyce (approved):

pattonwebz (approved):
Looks good to me, thanks for the content rewrite @kjellr. I agree we probably can't make it much shorter than this.

PR #424, opened 2020-10-12
Try/menu descriptions

### LINKED ISSUE #414: Feature: Menu description
**Is your feature request related to a problem? Please describe.**
A menu description is a text that is displayed below the link in a menu item.

The menu description is an option that is turned off and on by the site creator:

description

But it does not show if the theme does not have support for it -And this can lead to unexpected situations like the users text not showing when they enable it.

Example of menu description in the wild, where "find out..." is the description:

description-example

And a better example here: https://avada.theme-fusion.com/spa/

**Describe the solution you'd like**
I think this would be a good addition that brings something extra to the theme.

aristath:
Fixes #414 

## Summary
Adds the description to menu items.

In the screenshot I'm including items with and without descriptions as well as short/long titles/descriptions to get an idea of the alignments.

Screenshot_2020-10-12 Lorem Ipsum – My Blog

## Test instructions

1. Add descriptions to menu items
1. Check out the front-end.

## Quality assurance


aristath:
@melchoyce as discussed on the team meeting, pinging you for feedback.
This was a quick proof of concept, we can tweak the styles. :+1:

carolinan:
I wonder if the description needs to be smaller or perhaps have a lower font weight? But the parent menu item already has a font weight of 400.

carolinan:
Responsive view:

responsive-description

Hover:
hover-description

Focus:
focus-description

assets/sass/06-components/navigation.scss
```diff
@@ -422,6 +421,13 @@
 
 	}
 
+	.menu-item-description {
+		display: block;
+		clear: both;
+		font-size: var(--global--font-size-xs);
+		text-transform: none;
+		line-height: 1.7;
```
carolinan, line 429: 1.7 line height is also the `--global--line-height-body`, I suggest using that.

melchoyce:
Technical question: does the description have to be a part of the navigation link?

aristath:
> Technical question: does the description have to be a part of the navigation link?

If inside the [`twenty_twenty_one_add_menu_description_args`](https://github.com/WordPress/twentytwentyone/pull/424/files#diff-3f4057e51a7f16fbb630eef28a9d80450ce21e41ea71c5091567e003334f90fcR52) function I change `$args->link_after` to `$args->after`, the description is no longer part of the link. The problem with that, is that the HTML & CSS will get more complicated - specifically for menus that have children. If we go that route, we'll need to play a lot with positioning etc to make sure that buttons stay where they should, and descriptions go below them (the problem is the dropdown buttons positioning)

kjellr (approved):
I agree that it makes sense to keep these in the links.

Design-wise, I think this is ok for now. May need to change depending on the outcome of #469, but we'll work through that in one PR if need be. 

One question: It doesn't look like descriptions are showing up for subnav items. Is that expected? (I've never used this feature in WP myself, so I'm not sure 😄).

aristath:
> One question: It doesn't look like descriptions are showing up for subnav items. Is that expected? (I've never used this feature in WP myself, so I'm not sure smile).

Yeah, that is expected.
We can add them to submenus if we want to (just have to remove the `0 === $depth` condition from https://github.com/WordPress/twentytwentyone/pull/424/files#diff-3f4057e51a7f16fbb630eef28a9d80450ce21e41ea71c5091567e003334f90fcR51), but I don't think I've ever seen descriptions used in submenus...

kjellr:
Cool, I'm happy to leave them out in that case. 👍

melchoyce:
I know this is merged, but it is possible to instead remove the active/hover effect on just the description? (We'd leave the focus the same)

aristath:
> I know this is merged, but it is possible to instead remove the active/hover effect on just the description? (We'd leave the focus the same)

PR #509

PR #425, opened 2020-10-12
Add a theme description to readme and stylesheet

### LINKED ISSUE #119: Add a theme description
**Describe the bug**
The theme is missing a description, this needs to be added to 
https://github.com/WordPress/twentytwentyone/blob/trunk/assets/sass/01-settings/file-header.scss#L6
and the readme.txt file.

pattonwebz:
## Summary
Adds a description for the theme in all the relevent places. There description in this PR I feel is a good one to use based on discussions on the linked issue. There may still need to be minor tweaks to it before it is the _final, final_ version but I think what is in this PR is a good balance between all the suggestions.

Props to: @francescamarano, @marybaum, @melchoyce from the original thread for their suggestions.

Closes #119 

## Quality assurance

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,9 @@ License URI: http://www.gnu.org/licenses/gpl-2.0.html
 
 ## Description
 
-Twenty Twenty-One is a blank canvas for your ideas - it makes the block editor your best brush.
+Twenty Twenty-One is a blank canvas for your ideas and it makes the block editor your best brush.
 
-With new block patterns, which allow you to create a beautiful layout in a matter of seconds, this theme’s soft colors and eye-catching—yet timeless—design will let your work shine.
+With new block patterns, which allow you to create a beautiful layout in a matter of seconds, this theme’s soft colors and eye-catching — yet timeless — design will let your work shine.
 
 Take it for a spin! See how Twenty Twenty-One elevates your portfolio, business website, or personal blog.
 
--- a/README.md
+++ b/README.md
@@ -1,13 +1,19 @@
 # Twenty Twenty-One
 
-Contributors: wordpressdotorg  
-Requires at least: 5.3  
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -1,14 +1,20 @@
...
+Contributors: wordpressdotorg
+Requires at least: 5.3
+Tested up to: 5.5
+Requires PHP: 5.6
+License: GPLv2 or later
+License URI: http://www.gnu.org/licenses/gpl-2.0.html
 
 ## Description
 
+Twenty Twenty-One is a blank canvas for your ideas - it makes the block editor your best brush.
```
melchoyce: This should be an em dash —

README.md
```diff
@@ -1,14 +1,20 @@
...
+
+With new block patterns, which allow you to create a beautiful layout in a matter of seconds, this theme’s soft colors and eye-catching—yet timeless—design will let your work shine.
```
melchoyce: The dashes here are also incorrect — it should be:

> eye-catching — yet timeless — design

melchoyce (dismissed):

README.md
melchoyce: Actually, let's drop the em dash entirely and go with:

> Twenty Twenty-One is a blank canvas for your ideas and it makes the block editor your best brush.

README.md
melchoyce: Looks like the issue here is actually GitHub's monospace font, copying it into the editor shows the em dashes correctly.

README.md
pattonwebz: Made this change

README.md
pattonwebz: Swapped, good catch on this :+1:

melchoyce (approved):
Looks good 👍 Thanks!

PR #404, opened 2020-10-12
Reset the cover block overlay color to black and text to white

JustinyAhin:
Fixes #389.

## Summary
In the editor and in frontend, resetted the overlay of cover block to black and the text inside to white.

## Relevant technical choices:
No.

## Test instructions
This PR can be tested by following these steps:

Add a cover block and add a background image
Add some texts inside, paragraph and heading
In the editor and in frontend, the cover block overlay should be black and the text inside white.

## Quality assurance
* [x] I have thought about any security implications this code might add.
* [x] I have checked that this code doesn't impact performance (greatly).
* [x] I have tested this code to the best of my abilities
* [x] I have checked that this code does not affect the accessibility negatively

PR #427, opened 2020-10-12
Fix default text colors within the Cover Block

kjellr:
Followup from https://github.com/WordPress/twentytwentyone/pull/404. 

That PR overrides some of the existing logic for changing the default text color. This fixed an issue with the default cover background, but it caused issues with any light-colored background colors. 

The core issue appears to just be that the theme was using the wrong variable for the text color (`--cover--color-background` instead of `--cover--color-foreground`). This PR removes the changes from #404, and makes that small update. 

---

Editor

Before|After
---|---
gutenberg test_wp-admin_post php_post=1 action=edit (1)|gutenberg test_wp-admin_post php_post=1 action=edit

Front End

Before|After
---|---
gutenberg test__p=1 (1)|gutenberg test__p=1


aristath (approved):
Fix confirmed :+1: 
I'll wait for a review by @JustinyAhin before merging since this PR basically reverts #404, so I may have missed something.

JustinyAhin (approved):
Looks good to me too.
Thanks :).

PR #428, opened 2020-10-12
Update large text block pattern to use a wide-width header

kjellr:
This removes the columns block hack that was in place previously. It's made possible with the addition of the wide/full alignment for headers, as added in https://github.com/WordPress/gutenberg/pull/25917, shipping in Gutenberg 9.2. 

I tested this with the plugin inactive as well, and it (surprisingly?) did not throw any error in the editor. 

Screen Shot 2020-10-12 at 12 16 01 PM


melchoyce (approved):
LGTM 👍

Issue #386, opened 2020-10-09
Edit link is missing a space


carolinan:
**Describe the bug**
Tested with NVDA screen reader on Windows 10.
The edit link and post title is missing a space and is read out as
_editpost title_

**To Reproduce**
1. Start the screen reader
2. Listen to the edit link / read the speech viewer.

**Expected behavior**
The text should be read out as separate words.

aristath:
I have a solution to suggest...
We could just remove the edit links.
They were initially added to WordPress themes as a shortcut to quickly edit posts, but that was before the "Edit Post" link was added to the adminbar. Other than tradition I really see no reason for their existence anymore.

carolinan:
During Fridays bug scrub, it was decided that the edit link should be kept.

PR #433, opened 2020-10-12
Update "read more" text to "Continue reading"

### LINKED ISSUE #332: I18N: Read more link might be hard to translate correctly

ryelle:
Fixes #332

## Summary

The "Read more about Post Name" string is hard to translate correctly, so we're switching to "Continue reading Post Name".

## Test instructions

1. Set your site to use full posts in Customizer > Theme settings
2. Make a post with a "read more" break
3. View the post list, it should have a "Continue reading" link
    Screen Shot 2020-10-12 at 1 37 58 PM
4. Switch your posts list to "excerpt"
5. If your content is long enough, your post should get cut off with the Continue reading link
    Screen Shot 2020-10-12 at 1 38 04 PM
6. Testing with a screen reader announces the titles correctly, without smushing "ReadingPost Name" (see #386 for example)

## Quality assurance

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -195,8 +195,8 @@ function twenty_twenty_one_get_avatar_size() {
 function twenty_twenty_one_continue_reading_text() {
 	$continue_reading = sprintf(
 		/* translators: %s: Name of current post. */
-		wp_kses( esc_html__( 'Read more %s', 'twentytwentyone' ), array( 'span' => array( 'class' => array() ) ) ),
-		the_title( '<span class="screen-reader-text">&nbsp;' . esc_html__( 'about ', 'twentytwentyone' ), '</span>', false )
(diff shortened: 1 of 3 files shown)


aristath (approved):
LGTM :+1:

Issue #430, opened 2020-10-12
Media & Text Article Title Pattern: Ampersand doesn't render


melchoyce:
**Describe the bug**
The `&` in the block pattern title is displayed as `&amp;`

**Screenshots**
![image](https://user-images.githubusercontent.com/2846578/95770945-33424a00-0c88-11eb-919e-3b9160cb29aa.png)

**Editor version (please complete the following information):**
- WordPress version: 5.5.1
- Does the website has Gutenberg plugin installed?: No

kjellr:
Closed via #434.

PR #434, opened 2020-10-12
Use dotted separator in Media & Text Article Title pattern

### LINKED ISSUE #431: Media & Text Article Title Pattern: Sep block too wide
**Describe the bug**
Because of changes we made to the width of the default separator block, the sep in this block pattern is too wide. Instead, let's consider using the dots block style, which is narrower and better frames the content of the block pattern.

This might also be a good opportunity to update the design of the dots style to use thicker dots.

**Screenshots**

Before:

image

After:

image

- WordPress version: 5.5.1
- Does the website has Gutenberg plugin installed?: No

kjellr:
Fixes #431 

Before:

Screen Shot 2020-10-12 at 3 25 26 PM

After: 

Screen Shot 2020-10-12 at 3 24 56 PM


kjellr:
I also modified the title of the pattern, so that it should fix #430.

melchoyce (approved):
LGTM 👍

PR #439, opened 2020-10-12
Remove horizontal margins from post titles in post pagination

### LINKED ISSUE #438: Post title underline is uneven on hover in post navigation component
**Describe the bug**
Post title underline is off-centered on hover in post navigation component

See animated screenshot below:
79016d2d8f675321271d9f887a712d0d

**To Reproduce**
1. Go to the post navigation area in a single post (you'll need several posts to see the issue in action)
2. Hover the post title for previous or next post
3. Post title is off-centered 😛 

**Expected behavior**
I think we don't need any margin. It looks really great when the post title is aligned with the arrow.
…and it fixes the issue 😏 

a77c78eddf2cf1753ed8eb45031650b5

audrasjb:
Fixes #438 

## Summary

Post title underline is off-centered on hover in post navigation component.
See animated screenshot below:
79016d2d8f675321271d9f887a712d0d

I think we don't really need any margin. It looks really great when the post title is aligned with the arrow.
…and it fixes the issue 😏
a77c78eddf2cf1753ed8eb45031650b5

## Test instructions

1. Go to the post navigation area in a single post (you'll need several posts to see the issue in action)
2. Hover the post title for previous or next post
3. Post title is not off-centered anymore! 🎉 

## Quality assurance


carolinan:
Is there no way to align the two texts without making the links look bad?
It needs to be one link and not two, but maybe the upper text does not need to be linked at all, only the title?

melchoyce:
Yeah, I'd prefer these still line up. They also don't match the mockups:

image
[Figma](https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=1272%3A106)

Is there a reason we have to use bottom borders here, instead of text decorations?

kjellr:
Yeah, this should be using text-decoration instead of border-bottom. Must have been overlooked when https://github.com/WordPress/twentytwentyone/issues/377 was addressed.

audrasjb:
Thanks all for the review!

I updated the PR accordingly. And yeah, it looks better!

a77c78eddf2cf1753ed8eb45031650b5

melchoyce:
Hey @audrasjb, I think you may have uploaded an older gif?

audrasjb:
Yep I edited the gif, sorry 😅

melchoyce:
Still seeing a couple issues — let me walk through them with some visuals, which will hopefully be more clear than my typing 😅 

## 1. The text on the top and bottom lines still don't line up:

image

This is because there is a space between the inline SVG and "Previous/Next":

image

That space, which might be a different width depending upon which operating system you're using, means we can't reliably calculate the margin needed for the post title to line up. I can think of two solutions:

1. Remove the space, and add 5px left/right margins (using the CSS spacing variables) onto the SVGs. Then, calculate the margin from the width of the SVG + the 5px of spacing. This should get the text to more reliably align.
2. Remove the inline SVGs from _inside_ of the link, and instead put them into the before/after for each link.

## 2. The hover effects don't match the rest of the site links:

image

We should use the same styling as the rest of the links in the theme:

image

## 3. The font sizing doesn't match the mockups:

image

We decided to emphasize the post titles, over previous/next, since they're the most important information within this context.

Let me know if you have any questions!

carolinan:
After #483, do we need to rephrase this and split it into separate issue(s)? 
There is still an extra space in the link underline between the arrow icon and text.

PR #435, opened 2020-10-12
Header: Center logo above header if all 3 items are visible

ryelle:
## Summary
If the logo, site title/tagline, and nav menu are all visible, the logo should be centered above the header. If the title/tagline is hidden, the logo should stay to the left.

This also incorporates changes from #362 & #406 — so it also fixes #358 & fixes #398

## Relevant technical choices:

This is using `row-gap` on `site-header` to add extra spacing between the site title/tagline & menu when the menu is long enough to wrap under the title. This allows for better vertical centering vs using a margin, which pushes the nav down. This is not supported in Safari (yet, appears to be on the way), or IE11. 

Vertical alignment with the current `margin-bottom` approach:
Screen Shot 2020-10-12 at 3 42 38 PM

See below for the `row-gap` alignment in this branch.

On Safari, if you have a long menu, you don't get a space anymore:

Screen Shot 2020-10-12 at 4 38 26 PM

So this is a trade-off, but IMO the improved vertical alignment outweighs the edge case of a long menu + older browser, especially since it's not unreadable.

## Test instructions

This PR can be tested by following these steps:
1. Toggle menus, site titles, and logos.
2. Check that it looks like you expect.

## Screenshots

All taken on Mac/Firefox.

1. Has a logo, menu, and title is visible

    screenshot-1

2. Has a logo, menu, and title is not visible

    screenshot-2

3. Has a logo, does not have a menu, and title is visible

    screenshot-3

4. Has a logo, does not have a menu, and title is not visible

    screenshot-4

5. No logo, has menu, and title is visible

    screenshot-5

6. No logo, has menu, and title is not visible

    screenshot-6

7. No logo, does not have a menu, and title is visible

    screenshot-7

8. No logo, does not have a menu, and title is not visible

    screenshot-8

9. Long menu

    screenshot-9

## Quality assurance
* [x] I have thought about any security implications this code might add.
* [x] I have checked that this code doesn't impact performance (greatly).
* [x] I have tested this code to the best of my abilities
* [x] I have checked that this code does not affect the accessibility negatively

PR #442, opened 2020-10-13
Mobile Header Layouts

### LINKED ISSUE #441: Implement the design for the responsive menu with centered logo
With the new design of the header, the responsive design also needs to be updated/implemented.

Figma link for the header with and without logo:
https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=487%3A17

**Site title:**  Font size: 16px, font weight: 500, line-height: 115%;
**Tagline:** Font size: 16px, font-weight; normal (400), line-height: 160%;

The right edge of the button **icon** aligns with the right edge of the border below the logo.

The focus order needs to match the visual order, depending on if the logo or title are linked and enabled:
Linked logo
Linked site title
Menu toggle

aristath:
Fixes #441 

## Summary

This PR adds some header styles to acoomplish the layouts from #441 

In order to apply different styles it was necessary to differentiate in CSS what kind of elements we have. In order to do that, I added 2 new classes (`has-logo` and `has-title-and-tagline`) in the `.site-header` element. The `<header>` was moved to a separate template just to keep things cleaner. Other than that, it was mostly CSS changes.

## Test instructions

Try setting a logo, enabling/disabling the site-title & description from the customizer, and replicate the layouts from #441 

## Screenshots

| Logo + Menu + Title | Logo + Menu | Title + Menu |
| ----- | ----- | ----- |
| Screenshot_2020-10-13 My Blog | Screenshot_2020-10-13 My Blog(3) | Screenshot_2020-10-13 My Blog(1) |

## Quality assurance


aristath:
Changed this PR from draft to ready-for-review so we can test it and fix any remaining styling issues.

ryelle:
The animation when opening & closing the menu feels more distracting now that the button moves. I also think it's not great that the button moves in the first place, but I'll leave that for a designer to weigh in on.

menu-move

I've taken the same set of screenshots that I did on #435, again all taken on Mac/Firefox. There's a regression somewhere that's de-centering the title when there's no menu, and IMO the logo could be larger when it's just logo + menu.

1. Has a logo, menu, and title is visible

    screenshot-1

2. Has a logo, menu, and title is not visible

    screenshot-2

3. Has a logo, does not have a menu, and title is visible

    screenshot-3

4. Has a logo, does not have a menu, and title is not visible

    screenshot-4

5. No logo, has menu, and title is visible

    screenshot-5

6. No logo, has menu, and title is not visible

    screenshot-6

7. No logo, does not have a menu, and title is visible

    screenshot-7

8. No logo, does not have a menu, and title is not visible

    screenshot-8

melchoyce:
> The animation when opening & closing the menu feels more distracting now that the button moves. I also think it's not great that the button moves in the first place, but I'll leave that for a designer to weigh in on.

Is it technically possible to open the menu down, rather than moving the button up? Rough mockup:

image

(Ignore the bad spacing, splicing together screenshots)

aristath:
> Is it technically possible to open the menu down, rather than moving the button up?

It's possible, but not without a lot of javascript. The problem is that the menu will need to be offset horizontally (can't be `position:fixed`), and then we'll need to find the appropriate height for it (can't be auto-calculated just with CSS). Then that also leaves a lot of room on the top of the expanded menu that users can use to actually scroll the page, so we'd have to calc some things there too.

aristath:
Thank you @ryelle for the review. I hadn't considered that menus might be missing, took that into account now :+1: 

> The animation when opening & closing the menu feels more distracting now that the button moves. I also think it's not great that the button moves in the first place, but I'll leave that for a designer to weigh in on.

I agree, that specific scenario is a bit weird... but I couldn't find a better alternative without writing lots of JS. The good thing is that it will only happen if they show a logo, the site-title and a menu, all at the same time. Usually if they show a logo they won't show the site-name/description, so the likelihood of this issue decreases. It's still an issue though :thinking:

kjellr:
This is looking pretty good. To mitigate the jumping close button effect, I'd suggest that we modify the animation a little bit when the site logo is active here. The jumping button is currently accentuated by the sliding up of the menu — it gives the impression that the "Menu" button is sliding upwards too, while in reality, the "Close" button is a whole separate button. 

To try and clear that up, we should try having the menu slide in from the right in this case, and add a fade-in animation to the close button. Something like this (GIF is slowed down to show the effect): 

animation-update

If that sounds good to other folks, I can push to this branch.

aristath:
> If that sounds good to other folks, I can push to this branch.

It looks a lot better to me! :+1:

template-parts/header/site-header.php
```diff
@@ -0,0 +1,46 @@
...
+ * @package WordPress
+ * @subpackage Twenty_Twenty_One
+ * @since 1.0.0
+ */
+
+$has_primary_nav = has_nav_menu( 'primary' );
+
+$wrapper_classes  = 'site-header';
+$wrapper_classes .= has_custom_logo() ? ' has-logo' : '';
+$wrapper_classes .= true === get_theme_mod( 'display_title_and_tagline', true ) ? ' has-title-and-tagline' : '';
```
kjellr, line 14: I'm not seeing this class added when I'm testing. Would someone mind confirming that it's working?

aristath: Hmmm I can't replicate... As soon as I enable the Site Title & Tagline, I can see the class gets added :thinking:

kjellr: Weird. It's not working on my test site for some reason. In any case, I pushed the animation changes in 6ad3f33. They rely on this class being present.

ryelle:
Header without menu is looking much better 👍 

I'm still seeing some awkwardness with the menu animation— let me grab a frame from the gif above. The menu moves down and off, kind of collapsing into the menu button. Maybe the vertical position of the menu needs to be set even while it's hidden, so it's not traveling back down to a lower "resting point"?

Screen Shot 2020-10-14 at 6 31 36 PM

template-parts/header/site-header.php
ryelle: I see the class on my test site — just to ask the obvious question, you're looking at `header.site-header` for it, right?

aristath:
The purpose of a transition is to make the experience smoother and more natural. In this case the transition is making it weird and unnatural. So... maybe we should just remove the transition for this specific scenario?

kjellr:
> The menu moves down and off, kind of collapsing into the menu button. Maybe the vertical position of the menu needs to be set even while it's hidden, so it's not traveling back down to a lower "resting point"?

Just pushed a fix for that: 

animation-update2

melchoyce:
Can we speed up the transition just a tad?

carolinan:
In the gif, the menu content also shifts to the left?

template-parts/header/site-header.php
kjellr: > just to ask the obvious question, you're looking at header.site-header for it, right?

Ha, I was not. 😄 All set now.

kjellr:
> Can we speed up the transition just a tad?

Done. I also fixed up the classnames so that it actually works. (Thank you @ryelle for helping me realize that I was looking in the wrong place for them. 😅)

> In the gif, the menu content also shifts to the left?

@carolinan can you clarify that? [As discussed above](https://github.com/WordPress/twentytwentyone/pull/442#issuecomment-708420501), the menu intentionally comes in from the right with this configuration.

ryelle (approved):
This is looking great now, the horizontal-only animation looks much smoother. I pushed up a tiny fix for #510 since I was noticing the same issue with embeds.

I think this is good to merge now, and if anything else is not right, we can address in another PR.

melchoyce (approved):

PR #450, opened 2020-10-13
Remove twenty_twenty_one_switch_theme function

### LINKED ISSUE #446: Reconsider the WordPress version check
**Describe the bug**
The theme has a version check where it activates a default theme if the WordPress version is lower than 5.3
https://github.com/WordPress/twentytwentyone/blob/trunk/inc/back-compat.php#L23

The current way of implementing this will not work when the theme is the latest default theme.

One option is to remove the backwards compatibility check.

aristath:
Fixes #446

## Summary
Removes the `twenty_twenty_one_switch_theme` function & hook to avoid switching the theme to itself on fail.

## Quality assurance

### DIFF
--- a/inc/back-compat.php
+++ b/inc/back-compat.php
@@ -12,22 +12,6 @@
  */
 
 /**
- * Prevent switching to the theme on old versions of WordPress.
- *
- * Switches to the default theme.
- *
- * @since 1.0.0
- *
- * @return void
- */
-function twenty_twenty_one_switch_theme() {
-	switch_theme( WP_DEFAULT_THEME );
-	unset( $_GET['activated'] ); // phpcs:ignore WordPress.Security.NonceVerification
-	add_action( 'admin_notices', 'twenty_twenty_one_upgrade_notice' );
-}
-add_action( 'after_switch_theme', 'twenty_twenty_one_switch_theme' );
-
-/**
  * Adds a message for unsuccessful theme switch.
  *
  * Prints an update nag after an unsuccessful attempt to switch to


carolinan:
I created a branch, but you were first:
https://github.com/WordPress/twentytwentyone/commit/821aba5778a323294f0355c2ba7c47709c5cfa5c

Removes the return after the back-compat file is included
Keeps the theme switch function, only to show the notice

carolinan:
How to test:
Deactivate the theme if it is already active
Move the minimum version up, to a version higher than your current, say 5.6.
Activate the theme
See notice
View the front, make sure it's not broken.

ryelle:
I'm getting a PHP error on the frontend with this branch:
```
Fatal error: Uncaught Error: Call to undefined function twenty_twenty_one_post_thumbnail() in /var/www/html/wp-content/themes/twentytwentyone/template-parts/header/excerpt-header.php:20 
```

Probably due to the `return` after the back-compat include. I tested @carolinan's branch too, and that solution seems to work better — I see the notice post activation (which I don't on `fix/446`), and the frontend still technically works.

aristath:
> Probably due to the return after the back-compat include. I tested @carolinan's branch too, and that solution seems to work better — I see the notice post activation (which I don't on fix/446), and the frontend still technically works.

I concur. Closing this one and we'll merge Carolina's branch :+1:

aristath:
PR with @carolinan's branch: https://github.com/WordPress/twentytwentyone/pull/462

PR #391, opened 2020-10-09
Post Formats

aristath:
Fixes #367 

## Summary

Currently WIP - do not merge. In this 1st pass I'll only do block-based content since this is the most urgent part. Once that is done we can move on to legacy content.

## Tasks
* [x] `image`: If there's a featured image, use the featured image. If there isn't the featured image, pull out the first image block from the content and display it above the excerpt.
* [x] `video`: Pull out the first video or video embed block (VideoPress, YouTube, Vimeo, Dailymotion, TikTok) from the content and show it above the excerpt.
* [x] `aside`: Don't show the post title. Show the post content in larger text (Large Paragraph style) instead of an excerpt.
* [x] `status`: Don't show the post title. Show the post content in larger text (Large Paragraph style) instead of an excerpt.
* [x] `gallery`: Pull out the first gallery in the post and show it above the excerpt.
* [x] `audio`: Pull out the first audio block or audio embed block (SoundCloud, Spotify, Mixcloud) from the content and show it above the excerpt.
* [x] `link`: If possible, show the first link as the excerpt in larger text (Large Paragraph style). If not, show the first paragraph of the content in larger text, retaining formatting.
* [x] `quote`: Pull out the first quote or pullquote block from the content and display it instead of the excerpt.
* [x] `chat`: Show the first two paragraphs of the content, retaining formatting.

## Test instructions

This PR can be tested by following these steps:
1. 
2. 
3. 

## Quality assurance
* [x] I have thought about any security implications this code might add.
* [x] I have checked that this code doesn't impact performance (greatly).
* [x] I have tested this code to the best of my abilities
* [x] I have checked that this code does not affect the accessibility negatively

Issue #367, opened 2020-10-08
Add Post Formats support


melchoyce:
Now that we're in a good place with the base of the theme, let's consider adding some minimal post formats support.

Let's take a look at what the archive could look like:

![Archive](https://user-images.githubusercontent.com/2846578/95475768-949aae00-0954-11eb-9a7b-2a40e8259a8f.jpg)

[Figma mockup](https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=363%3A0)

## Standard

Exactly as the archive exists now. Title, featured image if selected, excerpt.

## Image

If there's a featured image, use the featured image. If there isn't the featured image, pull out the first image block from the content and display it above the excerpt.

## Video

Pull out the first video or video embed block (VideoPress, YouTube, Vimeo, Dailymotion, TikTok) from the content and show it above the excerpt. 

## Aside and Status

Don't show the post title. Show the post content in larger text (Large Paragraph style) instead of an excerpt.

## Gallery

Pull out the first gallery in the post and show it above the excerpt.

## Audio

Pull out the first audio block or audio embed block (SoundCloud, Spotify, Mixcloud) from the content and show it above the excerpt.

## Link

If possible, show the first link as the excerpt in larger text (Large Paragraph style). If not, show the first paragraph of the content in larger text, retaining formatting.

## Quote

Pull out the first quote or pullquote block from the content and display it instead of the excerpt.

## Chat

Show the first two paragraphs of the content, retaining formatting.

PR #456, opened 2020-10-13
Post Formats: Fix embeds in audio & video formats

ryelle:
See #391 — This iterates on the post formats code a little. First, it fixes an issue with embeds, and then it swaps the `excerpt-*` templates to use `has_block` to check for blocks, rather than having the conditional also be responsible for outputting the content.

Embed block issue: Once created, embed blocks appear to use `core-embed/spotify`, `core-embed/youtube` etc, so the detection of `core/embed` isn't working. This introduces a rough wildcard match. Now we can pass in `core-embed/*` to `twenty_twenty_one_print_first_instance_of_block`, and it'll return the first embed.

The `has_block` change was introduced to avoid confusion when looking at the template code. It's not immediately clear where the block content is printed out, since it's happening in the conditional. This change cleans up the conditionals so they're more understandable.

(Happy to split this PR in two for each change if anyone asks, I was already here so ended up doing both in one 😬 )

## Test instructions

1. Create an audio or video post using an embed
2. Add some extra content to the post, maybe add an excerpt
3. View your post in an archive, it should display the embed and the excerpt
4. View the post (singular view), it should show the full post content

---

1. Try creating other post formats
2. They should each display using the logic on #367

## Quality assurance

### DIFF
--- a/template-parts/excerpt/excerpt-audio.php
+++ b/template-parts/excerpt/excerpt-audio.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Show the excerpt.
+ * Show the appropriate content for the Audio post format.
  *
  * @link https://developer.wordpress.org/themes/basics/template-hierarchy/
  *
--- a/template-parts/excerpt/excerpt-audio.php
+++ b/template-parts/excerpt/excerpt-audio.php
@@ -9,17 +9,13 @@
  * @since 1.0.0
  */
 
-// Print the 1st instance of an audio block.
-// If there was no audioblock, print the 1st audio-embed block.
-if (
-	! twenty_twenty_one_print_first_instance_of_block( 'core/embed', get_the_content() ) &&
-	! twenty_twenty_one_print_first_instance_of_block( 'core/audio', get_the_content() )
-) {
+$content = get_the_content();
 
-	// Fallback to the content.
-	the_content();
+if ( has_block( 'core/audio', $content ) ) {
+	twenty_twenty_one_print_first_instance_of_block( 'core/audio', $content );
 } else {
-
-	// Add the excerpt.
-	the_excerpt();
+	twenty_twenty_one_print_first_instance_of_block( 'core-embed/*', $content );
 }
+
+// Add the excerpt.
+the_excerpt();
--- a/template-parts/excerpt/excerpt-audio.php
+++ b/template-parts/excerpt/excerpt-audio.php
@@ -13,6 +13,8 @@ $content = get_the_content();
 
 if ( has_block( 'core/audio', $content ) ) {
 	twenty_twenty_one_print_first_instance_of_block( 'core/audio', $content );
+} elseif ( has_block( 'core/embed', $content ) ) {
+	twenty_twenty_one_print_first_instance_of_block( 'core/embed', $content );
 } else {
 	twenty_twenty_one_print_first_instance_of_block( 'core-embed/*', $content );
 }
--- a/template-parts/excerpt/excerpt-chat.php
+++ b/template-parts/excerpt/excerpt-chat.php
@@ -9,7 +9,6 @@
  * @since 1.0.0
  */
 
-// Print the 1st 2 paragraphs, or fallback to the excerpt.
-if ( ! twenty_twenty_one_print_first_instance_of_block( 'core/paragraph', get_the_content(), 2 ) ) {
-	the_excerpt();
-}
+twenty_twenty_one_print_first_instance_of_block( 'core/paragraph', get_the_content(), 2 );
+
+the_excerpt();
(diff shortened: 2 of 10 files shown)


template-parts/excerpt/excerpt-chat.php
```diff
@@ -9,7 +9,6 @@
  * @since 1.0.0
  */
 
-// Print the 1st 2 paragraphs, or fallback to the excerpt.
-if ( ! twenty_twenty_one_print_first_instance_of_block( 'core/paragraph', get_the_content(), 2 ) ) {
-	the_excerpt();
-}
+twenty_twenty_one_print_first_instance_of_block( 'core/paragraph', get_the_content(), 2 );
+
+the_excerpt();
```
aristath: The excerpt here should only be printed if we were unable to find paragraph blocks, so basically for old, non block-based content.

template-parts/excerpt/excerpt-video.php
```diff
@@ -9,16 +9,13 @@
...
 
-// Print the 1st video we can find.
-if (
-	! twenty_twenty_one_print_first_instance_of_block( 'core/embed', get_the_content() ) &&
-	! twenty_twenty_one_print_first_instance_of_block( 'core/video', get_the_content() )
-) {
+$content = get_the_content();
 
-	// Fallback to the content if no block was found.
-	the_content();
```
aristath, line 19: If there is no `core/video` or `core/embed`/`core-embed/*` block we'll need to fallback to `the_content()` instead of `the_excerpt()`

template-parts/excerpt/excerpt-audio.php
```diff
@@ -9,17 +9,13 @@
...
-// Print the 1st instance of an audio block.
-// If there was no audioblock, print the 1st audio-embed block.
-if (
-	! twenty_twenty_one_print_first_instance_of_block( 'core/embed', get_the_content() ) &&
-	! twenty_twenty_one_print_first_instance_of_block( 'core/audio', get_the_content() )
-) {
+$content = get_the_content();
 
-	// Fallback to the content.
-	the_content();
```
aristath, line 20: If there is no `core/audio` or `core/embed`/`core-embed/*` block we'll need to fallback to `the_content()` instead of `the_excerpt()`

aristath (changes requested):

template-parts/excerpt/excerpt-video.php
ryelle: @melchoyce & I just talked this through, and since these templates are only used when the "excerpt" view is shown, it doesn't make sense to show the full content in some cases. We should show the audio/video/embed if we can find it, and the excerpt always.

Based on that convo I've also updated the other `excerpt-*` templates, since only aside, status, and link should use `the_content`.

aristath: This is a problem....
There are hundreds of sites that have older content, and the current approach will only work for block-based posts. That's why we had the `the_content()` fallbacks.
If the site has posts that are not block-based, then `the_excerpt` will not render anything.
You can test this for example by importing the test XML file from https://raw.githubusercontent.com/WPTT/theme-unit-test/master/themeunittestdata.wordpress.xml which contains old post-format posts.
Falling-back to `the_content` instead of `the_excerpt` will take into account those sites.

I added a PR against your fork with the tweaks I had in mind (https://github.com/ryelle/twentytwentyone/pull/2)
I apologize for the extra files in the diff, I pulled from trunk first

carolinan:
Screenshot of post formats with trunk:

postformat

With this PR:
postformats-456


-I prefer the trunk version that is not empty.

template-parts/excerpt/excerpt-video.php
melchoyce: > There are hundreds of sites that have older content, and the current approach will only work for block-based posts. That's why we had the the_content() fallbacks.

I'm okay with this. This theme should show off the best of new features and show off blocks. Older posts aren't broken — you can convert the older content to blocks if you want your site to take advantage of the block-formatted post format excerpts on archive pages. And if that's a problem, the option to show all of your content exists via the Customizer setting.

aristath: This PR changes the syntax and makes an awesome job at making it more readable for developers. But it doesn't actually "fix" anything for users, in fact it breaks backwards-compatibility. 
Is improving code readability for developers worth a clear regression for end-users/visitors?

melchoyce: This theme is designed to show off the block editor rather than the classic editor. If folks want to use the classic editor, there are previous default themes available that look great. If they want their site to support post formats with this theme, there are a few plugins I can recommend to convert old classic content to blocks.

PR #460, opened 2020-10-14
Add text-decoration-skip-ink:all

### LINKED ISSUE #457: Ensure use of text-decoration-skip-ink across all links for the site
From #390:

> **Ensure use of text-decoration-skip-ink across all links for the site**
> This should be on by default in most browsers, but just in case, let's double check. The text-decoration-skip-ink feature can improve readability for dyslexic users, though it is not a 100% guarantee that it will work for everyone.

> ...we should define it just in case older/ less common browsers don't use it by default.

aristath:
Fixes #457

## Summary

Add `text-decoration-skip-ink: all` for links.

## Quality assurance


rolf-yoast (approved):
LGTM 👍

PR #461, opened 2020-10-14
Remove underline from post-titles

### LINKED ISSUE #390: Consider removing the underline on the post titles to improve readability.
As a person who has dyslexia, underline makes it harder for me to read. For the posts page I find it difficult to read the titles. I’ve checked the Figma files, and there is no underline for the post titles. This ticket #110 has been closed, but I’m not sure if there was a consensus.

aristath:
Fixes #390

## Summary

Removes underline from post-titles on post-archives.
The underline still exists on hover/active.

## Quality assurance


assets/sass/06-components/entry.scss
```diff
@@ -8,8 +8,11 @@
 
 	a {
 		color: var(--entry-header--color-link);
+		text-decoration: none;
 
 		&:hover {
+			text-decoration: underline dotted;
```
kjellr: This shorthand is being discarded in Safari (v14.0). Works fine in Chrome and FF, but I think we need to separate it out for now. 

Screen Shot 2020-10-14 at 9 06 15 AM

aristath: Huh... Didn't know about that, thanks!
Separated.
I'll create a different PR because I see we use the same shorthand in other places too.

carolinan:
I am not found of the larger dots on hover in Chrome, maybe now that the default link state has no underline, the hover could be a plain underline?

Chrome:
dots

Firefox:
dots-firefox

aristath:
Closing in favor of #502

PR #462, opened 2020-10-14
Remove the theme switch from the upgrade notice

### LINKED ISSUE #446: Reconsider the WordPress version check

aristath:
Fixes #446 

## Summary
Discussion on https://github.com/WordPress/twentytwentyone/pull/450

## Test instructions

* Deactivate the theme if it is already active
* Move the minimum version up, to a version higher than your current, say 5.6.
* Activate the theme
* See notice
* View the front, make sure it's not broken.

## Quality assurance

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -12,7 +12,6 @@
 // This theme requires WordPress 5.3 or later.
 if ( version_compare( $GLOBALS['wp_version'], '5.3', '<' ) ) {
 	require get_template_directory() . '/inc/back-compat.php';
-	return;
 }
 
 if ( ! function_exists( 'twenty_twenty_one_setup' ) ) {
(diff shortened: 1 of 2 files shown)


ryelle (approved):
👍   I'm seeing the notice, and the frontend works.

> This theme requires at least WordPress version 5.3. You are running version 5.2. Please upgrade and try again.

The notice text should probably be updated too, at least to remove the "try again" bit.

PR #465, opened 2020-10-14
fix skip-link position & selector

### LINKED ISSUE #464: The skip link appears at the top right
**Describe the bug**
The skip link appears at the top right of the screen, while it is expected to appear at the top left.

**To Reproduce**
- go to any page
- use the Tab key to navigate through focusable elements
- once the "skip link" receives focus, it appears at the top right of the screen

**Expected behavior**
The skip link to appear at the top left, which is the "canonical" place where skip links are expected to appear. The "Skip to toolbar" and "Log out" links appear at the top left *when logged in and the admin bar is displayed). After those 2 links, the Skip to content appears at the top right instead, which is likely out of view for many users with low vision, users who use screen magnification etc.  Also impacts all users, considering the F pattern.

**Additional context**
Right now, the skip link is revealed via the CSS used for the `.screen-reader-text:focus` ruleset. I think it should be overridden by a specific ruleset using the selector `.skip-link:focus`, like other bundled themes do.

aristath:
Fixes #464

## Quality assurance


afercia (approved):
LGTM thanks!

PR #466, opened 2020-10-14
Added reduced motion scss

### LINKED ISSUE #409: Add support for prefers reduced motion
As our goal is to create an accessible theme complying to WCAG level AAA,
and with the hopes that all transforms and animations has now been added, the theme should support 
_prefers reduced motion:_

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

2.3.3 Animation from Interactions Level AAA
https://www.w3.org/TR/WCAG21/#animation-from-interactions

rolf-yoast:
Fixes https://github.com/WordPress/twentytwentyone/issues/409

## Summary

Added a fix for reduced motion

## Test instructions

1. Check if the motion is reduced when turned on.

## Quality assurance


aristath (approved):
LGTM :+1:

PR #473, opened 2020-10-14
Rename navigations aria-label attributes

### LINKED ISSUE #470: Navigation landmark should not be labelled as "navigation"
**Describe the bug**

Screen readers already announce `<nav>` elements as "navigation".  Labelling a nav element with a string that contains the word "navigation" is redundant:

`<nav aria-label="Footer Navigation" class="footer-navigation">`

The above is announced as:

"Footer Navigation, navigation".

The word "Navigation" should be removed. That would leave us with only the word "Footer", which would be confusign for a navigation element (there's already a "footer" landmark).

I'd suggest to just rename this nav to "Secondary menu" or, if this menu is meant for a specific purpose e.g. social links or whatever else, use a more specific name. With the first option, the primary menu should be renamed to "Primary menu".

aristath:
Fixes #470

## Summary
Renames the primary navigation aria-label to `Primary menu`, and the footer-navigation to `Secondary menu`.

## Quality assurance

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -21,7 +21,7 @@
 		<?php get_template_part( 'template-parts/footer/footer-widgets' ); ?>
 
 		<?php if ( has_nav_menu( 'footer' ) ) : ?>
-			<nav aria-label="<?php esc_attr_e( 'Footer Navigation', 'twentytwentyone' ); ?>" class="footer-navigation">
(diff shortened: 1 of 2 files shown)


afercia (approved):
Cool, thanks!

PR #474, opened 2020-10-14
Remove uppercase text-transform from navigation

### LINKED ISSUE #469: Navigation menu: avoid all-caps text
**Describe the bug**
All-cap text is known to be harder to read, especially for some users, for example users with dyslexia, and it's generally less readable for all users. Worth reminding dyslexia comes in many different types and what may be good for some users may not be good for others. Ideally, there should be a mechanism to customize many typeface metrics including spacing between characters, spacing between words, line-height, and the ability to disable all-caps. Given that this is not possible, the best option is to provide safe defaults and avoid all-caps.

Also, all-caps text causes some screen readers to spell the words instead of speaking them as words.

As far as I can tell, the Web Content Accessibility Guidelines don't explicitly mention all-caps text. However, it is mentioned in the W3C recommendation for Low Vision:

Accessibility Requirements for People with Low Vision
3.3.4 Capitalization
https://www.w3.org/TR/low-vision-needs/#capitalization
> Text in all capital letters is more difficult to read for most people, with and without disabilities.

This is still a W3C "Working Draft" so, technically, it's not normative. That's what we have.

Other authoritative references:

Harvard University
https://accessibility.huit.harvard.edu/design-readability
> Avoid using all caps. Readability is reduced with all caps because all words have a uniform rectangular shape, meaning readers can't identify words by their shape.

WebAIM (Center for Persons with Disabilities Utah State University)
https://webaim.org/articles/evaluatingcognitive/
> Do not use italics or bold on long sections of text. Be careful with ALL CAPS.

Some accessibility auditing tools do even have a specific rule to detect all-caps text, for example the tool from the UK Government Digital Service:
https://alphagov.github.io/accessibility-tool-audit/tests/typography-all-caps-text-found.html

I do realize removing the all-caps text has an impact on the design. However, if we want to provide users with the highest possible level of accessibility I'd strongly recommend to reconsider the design choice to use all-caps and use normal text instead. 

**To Reproduce**
- see the all-caps text in the navigation menu

aristath:
Fixes #469

## Summary
Removes the `text-transform: uppercase` as per #469 

Personal opinion: I very much prefer the non-uppercase version. I can read it a lot easier/faster, and as an additional benefit it has a slightly smaller width so it won't break to 2 lines as early on smaller screens.

## Screenshots

Before:

Screenshot_2020-10-14 My Blog(2)

After:

Screenshot_2020-10-14 My Blog(3)

## Quality assurance


kishanjasani:
@aristath It will be better to use `text-transform: capitalize` for a better look.

aristath:
> @aristath It will be better to use text-transform: capitalize for a better look.

Not really necessary... If they want they can just add the menu titles the way they want them. If they want the 1st letter capitalized they should type it like that :laughing:

kishanjasani (approved):
LGTM :+1:

pattonwebz (approved):
Changes here are good on a technical level. We should probably wait on design to weight in as well on the adjustment though

PR #477, opened 2020-10-14
Don't use text-decoration shorthand

aristath:
## Summary
Initially reported on https://github.com/WordPress/twentytwentyone/pull/461/files/b0db7f2ec340253c957c39011e34119c6163032b#r504661523 by @kjellr:

> This shorthand is being discarded in Safari (v14.0). Works fine in Chrome and FF, but I think we need to separate it out for now.

This is a better-safe-than-sorry PR and separates the `text-decoration` shorthand to separate rules where necessary.

## Quality assurance


kjellr (approved):
Thanks!

Issue #429, opened 2020-10-12
Test and add support for high contrast mode.


carolinan:
Test the theme with the OS high contrast mode.

Make sure that all controls and links work.

Resources:
https://adrianroselli.com/2017/11/os-high-contrast-versus-inverted-colors.html
https://www.a11yproject.com/posts/2020-01-23-operating-system-and-browser-accessibility-display-modes/

carolinan:
I have tested Windows high contrast mode with Edge (chromium), Internet Explorer 11 and Firefox.
The visible focus indicator works well. The input fields like the search and comment form works well. 

High contrast black:
The controls works well with one exception:
In IE11, the plus and minus icon for the menu is not visible at all.
In Edge, and Firefox, the plus and minus icon is difficult to see because they are dark grey over black background.

High contrast white:
No issues found.

ryelle:
I took a few screenshots, but TBH I'm not really sure what I'm looking for here. I noticed that Edge & Firefox use different link colors in High Contrast Black (which makes FF almost unreadable, but it was better on my PC monitor than it looks in the screenshot). I don't know if the +/- icon is an issue since it seems fairly high contrast against the black, but I didn't check the contrast ratio.

| Firefox | Edge |
|--------|------|
| ![high-contrast-black-menu](https://user-images.githubusercontent.com/541093/95924504-b2697800-0d85-11eb-899a-1d2cf9ec1b3a.jpg) | ![high-contrast-black-menu-edge](https://user-images.githubusercontent.com/541093/95924503-b2697800-0d85-11eb-94ce-19ec48aba3d7.jpg) |

In High Contrast White, I noticed the footer links have this double underline. I didn't notice it anywhere else, but didn't look very hard.

![high-contrast-white-double-link](https://user-images.githubusercontent.com/541093/95924505-b2697800-0d85-11eb-8b6b-ddfc35191532.jpeg)

There's an extra border under the entry meta…

![high-contrast-white-extra-border](https://user-images.githubusercontent.com/541093/95924506-b3020e80-0d85-11eb-880b-19255684ccda.jpg)

The menu looks fine in White 👍 

![high-contrast-white-menu](https://user-images.githubusercontent.com/541093/95924507-b3020e80-0d85-11eb-984a-491ee893b705.jpg)

carolinan:
I tested the four default color schemes again because I did not get the same result as in the screenshots above in my first test. 

The plus and minus icon **is** visible in IE11.
[...]

PR #483, opened 2020-10-14
Remove borders in post navigation

### LINKED ISSUE #377: Use text-decoration for all link underlines

ryelle:
Fixes #377, see https://github.com/WordPress/twentytwentyone/issues/429#issuecomment-708384060 — This removes the last border-bottom on `nav a` in favor of intentionally setting text-decoration. It also removes a few leftover `border-bottom: none` where (I believe?) they're not necessary anymore.

@carolinan Were there any other places where you've seen borders?

I've moved the pagination & navigation styles out of the generic `nav a` selector and into their respective sass partials, so that it doesn't cause any unexpected side effects.

## Test instructions

1. View a few pages, archives, etc
2. Links should be styled as expected
3. Hover & focus on links, they should be styled correctly.

## Screenshots

Screen Shot 2020-10-14 at 3 18 18 PM
Screen Shot 2020-10-14 at 3 14 20 PM

This should be the only "big change", part of which is already underway in #439, this uses text-decoration now instead of borders.
Screen Shot 2020-10-14 at 3 19 30 PM

<details>
<summary>Full page in Mac/Firefox</summary>

Screen Shot 2020-10-14 at 15 13 23-fullpage

</details>

## Quality assurance


kjellr (approved):
Looks good to me! I hovered, focused, and clicked around and didn't notice any negative effects. Thanks!

carolinan (approved):
LGTM, good riddance of the _nav a_ style :)

PR #484, opened 2020-10-15
Move footer widget area

### LINKED ISSUE #472: The footer should not contain the widgets
**Describe the bug**
Not sure if this has been discussed previously and what the argumentation behind this design choice are but the footer should not contain the widgets.

Semantically, a footer element that is child of the body represents "content information" related to the page/site such as copyright info, privacy and legal info, and maybe a short set of links (a menu) related to this kind of information.

According to the HTML 5.2 spec:

https://www.w3.org/TR/html52/sections.html#the-footer-element
> A footer typically contains information about its section, such as who wrote it, links to related documents, copyright data, and the like.
> A footer element can also contain entire sections representing appendices, indexes, long colophons, verbose license agreements, and other such content.

A more appropriate element for the widgets, which represent "ancillary information" related to the content, would be an `<aside>` element which has a semantic of "complementary".

I'd recommend to move the widgets template part to a new `<aside>` element  to be added before the footer.

The new aside element would probably require some minimal CSS, basically some of the CSS rules already used for the `.site-footer` especially the ones related to the width.

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/472

## Summary
Moves the widget area above the `<footer>` element.
Changes the wrapping `<div>` to `<aside>`.

Updates vertical spacing depending on which elements are active (menu, widgets, or none of them).
Creates a new body class if there are no active widgets in the widget area, to add the extra large top margin to the footer element instead of the widget area.

Makes better use of the CSS variables for text and link color so that these can be changed.

See summary.

## Test instructions

1. Add widgets to the footer area
2. View the front
3. Check that the widgets are no longer inside the <footer>,
4. Check that the text sizes, link styles, width and spacing is correct across screen sizes.

## Screenshots
Without widgets or menu:
no-menu-or-widgets

Menu only:
footer-menu

Widgets only:
footer-widgets

Both menu and widgets:
footer-menu-and-widgets

## Quality assurance


footer.php
```diff
@@ -28,6 +28,7 @@
 						array(
 							'theme_location' => 'footer',
 							'items_wrap'     => '%3$s',
+							'container'      => 'false',
```
carolinan, line 31: There was a container div here between the ul and li that was creating an error in [axe](https://www.deque.com/axe/) when testing,

aristath (approved):
Reviewed and tested, LGTM :+1:

PR #489, opened 2020-10-15
Remove the link on published date

### LINKED ISSUE #481: Unlink the publish date
**Describe the bug**
Splitting this out from https://github.com/WordPress/twentytwentyone/issues/390#issuecomment-708456489 

For a number of years, bundled themes have been linking the post "publish date" in the archive pages to the post permalink. Following this example, many WordPress themes do the same thing.

date

However, a link with text, for example, "October 14, 2020" which is then linked to the post doesn't make much sense as the link text doesn't identify the link destination/purpose. This impacts all users, regardless whether they use assistive technology or not. It also impacts SEO, in a way.

Visually: users see a date and they might think the link points to something specific to that date e.g. a day archive or events for that day, etc.
Screen readers will read a date, so users won't have a clue the link actually points to the post. This is especially true when the link text is read out of context.

Also, it would be great to reduce the amount of redundant links in the archive pages. Worth reminding that all post titles are already linked with the notable exception of two post formats: Aside and Status.

For these two post formats there's the need of an alternative solution. It is certainly possible to keep the date linked only for these two post formats but the link semantics wouldn't be correct. Maybe a "read more" link would be a better solution to explore. 

**To Reproduce**
- go to an archive page
- see the posts publish date is linked to the post, which just repeats the functionality of the linked post title
- see that posts with the Aside and Status formats don't have a title, so they would need an alternative solution

rolf-yoast:
Fixes https://github.com/WordPress/twentytwentyone/issues/481

## Summary

Unlinks the publish date

## Test instructions

1. Go to the archive
1. See that the published date aren't linked anymore

## Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -930,6 +930,24 @@ template {
 	}
 }
 
+.widget-area {
+	max-width: calc(100vw - 30px);
+	margin-left: auto;
+	margin-right: auto;
+}
+
+@media only screen and (min-width: 482px){
+	.widget-area{
+	max-width: calc(100vw - 100px);
+	}
+}
+
+@media only screen and (min-width: 822px){
+	.widget-area{
+	max-width: min(calc(100vw - 200px), 1240px);
+	}
+}
+
 .full-max-width {
 	max-width: 100%;
 	width: 100%;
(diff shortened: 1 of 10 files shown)


carolinan:
On archive pages, a space between the date and edit is missing when the user is logged in:

archive

carolinan:
The above spacing is resolved, Looks good to me, can be merged after the conflict is solved.

PR #490, opened 2020-10-15
Fix button heights when menu-item has description

### LINKED ISSUE #487: Sub-menu toggle size is larger when a description is used
**Describe the bug**
When a menu item has a description, the toggle button height is the same as the combined menu item and description text.
The plus and minus icons is vertically centered, even for items that has no description.

**To Reproduce**
1. Create a menu that has both sub menus and descriptions
2. Assign the menu to the primary menu location
3. View the front

**Expected behavior**
I think the button should be in the same position no matter if there is a description or not.

**Screenshots**

menu-description-icon

Button in focus, here the button size is more obvious:
menu-description-icon-focus

WordPress version: 5.5.1
Desktop (please complete the following information):

OS: Windows 10
Browser Chrome version 86.0.4240.75

aristath:
Fixes #487 

## Summary
Changes the button height.

## Quality assurance


carolinan:
LGTM

description-updated

assets/sass/06-components/navigation.scss
```diff
@@ -212,7 +212,7 @@
 		// Sub-menu buttons
 		.sub-menu-toggle {
 			display: flex;
-			height: 100%;
+			height: calc(2 * var(--primary-nav--padding) + 1.15em + 1px);
```
ryelle, line 215: I know this is already merged, but it would be nice to see a comment explaining where this formula came from.

PR #493, opened 2020-10-15
Remove remnants of microformats

### LINKED ISSUE #482: Remove remnants of microformats: rel=bookmark and the like
**Describe the bug**
For a number of years, bundled themes have been using `rel=bookmark` attributes on the post links and CSS classes related to microformats like `vcard`, `url fn n`, and maybe more.

This practice comes from a time when microformats were still kept in some consideration. I'm not sure that today they add great value as microformats are basically not used any longer and there are other, more modern, ways to add structured data to web content.

Some examples:

Link to a post in the archive pages:
`<a href="http://my-awesome-site.test/hello-world/" rel="bookmark">Hello World</a>`

Content of the link to the Author page:
`<span class="author vcard"><a class="url fn n" href=...`

From a semantic / SEO perspective, all these inline microformats show their age and don't add great value. Best case scenario, they do nothing. Worst case scenario, they conflict with other/proper/modern schema in a way which is hard to un-conflict.

All this is output by theme-specific functions that are easy to change.

It would be great to consider to set new best practices for other themes to follow and remove:
- all the `rel="bookmark"` from everywhere
- all the CSS classes related to microformats from everywhere

aristath:
Fixes #482 

## Quality assurance


carolinan:
There is a rel="author" in https://github.com/WordPress/twentytwentyone/blob/trunk/template-parts/post/author-bio.php#L20
And a rel="pingback" in https://github.com/WordPress/twentytwentyone/blob/trunk/inc/template-functions.php#L73
Is this intentional?

aristath:
`author` & `pingpack` are both valid and documented on https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel so I did not remove them, these 2 :wink:

inc/template-tags.php
```diff
@@ -47,7 +47,7 @@ function twenty_twenty_one_posted_by() {
 			printf(
 				/* translators: %s author name. */
 				esc_html__( 'By %s', 'twentytwentyone' ),
-				'<span class="author vcard"><a class="url fn n" href="' . esc_url( get_author_posts_url( get_the_author_meta( 'ID' ) ) ) . '">' . esc_html( get_the_author() ) . '</a></span>'
```
carolinan, line 50: Then I think author should also be added to this author link, that is used when there is no biography?

aristath: Done

PR #495, opened 2020-10-15
Cleanup: Remove .entry-meta

### LINKED ISSUE #494: Cleanup: Remove `.entry-meta` and related code.
The only place where we have a `.entry-meta` element is here:
https://github.com/WordPress/twentytwentyone/blob/28766adff5e8082e30b9973ac1193dd53fe30286/template-parts/header/entry-header.php#L14-L18

So it only exists inside the `template-parts/header/entry-header.php` file, and is wrapped in a `! is_page()` condition so it will only render on post-types that are not pages.

The `template-parts/header/entry-header.php` template is only called in the `template-parts/content/content-page.php` file:
https://github.com/WordPress/twentytwentyone/blob/28766adff5e8082e30b9973ac1193dd53fe30286/template-parts/content/content-page.php#L18

The `template-parts/content/content-page.php` file is only called in `page.php`:
https://github.com/WordPress/twentytwentyone/blob/28766adff5e8082e30b9973ac1193dd53fe30286/page.php#L17

So basically we have this flow: `page.php` :arrow_right: `template-parts/content/content-page.php` :arrow_right: `template-parts/header/entry-header.php`

As a result, the `.entry-meta` is **never used** and can be removed.

aristath:
Fixes #494

## Summary

Removes the `.entry-meta` element and all styles specific to that.
See description on #494 for details.

## Quality assurance


rolf-yoast (approved):
CR 👍

PR #496, opened 2020-10-15
exclude .skip-link from .site a:focus

### LINKED ISSUE #467: The skip link inherits the outline style from .site a:focus
**Describe the bug**
The skip link inherits the outline style from the general ruleset for links `.site a:focus` thus adding a rectangular shape that shouldn't be there.

To reproduce:
- select the skip to content link in the dev tools inspector
- force its state to `:focus` to make it appear
- in the dev tools Styles tab, see the ruleset from `.site a:focus ` is applied to the skip link
- toggle the outline property to see the difference

See the animated GIF below: 

skip outline

It would be nice to reset this outline to make the skip to content link look like the other skip links (skip to toolbar, logout), e.g.:

skip to toolbar

SResok:
Fixes #467 

## Summary

exclude .skip-link from .site a:focus
wrapped the `.site a:focus` css rules with a `:not()` pseudo-class 

## Quality assurance

### DIFF
--- a/assets/sass/04-elements/links.scss
+++ b/assets/sass/04-elements/links.scss
@@ -16,8 +16,11 @@ a:hover {
 }
 
 .site a:focus {
-	outline: 2px solid var(--wp--style--color--link, var(--global--color-primary));
-	text-decoration: none;
+
+	&:not(.skip-link) {
+		outline: 2px solid var(--wp--style--color--link, var(--global--color-primary));
+		text-decoration: none;
+	}
 }
 
 // Enforce the custom link color even if a custom background color has been set.
--- a/assets/sass/04-elements/links.scss
+++ b/assets/sass/04-elements/links.scss
@@ -17,9 +17,14 @@ a:hover {
 
 .site a:focus {
 
-	&:not(.skip-link) {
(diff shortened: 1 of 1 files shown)


rolf-yoast:
CR 👍

assets/sass/04-elements/links.scss
```diff
@@ -16,8 +16,11 @@ a:hover {
 }
 
 .site a:focus {
-	outline: 2px solid var(--wp--style--color--link, var(--global--color-primary));
-	text-decoration: none;
+
+	&:not(.skip-link) {
```
aristath: This would increase the specificity of all `a:focus` styles...
Could we instead add an override for the skip-link?

aristath (changes requested):

SResok:
Override outline style for the `skip-link`.
Used the same outline style as the `screen-reader-shortcut` in the wp-admin dashboard

aristath (approved):
Thank you! LGTM :+1:

carolinan:
It doesn't look exactly like the other skip links, but the rectangular shape that was inherited, has been removed.

PR #503, opened 2020-10-15
Fix styling for post-pagination

### LINKED ISSUE #499: Update font size and font weight of the post pagination
**Describe the bug**
From https://github.com/WordPress/twentytwentyone/pull/439#issuecomment-708034454

The font size and font weight of the post pagination does not match the design:

95920946-60712400-0d7e-11eb-96f5-06ff1b593c0c

aristath:
Fixes #499 

## Summary

Fixes font-sizes for post-navigation.

After I changed the font-size, I noticed that if the title is too long it flows to the next screen but doesn't have the margin so it looks a bit funky:

Screenshot_2020-10-15 Template Sticky – My Blog
Note that this is not directly related to this commit, it was just easier to spot now.

To fix that, I added `display:inline-block` to the elements and now it looks OK:

Screenshot_2020-10-15 Template Sticky – My Blog(1)

However, that change caused the title to not be underlined on hover, so I had to specifically add the hover styles.

##x Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -5907,12 +5907,12 @@ h1.page-title {
 	font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
 	font-size: 1.5rem;
 	font-weight: 600;
-	line-height: 1.3;
+	line-height: 1.7;
 }
 
 @media only screen and (min-width: 822px) {
 	.post-navigation .post-title {
-		margin: 5px 25px 0;
+		margin: 0 25px;
 	}
 }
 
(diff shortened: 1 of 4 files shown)


kjellr (approved):
LGTM

carolinan:
With the lower text being larger, I think the vertical spacing between the two texts needs to increase, there is a visible difference:

Left: New, Right: before:
spacing

aristath:
Good point. I changed the line-height to what it previously had and it's better now

Screenshot_2020-10-15 Template Sticky – My Blog(2)

melchoyce:
Hey, any chance you could revert that line-height change and go with margin or padding instead to create more breathing room?

aristath:
> Hey, any chance you could revert that line-height change and go with margin or padding instead to create more breathing room?

Done, PR on https://github.com/WordPress/twentytwentyone/pull/508 :+1:

PR #504, opened 2020-10-15
Remove space in post navigation and re-align the arrows. 

### LINKED ISSUE #501: Replace the space between the text and the arrows in the pagination with CSS
**Describe the bug**
In both the comment pagination, archive pagination and the single post pagination,
there is an extra space in the link between the text and the arrow.

From https://github.com/WordPress/twentytwentyone/pull/439#issuecomment-708034454

_Remove the space, and add 5px left/right margins (using the CSS spacing variables) onto the SVGs. 
Then, calculate the margin from the width of the SVG + the 5px of spacing. 
Or: 
Remove the inline SVGs from inside of the link, and instead put them into the before/after for each link._

space

kjellr:
Fixes #501. Should be compatible with #503. 

Before: 

Screen Shot 2020-10-15 at 10 18 00 AM

After: 

Screen Shot 2020-10-15 at 10 17 13 AM

### DIFF
--- a/assets/sass/06-components/pagination.scss
+++ b/assets/sass/06-components/pagination.scss
@@ -53,11 +53,13 @@
 	.nav-previous .svg-icon,
 	.prev .svg-icon {
 		top: -2px;
+		margin-right: calc(0.25 * var(--global--spacing-unit));
 	}
 
 	.nav-next .svg-icon,
 	.next .svg-icon {
 		top: -1px;
+		margin-left: calc(0.25 * var(--global--spacing-unit));
 	}
 }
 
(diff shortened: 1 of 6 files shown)


assets/sass/06-components/pagination.scss
```diff
@@ -24,6 +24,13 @@
 	}
 
 	.nav-links {
+
+		.nav-next a,
+		.nav-previous a {
+			display: flex;
+			flex-direction: column;
+		}
```
kjellr, line 32: I had to switch to using flexbox here to override some annoying alignment issues with the next post item: 

Before|After
---|---
Screen Shot 2020-10-15 at 10 17 26 AM|Screen Shot 2020-10-15 at 10 17 48 AM

aristath (approved):
LGTM :+1:

Issue #458, opened 2020-10-14
Focus indicator for input fields and select lists


carolinan:
**Describe the bug**

Quote from @sarahricker:
_keyboard focus on the search box should also have a focus state, but it seems the input box is getting double boarder while the button gets an inset dashed line._

See https://github.com/WordPress/twentytwentyone/issues/444

If I understood correctly, the focus indicator for text input fields like in the comment form text fields and text area, password and search input field is not enough.

![input-focus](https://user-images.githubusercontent.com/7422055/95946618-36a31600-0ded-11eb-83ce-057a9581c10d.gif)

Visually when an input field is in focus, the background color is changed to white, the border is thicker, and there is a cursor marker. 

The background color change itself is minor and does not comply with any contrast requirements, 
it is decorative, and it is not added to select lists where there is already text content.

Technically, the larger "growing border" is an outline, and for some fields this is combined with an offset.

```
input[type="search"]:focus {
    outline-offset: 0;
}
input[type="text"]:focus, input[type="email"]:focus, input[type="url"]:focus, input[type="password"]:focus, input[type="search"]:focus, input[type="number"]:focus, input[type="tel"]:focus, input[type="date"]:focus, input[type="month"]:focus, input[type="week"]:focus, input[type="time"]:focus, input[type="datetime"]:focus, input[type="datetime-local"]:focus, input[type="color"]:focus, .site textarea:focus {
    color: var(--form--color-text);
    outline: 1px solid var(--form--border-color);
    background: var(--global--color-white);
}
```
See:
https://github.com/WordPress/twentytwentyone/blob/trunk/assets/sass/04-elements/forms.scss#L23

carolinan:
Example where the outline is changed from solid to dotted to look more like the default, and an offset of 4px is added:

![comment-form-focus](https://user-images.githubusercontent.com/7422055/95946957-e9737400-0ded-11eb-89c4-b7c8268d05d1.png)

And an example with a dotted outline with no offset:
![comment-form-focus-no-offset](https://user-images.githubusercontent.com/7422055/95947024-0ad46000-0dee-11eb-8336-4dd426c0892c.png)


To me the first version looks OK, but the version without the offset is more difficult for me to see than the existing style with the growing border.

carolinan:
But the search widget input field does not look as good with the 4px offset because of the button position:

[...]

Issue #444, opened 2020-10-13
Link styles: consider using the default focus outline style set by the browsers


marctison75:
Following Jeff Ong’s suggestion to create a new issue, I’m opening this one based on the discussion of  #165.

The current link styles don’t look good on Safari. 

![Safari](https://user-images.githubusercontent.com/72606415/95842202-8c70b300-0d46-11eb-903d-0cf82c28bc15.png)

In Firefox on Windows, you can see how the outline can interact with the shape of the letter and decrease the readability of the text.

![Firefox](https://user-images.githubusercontent.com/72606415/95842250-9a263880-0d46-11eb-96d8-3c3484a3a52d.jpg)

A solution might be to use the default focus style set by the browsers (user agent stylesheet), which is already an outline (https://www.w3.org/ is using it). This solution is close to the initial design and doesn’t have any issues.

sarahricker:
Thanks for pointing this out @marcduluc!

@melchoyce a11y and Design discussed this one in our new weekly office hours. Overall, we don’t think the browser’s default outline style is enough. We can definitively design something better. It seems though, that this issue might be caused by the larger border width, possibly also line-height and the wrapping concern. 

Am I right this was intended to match the other borders on the site (like the search box)? Asking, because keyboard focus on the search box should also have a focus state, but it seems the input box is getting double boarder while the button gets an inset dashed line.

Seems we need a more global focus style that would better fit all concerns, but I'm open to ideas!

carolinan:
There is not a lot of time left to redesign the links again. Please be mindful that it is late into the project now.
For example the last call for features was yesterday.

sarahricker:
Gotcha, thanks @carolinan! Maybe we can limit the focus here to the cross-browser bugs.

enriquesanchez:
I think we can try some small variations that should improve the focus style, I don't think these to be big changes but I'll leave it to the project leads to consider.

**Use the default browser focus style**

![2020-10-14 10-02-28 2020-10-14 10_03_04](https://user-images.githubusercontent.com/3276087/96008486-3adb2d80-0e05-11eb-8b46-c1fdd0ed1f40.gif)

**Change the thickness of the underline**

![2020-10-14 09-59-56 2020-10-14 10_00_51](https://user-images.githubusercontent.com/3276087/96008571-50e8ee00-0e05-11eb-9879-b1388d65f6b8.gif)

**Use a background color**

[...]

PR #507, opened 2020-10-15
Try the default outline

carolinan:
This is a test of  #444 where the default outlines are used for links.
And for #458

## Summary
Removes the 2px solid outline.

## Test instructions

1. Use tab to move between links and other focusable elements
1.
1.

## Screenshots

Firefox on PC, Windows 10
outline-black
outline

overlap-outline2

overlap-outline

Chrome on PC Windows 10
outline-chrome

## Quality assurance

### DIFF
--- a/assets/sass/04-elements/links.scss
+++ b/assets/sass/04-elements/links.scss
@@ -16,7 +16,6 @@ a:hover {
 }
 
 .site a:focus {
-	//outline: 1px solid var(--wp--style--color--link, var(--global--color-primary));
 	text-decoration: none;
 }
 
--- a/assets/sass/04-elements/links.scss
+++ b/assets/sass/04-elements/links.scss
@@ -16,7 +16,7 @@ a:hover {
 }
 
 .site a:focus {
-	outline: 2px solid var(--wp--style--color--link, var(--global--color-primary));
+	//outline: 1px solid var(--wp--style--color--link, var(--global--color-primary));
 	text-decoration: none;
 }
 
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -383,7 +383,6 @@ a:hover {
 }
 
 .site a:focus {
-	outline: 2px solid #28303d;
 	text-decoration: none;
 }
 
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -2086,7 +2086,6 @@ a:hover {
 }
 
 .site a:focus {
-	outline: 2px solid #28303d;
 	text-decoration: none;
 }
 
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -1819,6 +1819,7 @@ select {
 }
 
 select:focus {
+	outline-offset: 4px;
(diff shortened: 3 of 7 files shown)


assets/sass/04-elements/links.scss
```diff
@@ -16,7 +16,7 @@ a:hover {
 }
 
 .site a:focus {
-	outline: 2px solid var(--wp--style--color--link, var(--global--color-primary));
+	//outline: 1px solid var(--wp--style--color--link, var(--global--color-primary));
```
carolinan, line 19: (This is just a test to allow people to try the default to see which would work best.)

carolinan:
This screenshot shows that the default focus outline will not work for form input fields in Firefox:

form-firefox

And is almost as invisible in chrome:
form-chrome


An outline offset of 4px in Firefox vs Chrome:
form-offset-firefox
form-offset-chrome

And with 1px solid outline and 4px offset,
Left: Chrome, right: Firefox

1px-solid-4pxoffset

carolinan:
For the search, try an inset that will match the button better and may be more visible than the growing border.

I am not sure what accessibility impacts this has, there is still a lot of space between the edge of the field and the text:
search-inset
search-inset-text

melchoyce:
@sarahricker @enriquesanchez would love your feedback on this!

afercia (commented):
Personally, I wouldn't be opposed to use the browsers native focus styles. They give the advantage they can easily be overridden by users who use custom stylesheets for their special needs and by assistive technology tools that augment the focus styles.

I see a couple issues though:

1: technical
As pointed out by @carolinan, the form elements would need some special focus style anyways. It is also worth mentioning that outline-offset doesn't work in IE11 so there would be the need for an alternate solution for that browser.

2: general
From a theme that aims to be level AAA "ready", I would expect something better than the native focus styles. Websites that are strongly focused on accessibility do use very strong focus styles. One example for all: https://www.gov.uk/ 
If the intent is still to make Twenty Twenty-One "AAA ready", I would expect very strong focus styles. Otherwise, I'm afraid it would be inappropriate to "sell" the theme as "AAA ready". Which is not necessarily bad _per se_.  Twenty Twenty-One can still be a highly accessible theme but not declare any compliance with WCAG conformance levels.

carolinan:
For the **link** focus styles, perhaps reduce the thickness from 2px to 1 but keep it to the primary color and solid?

melchoyce:
@kjellr and I are going to explore some more explicit focus styles next week — we'll try to get this finalized by Beta 2.

enriquesanchez:
>the form elements would need some special focus style anyways

I agree with @afercia and mentioned it on https://github.com/WordPress/twentytwentyone/issues/458#issuecomment-709629586. The only focus style we need to find a better style for is for text links.

As for alternatives to the outline style, I added some suggestions here https://github.com/WordPress/twentytwentyone/issues/444#issuecomment-708468634. 

@kjellr @melchoyce Happy to help explore solutions.

carolinan:
Closing this, a new PR can be done when there is a new style for focused links.

Issue #184, opened 2020-09-30
Implement the design of the quote block


carolinan:
Current design: 
![quote_](https://user-images.githubusercontent.com/7422055/94671989-e750e600-0314-11eb-8055-cb744631e61e.png)

In comparison to the pull quote these blocks only have one width and no color settings.
Related files:
https://github.com/WordPress/twentytwentyone/tree/trunk/assets/sass/05-blocks/quote

carolinan:
New design:
Figma:
https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=1254%3A0

![94748377-2865ec00-034f-11eb-823e-d63502c6eb35](https://user-images.githubusercontent.com/7422055/94779717-b1b90500-03c7-11eb-9d3c-c778422c1718.png)

paaljoachim:
I am thinking an issue in the Gutenberg github repo would be fitting, so that we can add alignment and the additional color option located in the Pullquote block into the Quote block.

Using Twenty Twenty theme.
Chrome.
Localhost dev site.

Quote Block.
![Quote-Block](https://user-images.githubusercontent.com/5323259/95395159-39d66780-08fe-11eb-9bd4-ed0be04d7abe.png)

Pullquote block.
![Pullquote-Block](https://user-images.githubusercontent.com/5323259/95395394-c84ae900-08fe-11eb-9e6e-be076db78d83.png)

carolinan:
A quote block does not have anything to add the "main" color to. Pull quote has borders by default.

I don't think it is something the editor team should prioritize because of this theme.

carolinan:
The quote symbol itself should use the same symbol as used in the pull quote:

For refrence:
```
&::before {
	color: currentColor;
	content: "\201C";
	display: block;
	font-size: 3rem;
	font-weight: 500;
	line-height: 1;
}
```
https://github.com/WordPress/twentytwentyone/blob/trunk/assets/sass/05-blocks/pullquote/_style.scss#L14

paaljoachim:
Btw
A PR is being worked on:
**Quote: convert to nested blocks**
https://github.com/WordPress/gutenberg/pull/25892

Related to this issue: https://github.com/WordPress/gutenberg/issues/15486

JustinyAhin:
Thanks Carolina. 
Saw it this morning.

[...]

PR #453, opened 2020-10-13
Fix/implement block quote design

JustinyAhin:
Fixes #184.

## Relevant technical choices:
No.

## Test instructions

Add two block quotes in the editor
In frontend, they should look like the model in this figma file https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=1254%3A0.

## Screenshots

Quote block in the editor before
block-quote-editor-before

Quote block in the editor after
block-quote-editor-after

Quote block in frontend before
block-quote-frontend-before

Quote block in frontend after
block-quote-frontend-after

## Quality assurance
* [x] I have thought about any security implications this code might add.
* [x] I have checked that this code doesn't impact performance (greatly).
* [x] I have tested this code to the best of my abilities
* [x] I have checked that this code does not affect the accessibility negatively

PR #513, opened 2020-10-15
Try adding blockquote styles

kjellr:
Alternate to #184. I spent a little time giving this a try too, and got it pretty close. 

Screen Shot 2020-10-15 at 4 04 10 PM

Should work well with left/center/right-aligned quotes. The only bug I know of at the moment is that the extra margin I add on small screen sizes to house the blockquotes is eaten up by higher-specificity global margins on the front end:

Screen Shot 2020-10-15 at 4 07 00 PM

I won't be able to work on solving that until tomorrow morning, so if someone else would like to pick this up feel free. If we get #184 cleaned up first, that's fine too. Thanks!


ryelle:
I've iterated on this a bit to fix up the small screen quotes and some issues in the editor. The only question I have left is whether the de-bolding of `strong` should also affect the `cite` content — I've talked this through with @melchoyce and toggling bold on in the quote correctly unbolds the content, but if you bold something in the citation it does nothing right now.

Some screenshots from Mac/Firefox:

<details>
<summary>Front end, desktop</summary>

Screen Shot 2020-10-15 at 19 00 15-fullpage

</details>

<details>
<summary>Editor, desktop</summary>

Screen Shot 2020-10-15 at 19 02 01-fullpage

</details>

<details>
<summary>Front-end, mobile</summary>

Screen Shot 2020-10-15 at 19 11 51-fullpage

</details>

ryelle:
@melchoyce mentioned the citations should not be italics, so I've pushed a fix for that; and I also noticed that things were out of alignment in the editor on a wide screen (wider than my screenshot above), so I've fixed that too.

melchoyce (approved):
LGTM 👍

carolinan:
This is a comparison between this PR (right) and https://github.com/WordPress/twentytwentyone/pull/453

quote

And I think they both have _different_ issues, with the alignments, weight, and the citation.

kjellr:
Thank you for the improvements, @ryelle!

> I think they both have different issues, with the alignments, weight, and the citation.

Regarding the alignments, this PR intentionally leaves off the quotation mark accent on the center version, and right-aligns it when the quote is right aligned. I prefer that to the treatment in #453. 

This PR uses a 600 font weight for the text instead of 700. The theme uses 600 for bold headings and the pagination titles and that's what I pulled that from, but I see that the pullquote block and widget font titles use a 700 weight. We may want to align those, but I'll leave it to @melchoyce to make the final call on that. 

It looks like #453 adds a hyphen + a non-breaking space before the citation (If we go with that, typography rulebooks typically suggest it should be an em-dash instead of a hypen). In general though, I personally prefer letting users enter that themselves.

carolinan:
-Then let's use this one.

I can't take preference in consideration if it is not documented why decisions were made.

melchoyce:
This PR more accurately reflects the design I intended.

> This PR uses a 600 font weight for the text instead of 700. The theme uses 600 for bold headings and the pagination titles and that's what I pulled that from, but I see that the pullquote block and widget font titles use a 700 weight. We may want to align those, but I'll leave it to @melchoyce to make the final call on that.

Let's go with 700 👍 

> It looks like #453 adds a hyphen + a non-breaking space before the citation (If we go with that, typography rulebooks typically suggest it should be an em-dash instead of a hypen). In general though, I personally prefer letting users enter that themselves.

Yeah, @ryelle asked me about this too, and I said I'd rather leave it up to people to add it if they want it.

kjellr:
> Let's go with 700

Updated!

carolinan:
The conflict needs to be resolved, but I am not sure what else this PR is waiting for?

aristath (approved):
Code looks good to me, and since design has already approved it I believe this is OK to merge :+1:

aristath:
Resolved merge conflicts and merged

PR #522, opened 2020-10-16
Add spaces where needed for screen-readers

### LINKED ISSUE #517: Texts missing spaces when using a screen reader
**Describe the bug**
These texts are missing spaces when using a screen reader:

Parent menu item + description
Continue reading + post title
Edit link + post title

Can also be improved:
Previous post  + post title, Next post + post title
has no spacing after the colon, reads like: Previous post:title

**Additional context**
Tested with NVDA on windows 10

aristath:
Fixes #517 

## Summary

Added spaces where needed.
While doing so, I fixed code that didn't make sense (for example removed a few `wp_kses` calls) or was just hard to read & understand.

## Quality assurance

### DIFF
--- a/inc/menu-functions.php
+++ b/inc/menu-functions.php
@@ -90,7 +90,8 @@ add_filter( 'walker_nav_menu_start_el', 'twenty_twenty_one_nav_menu_social_icons
 function twenty_twenty_one_add_menu_description_args( $args, $item, $depth ) {
 	$args->link_after = '';
 	if ( 0 === $depth && isset( $item->description ) && $item->description ) {
-		$args->link_after = '<span class="menu-item-description"><span>' . $item->description . '</span></span>';
+		// The extra <span> element is here for styling purposes: Allows the description to not be underslined on hover.
+		$args->link_after = ' <span class="menu-item-description"><span>' . $item->description . '</span></span>';
 	}
 	return $args;
 }
--- a/inc/menu-functions.php
+++ b/inc/menu-functions.php
@@ -91,7 +91,7 @@ function twenty_twenty_one_add_menu_description_args( $args, $item, $depth ) {
 	$args->link_after = '';
 	if ( 0 === $depth && isset( $item->description ) && $item->description ) {
 		// The extra <span> element is here for styling purposes: Allows the description to not be underslined on hover.
-		$args->link_after = ' <span class="menu-item-description"><span>' . $item->description . '</span></span>';
+		$args->link_after = '<p class="menu-item-description"><span>' . $item->description . '</span></p>';
 	}
 	return $args;
 }
--- a/inc/menu-functions.php
+++ b/inc/menu-functions.php
@@ -90,7 +90,7 @@ add_filter( 'walker_nav_menu_start_el', 'twenty_twenty_one_nav_menu_social_icons
 function twenty_twenty_one_add_menu_description_args( $args, $item, $depth ) {
 	$args->link_after = '';
 	if ( 0 === $depth && isset( $item->description ) && $item->description ) {
-		// The extra <span> element is here for styling purposes: Allows the description to not be underslined on hover.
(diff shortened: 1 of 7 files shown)


inc/menu-functions.php
```diff
@@ -90,7 +90,8 @@ function twenty_twenty_one_nav_menu_social_icons( $item_output, $item, $depth, $
 function twenty_twenty_one_add_menu_description_args( $args, $item, $depth ) {
 	$args->link_after = '';
 	if ( 0 === $depth && isset( $item->description ) && $item->description ) {
-		$args->link_after = '<span class="menu-item-description"><span>' . $item->description . '</span></span>';
+		// The extra <span> element is here for styling purposes: Allows the description to not be underslined on hover.
```
afercia: Typo: underslined

aristath: :laughing: thank you, fixed

template-parts/content/content-page.php
```diff
@@ -38,16 +38,9 @@
...
-						/* translators: %s: Name of current post. Only visible to screen readers */
-						__( 'Edit <span class="screen-reader-text">%s</span>', 'twentytwentyone' ),
-						array(
-							'span' => array(
-								'class' => array(),
-							),
-						)
-					),
-					get_the_title()
+					/* translators: %s: Name of current post. Only visible to screen readers */
```
afercia: Minor, pre-existing, issue: translators comments should always end with a period.

aristath: Good point. I took the opportunity and fixed it for all translators comments :+1:

afercia:
I'm not sure the `&nbsp;` thing is really needed. Seems to me there's now a double space (see screenshot below).

I'm not sure I understand why a string that does have a space e.g. `Edit %s` shouldn't be read out correctly by NVDA and in my testing NVDA does gets the space correctly. I do understand the menu description case but I'm not sure about the other cases.

In which scenario the space wasn't taken into consideration? Which browser has been used together with NVDA?

Screenshots from this branch: there's a double space between "Edit" and "Hello". Reading character by character makes the double space evident:

Screenshot (158)

Screenshot from trunk with disabled styles: there is a space between "Edit" and the post title:


Screenshot 2020-10-16 114059


Maybe I'm missing something 🙂

aristath:
@afercia I'm currently on Linux so I couldn't test this myself, @carolinan reported it so she'll have all the details. I think the issue was on Windows with NVDA, and it was only happening on Firefox or Chrome, one of the 2 had an issue :thinking: 
It looks like maybe the `<span>` element is what causes it

carolinan:
I tested with NVDA version 2020.3 on Windows 10 with Chrome version 86.0.4240.75 where the space was not read correctly.
If it matters, my Windows install is in Swedish, so so is NVDA.

In Firefox there is now a double space.

afercia:
I'm on Windows too and I tested Firefox+NVDA (they get the space).  Will check with Chrome but it might be a Chrome issue. Generally, Firefox+NVDA are the best combo, while Chrome is still "so and so" on a11y related things

aristath:
I wouldn't mind removing the `&nbsp;` things... They make no sense at all, and they are there basically to address what I'd categorize as a Chrome bug.
If NVDA is mostly used with Firefox, then I'd say we can remove them :+1:

afercia:
Can confirm that with Chrome both NVDA and JAWS 2019 read the string as if the space wasn't there. However, I think this is a Chrome (or, better: Chromium) thing and I'm not sure we should fix a browser quirk.

It is weird though because it seems the accessible name of the links does have the space (see the dev tools accessibility tab):

Screenshot (159)

It can be reproduced with the Opera browser as well, which is powered by Chromium.

aristath:
Thank you. I agree 100%, the theme should not try to fix browser bugs. I removed the `&nbsp;` hacks so this looks a lot cleaner now :+1:

carolinan (approved):

PR #525, opened 2020-10-16
add export-ignore files & folders

aristath:
Makes it easier to get a working copy of the theme without all the dev files.
The package is then ready for inclusion in wp-core directly from the download link

Screenshot_2020-10-16 WordPress twentytwentyone

## Quality assurance

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,3 +12,26 @@
 *.jpg binary
 
 *.map linguist-generated
+
+.editorconfig export-ignore
+.eslintignore export-ignore
+.eslintrc export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
+.stylelintignore export-ignore
+.stylelintrc.json export-ignore
+.wp-env.json export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -12,3 +12,26 @@
...
+package-lock.json export-ignore
+package.json export-ignore
+phpcs.xml.dist export-ignore
+phpstan.neon export-ignore
+postcss.config.js export-ignore
+README.md export-ignore
+style.css.map export-ignore
+.github/ export-ignore
+assets/css/*.map export-ignore
+assets/sass/ export-ignore
```
kishanjasani, line 37: Please also exclude the `readme.txt` file.

kishanjasani: Don't know `readme.txt` is required or not others LGTM:+1:

aristath: readme.txt is required for themes hosted in the w.org repository, that's why it wasn't added.

Thank you for the review!

PR #528, opened 2020-10-16
Added dark theme preference

### LINKED ISSUE #410: Add support for prefers color scheme
As our goal is to create an accessible theme, the theme should support prefers-color-scheme.

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

I suggest adding a dark scheme at least, but it does not need to be black, using one of the existing darker color palettes would be enough.

rolf-yoast:
Fixes https://github.com/WordPress/twentytwentyone/issues/410

## Summary

This PR will add the dark mode preference to the theme

## Test instructions

1. Set the OS color preference to dark.
1. Check if the default color is selected in the customizer that the theme is dark.

## Quality assurance

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -32,6 +32,17 @@
 	/* Widgets */
 }
 
+@media (prefers-color-scheme: dark) {
+	:root .has-default-light-palette-background {
+		background-color: #28303d;
+	}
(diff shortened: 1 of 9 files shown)


rolf-yoast:
CR & Acceptance 👍

PR #529, opened 2020-10-16
Increase clickable area for list links.

aristath:
## Summary
Fixes an a11y issue reported by lighthouse, where the clickable area for links inside lists in widget areas is not large enough.
Visually there is no real difference, but it's easier to target and click the links.

## Quality assurance


carolinan (changes requested):
The added padding is being overwritten by WordPress recent comment style
`recent_comments_style`

https://developer.wordpress.org/reference/classes/wp_widget_recent_comments/recent_comments_style/


For other listed links in widgets, the clickable area is larger, but not 44px44px.

aristath:
> The added padding is being overwritten by WordPress recent comment style

Fixed (added filter)

> For other listed links in widgets, the clickable area is larger, but not 44px44px.

I don't think that rule applies to text... :thinking:

carolinan:
> I don't think that rule applies to text... 🤔

Then what did lighthouse complain about 🤔 -I'll test again.

carolinan:
Before:
before


After:
after

enriquesanchez:
@aristath I wasn't able to reproduce this with Lighthouse, would you have more info on the reported issue? 

I honestly find the list harder to read with the increased padding, it breaks the reading flow between items.

aristath:
I completely forgot about this PR... Yes,you are right,  it does indeed make it harder to read. I'm closing this since it looks like we'll increase the font-size instead (https://github.com/WordPress/twentytwentyone/issues/578)

PR #530, opened 2020-10-16
Eagerly load the thumbnail on single pages

jdevalk:
## Summary
As the thumbnail is going to be in the viewport immediately, we should make it load eagerly.

## Test instructions

1. Create a post that has a featured image.
1. See that `loading="eager"` is set on that image's `img` tag.

## Screenshots

No visual change.

## Quality assurance
* [x] I have checked that this code doesn't impact performance (greatly). --- It improves it.


rolf-yoast:
CR & A 👍

PR #556, opened 2020-10-16
Update default form label weight

melchoyce:
## Summary
I noticed we have two form labels weights. `-form--label-weight` which is 400, and
`-form--label-weight-strong` which is 500. We only need the one, 500. I removed `-form--label-weight-strong` and switched the default value to 500.

## Test instructions

1. Go to a missing page and check what weight the label is
2. Go to a password protected post and check what weight the label is
3. Check the weight of the search form label

## Quality assurance
* [ X ] I have thought about any security implications this code might add.
* [ X ] I have tested this code to the best of my abilities
* [ X ] I have checked that this code does not affect the accessibility negatively

### DIFF
--- a/assets/css/style-editor.css
+++ b/assets/css/style-editor.css
@@ -93,8 +93,7 @@
 	--form--line-height: var(--global--line-height-body);
 	--form--color-text: var(--global--color-dark-gray);
 	--form--color-ranged: var(--global--color-secondary);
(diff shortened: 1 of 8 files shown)


kjellr (approved):
Makes sense!

ryelle (approved):

Issue #533, opened 2020-10-16
Pagination links: the clickable area is too small


afercia:
**Describe the bug**
The clickable area of the various pagination links is a bit too small for a theme that aims to be WCAG level AAA "ready".

The posts pagination links have a target size of approximately 13 by 42 pixels:

![posts pagination](https://user-images.githubusercontent.com/1682452/96251166-5b93b680-0fb0-11eb-951d-c62f7d126e3c.png)

where the width is a bit too small for all users.

Similarly, the links of a paginated post have a clickable area that is too small:

![paginated post](https://user-images.githubusercontent.com/1682452/96251301-87af3780-0fb0-11eb-8158-08249c0e5d76.png)

Worth reminding that the WCAG Target Size (level AAA) success criterium requires a target size of 44 by 44 pixels, with some exceptions: https://www.w3.org/TR/WCAG21/#target-size

carolinan:
With #524, the clickable size of the post pagination is 71x41 which is still not high/tall enough.

JustinyAhin:
I think this has totally been solved by #570. Archives pages and search pages pagination have the text "Page" before the number of the page.

@carolinan @aristath?

aristath:
They will still need to be at least 3px taller :wink: 
I haven't tested it but believe a small vertical padding in the links should fix this :+1:

sarahricker:
@JustinyAhin @carolinan Looks like we need just a bit more. Possible to get this in before Beta 1?

JustinyAhin:
> @JustinyAhin @carolinan Looks like we need just a bit more. Possible to get this in before Beta 1?

Commit coming in a couple of minutes @sarahricker

JustinyAhin:
Let's fix that in Beta 2

afercia:
Question: are there languages where the translation of "Page" would be a shorter word? Not sure about Western languages but it's likely the case for some Asian languages?

In any case, I'm not sure we can rely on the space taken by the word "Page" to make sure these links are at least 44 by 44 pixels. I'd consider to add a min-width and min-height anyways.

Issue #475, opened 2020-10-14
Posts pagination numbers are just numbers


afercia:
**Describe the bug**
The Posts pagination numbers are just numbers, thus they're announced by screen readers just as, for example:

"2 link", "3 link", etc.

![Screenshot (153)](https://user-images.githubusercontent.com/1682452/95979995-ce6c2880-0e1c-11eb-9fff-60c01d9c6ed7.png)

Other bundled themes, for example Twenty Seventeen, add some visually hidden text to prepend the word "Page" and make the link text more meaningful for screen reader users:

![Screenshot (154)](https://user-images.githubusercontent.com/1682452/95980014-d4620980-0e1c-11eb-9fb7-c9c35f16cf62.png)

However, this would impact sighted speech recognition users. They would see a link with a visible text "2" and they would try to voice a command like "Click 2". The command would fail because the actual accessible name of the link would be "Page 2".

The only option to cover all needs would be to make the "Page" text visible, for example:

![page](https://user-images.githubusercontent.com/1682452/95987854-4855df00-0e28-11eb-9f2a-67733fc96cd5.png)

I do realize this would impact design, adding the Design label.

enriquesanchez:
We could place these links inside a `role=group` so they are announced with context to screen readers:

"link 2, Posts pagination, group"

<img width="772" alt="Screen Shot 2020-10-14 at 12 00 04" src="https://user-images.githubusercontent.com/3276087/96021515-2acb4a00-0e15-11eb-806b-6278d8fa0cbc.png">

Another similar idea is to have these links `aria-describedby` by a heading, label or hidden label.

If we do this both the design and users of voice recognition should not be affected.

ryelle:
The links are already in a `nav role="navigation"` with label "Posts", but that label isn't announced with each link. Would adding a group to just the numbers work better? There is also already a hidden heading ("Posts navigation"), we could attach that using `aria-describedby`, but would that be picked up by the link list menu?

enriquesanchez:
@ryelle Ah, you're right. `role="navigation"` makes more sense here, and this means the link is announced as:

"Link 2, Posts, navigation"

I personally think this is enough context.

>but that label isn't announced with each link

Right, VoiceOver only announces the `role` on the first item you focus on in the list. As far as I can tell, this is standard behavior and avoids unnecessary verbosity and repetition.

I think we could still change the label to "Posts pagination" so it's a bit clearer what the navigation is about. What do you think?

[...]

Issue #532, opened 2020-10-16
Sticky post isn’t visually different than the other posts


briceduclos:
**Describe the bug**
 Currently, a sticky post looks like a “normal” post. There’s no indication that it’s sticky.

**Expected behavior**
The sticky post should have a text indicating that it’s sticky or a different styling that distinguishes it from the other posts.

carolinan:
It is different from the others because it is placed at the top, out of order.

briceduclos:
All previous default themes (Twenty Twenty, Twenty Nineteen, Twenty Seventeen) and Seedlet add a label indicating to the user that the post is sticky/featured (similar to Twitter). For a website visitor, I think it can be weird not to show anything. Let’s say your sticky post is dated from 2019 and that you’ve written other posts in 2020. When visiting your posts page, the visitor might think that you haven’t published any post since 2019 and leave your blog as he thinks there’s nothing new to read.

melchoyce:
We could stick (hah!) it into the footer meta on archive pages?

What label do you think makes the most sense — Pinned? Sticky post? Featured?

aristath:
<details>
<summary>Edit: Comment hidden 'cause it looked worse after I looked at for a couple of minutes. (click to expand)</summary>

We could add something like this...
```css
.sticky {
  background: rgba(0,0,0,0.05);
  padding-top: 1em;
}
```
The result would look like this:
![Screenshot_2020-10-19 My Blog(2)](https://user-images.githubusercontent.com/588688/96404096-b3b1ff00-11e2-11eb-93a0-90d81762bb8f.png)

</details>

briceduclos:
@melchoyce Facebook uses "Pinned Post" and Twitter uses "Pinned Tweet"

carolinan:
But it is not called pinned in the WordPress editor: "Stick to the top of the blog"

melchoyce:
Could this work?

![image](https://user-images.githubusercontent.com/2846578/96515046-9d339400-1232-11eb-8a36-31ca2602ae64.png)

melchoyce:
If not, we could show this on archive pages, and make no changes on single:

![image](https://user-images.githubusercontent.com/2846578/96515115-c2c09d80-1232-11eb-8cdd-7ffa9dfcc594.png)

melchoyce:
Semi-related: #550

PR #570, opened 2020-10-17
Update content pagination to use a nav and increase the spacing

### LINKED ISSUE #557: clearing floats
When the user content has a floated element, it is not cleared before outputting the page links.

**Expected behavior**
The user content can contain floated elements, which should be cleared before anything else is output. Page links should not be inside the entry-content div, even though they are specified there.

**Screenshots**
This shows that the page links are wrapping around the floated image.
Clearing Floats – Local Current WP-34

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/557

Partial fix for https://github.com/WordPress/twentytwentyone/issues/533
Follow up on https://github.com/WordPress/twentytwentyone/issues/475

## Summary

Replaced the wrapping div with a nav element with an aria label that says "Page".
Cleared the element using CSS.
Removed the span with the text that said "Pages:" 
-I felt that having the text read: `Pages: Page 1 Page 2 Page 3` was redundant, but it also means the pagination looks different than what we are used to. :thinking:

Added the text "Page" before each link
Added spacing that is more consistent with the current older/newer posts pagination, but has a larger click target area of
minimum 59x 46px.

-The pagination was not the same for the image attachment and other types of content. 
This PR makes it more consistent.

## Test instructions

1. View a page or post that has pagination, for example "Template: Paginated" in the Theme Unit Test.
1.
1.

## Screenshots

If this is a visual change please include screenshots of the change at various screen sizes.

<details>
<summary>Desktop (Extra Large)</summary>

Before:
96293380-43239c00-0fb0-11eb-88ef-3efaa16e3fa1

paginated-content-before

After:

clear-float-after

paginated-content-after

</details>

## Quality assurance

### DIFF
--- a/assets/sass/06-components/posts-and-pages.scss
+++ b/assets/sass/06-components/posts-and-pages.scss
@@ -2,15 +2,21 @@
 	// This class is required to pass ThemeCheck.
 }
 
-.sticky-post {
-	color: var(--global--color-background);
-	background-color: var(--global--color-secondary);
-	font-family: var(--global--font-secondary);
-	font-size: var(--global--font-size-xs);
-	line-height: 1;
-	padding: calc(0.25 * var(--global--spacing-unit)) calc(0.33 * var(--global--spacing-unit));
-}
-
 .no-results.not-found > *:first-child {
 	margin-bottom: calc(3 * var(--global--spacing-vertical));
+}
+
+// Styling for wp_link_pages.
+.page-links {
+	clear: both;
+
+	.post-page-numbers {
+		margin-left: calc(0.66 * var(--global--spacing-unit));
+		margin-right: calc(0.66 * var(--global--spacing-unit));
+		padding: calc(0.33 * var(--global--spacing-vertical)) 0;
+
+		&:first-child {
+			margin-left: 0;
+		}
+	}
 }
--- a/image.php
+++ b/image.php
@@ -40,12 +40,10 @@ while ( have_posts() ) {
 
 			wp_link_pages(
 				array(
-					'before'      => '<div class="page-links"><span class="page-links-title">' . esc_html__( 'Pages:', 'twentytwentyone' ) . '</span>',
-					'after'       => '</div>',
-					'link_before' => '<span>',
-					'link_after'  => '</span>',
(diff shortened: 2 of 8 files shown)


assets/sass/06-components/posts-and-pages.scss
```diff
@@ -2,15 +2,21 @@
 	// This class is required to pass ThemeCheck.
 }
 
-.sticky-post {
```
carolinan, line 5: I removed this for now because it is not used by the theme. It is a left over from the "featured" label that is used in Seedlet. Can be reconsidered for  https://github.com/WordPress/twentytwentyone/issues/532

rolf-yoast:
CR 👍

aristath (approved):
LGTM :+1:

melchoyce:
LGTM 👍

image.php
```diff
@@ -40,12 +40,10 @@
...
-					'before'      => '<div class="page-links"><span class="page-links-title">' . esc_html__( 'Pages:', 'twentytwentyone' ) . '</span>',
-					'after'       => '</div>',
-					'link_before' => '<span>',
-					'link_after'  => '</span>',
-					'pagelink'    => '<span class="screen-reader-text">' . esc_html__( 'Page', 'twentytwentyone' ) . ' </span>%',
-					'separator'   => '<span class="screen-reader-text">, </span>',
+					'before'   => '<nav class="page-links" aria-label="' . esc_attr__( 'Page', 'twentytwentyone' ) . '">',
+					'after'    => '</nav>',
+					/* translators: There is a space after page. */
+					'pagelink' => esc_html__( 'Page ', 'twentytwentyone' ) . '%',
```
ryelle, line 46: I think the % should be in each of the translation strings, so that the number & text can be controlled by the translator (in case there's a language where "3 Page" makes more sense)

aristath: > I think the % should be in each of the translation strings, so that the number & text can be controlled by the translator (in case there's a language where "3 Page" makes more sense)

+1, makes sense. Good catch!
Should be in a new PR since this one was already merged. Apologies, I missed your comment by a few seconds.

PR #577, opened 2020-10-19
Use variable instead of function

mukeshpanchal27:
We have to use variable instead of calling function each time to reduce query.


aristath (changes requested):
I agree this would be more performant and since we use the same thing everywhere a var makes sense.
Please fix the PHP fatalities (extra closing parenthesis) and then this can be merged

mukeshpanchal27:
Thanks for the feedback.

Requested changes done and updates patch.

Let me know if anything else need to changes

classes/class-twenty-twenty-one-custom-colors.php
```diff
@@ -64,15 +64,16 @@ public function custom_get_readable_color( $background_color ) {
 	public function generate_custom_color_variables( $context = null ) {
 
 		$theme_css = 'editor' === $context ? ':root .editor-styles-wrapper{' : ':root{';
```
aristath: Coding Standards: Align `=` signs.
```suggestion
		$theme_css        = 'editor' === $context ? ':root .editor-styles-wrapper{' : ':root{';
```

aristath (changes requested):

aristath (approved):

PR #580, opened 2020-10-19
Group block clear floats

### LINKED ISSUE #555: Group block not clearing float, Search block out of center
The Group block is not clearing floats. This can be seen by the fact that the floated image is hanging out the bottom of the block.
The Search block is not centered like everything else. Its button styling looks different from the Search widget I have in the footer.

**Screenshots**
This is showing the front end, with a Group block containing an aligned right image, then a search block, and then the Recent Comments block (which is the only one set to alignwide).
group-and-search

aristath:
Fixes #555 

## Summary
Initially I added `display: flow-root;` to group blocks. That worked fine, but then I remembered that doesn't work in IE11 so pushed a solution that will work there too.

## Screenshots

Used the instructions from #555 for the content.

Frontend (dark mode)

Screenshot_2020-10-19 clearings #555 – My Blog

Editor

Screenshot_2020-10-19 Edit Post ‹ My Blog — WordPress

## Quality assurance

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -748,6 +748,10 @@ a:hover {
 	margin-bottom: 0;
 }
 
+.wp-block-group {
+	display: flow-root;
+}
+
 .wp-block-group.has-background {
 	padding: 30px;
 }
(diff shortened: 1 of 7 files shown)


carolinan (approved):
Works well for me. 

Before:
before

After:
after

Issue #592, opened 2020-10-19
Visibility issues with dark theme preference


briceduclos:
**Describe the bug**
#528 added the dark mode preference to the theme. When testing this mode, I encounter several issues (the test was made with the background color set to white inside the Customizer):
- inside a group with light gray background color (custom color #f5f5f5), the text inside is also light-gray therefore unreadable
- if the site logo is dark, it becomes nearly invisible, same for illustration like https://opendoodles.s3-us-west-1.amazonaws.com/dancing.png
- the social icons inside the footer are still dark
- inside the editor, the dark theme isn’t applied, so there’s a difference between the front and the back.

IMHO, even if it was working perfectly, the user should have a Customizer option to turn this off as it can really change the website design without him knowing.

**Screenshots**
| "Normal" mode  | Dark theme |
| ------------- | ------------- |
| ![normal-mode](https://user-images.githubusercontent.com/56188815/96458765-9bada000-1221-11eb-9b44-90a769d8e241.PNG) ![normal-mode-footer](https://user-images.githubusercontent.com/56188815/96459137-0c54bc80-1222-11eb-8d1b-9a6baa933dcb.PNG)  | ![dark-mode](https://user-images.githubusercontent.com/56188815/96458767-9c463680-1221-11eb-93fa-1ef81fbd95b0.PNG) ![dark-mode-footer](https://user-images.githubusercontent.com/56188815/96459141-0ced5300-1222-11eb-8fdc-9fed3ce15287.PNG) |

carolinan:
Please split this into separate issues, the social icon color problem is not only present when the user has a preference set.
I am going to assume that this is the menu, not the social icon block?

carolinan:
_if the site logo is dark, it becomes nearly invisible,_  
That is not up to the theme, but I agree there needs to be some information on the themes support page that the theme has support for _prefers-color-scheme_

ryelle:
> inside a group with light gray background color (custom color #f5f5f5), the text inside is also light-gray therefore unreadable

This is also a problem when a dark background color is selected by the user, the only difference is that it's obvious in the editor since the dark background is used here too. It's a custom color, we have no way of knowing whether it's light or dark to programmatically change the text color.

> inside the editor, the dark theme isn’t applied, so there’s a difference between the front and the back.

If the user has the dark-mode setting on, the editor should show the dark scheme, so this should be updated. (unless this was an intentional decision? maybe this needs core/gutenberg support)

[...]

PR #597, opened 2020-10-19
Support dark mode in the editor

jffng:
This PR tries an approach for supporting dark mode in the editor, addressing part of #592.

**To test**
- Check out this PR and enable dark mode on your system
- Update or verify your site is using the default background color 
- Visit the editor and verify the dark styles match the front-end

### DIFF
--- a/classes/class-twenty-twenty-one-custom-colors.php
+++ b/classes/class-twenty-twenty-one-custom-colors.php
@@ -47,6 +47,21 @@ class Twenty_Twenty_One_Custom_Colors {
 	}
 
 	/**
+	 * Determine if the background color is of the theme's default palette.
+	 *
+	 * @access public
+	 *
+	 * @since 1.0.0
+	 *
+	 * @return bool
+	 */
+	public function is_default_palette() {
+		$background_color             = get_theme_mod( 'background_color', 'D1E4DD' );
+		$light_colors_default_palette = array( '#D1E4DD', '#D1DFE4', '#D1D1E4', '#E4D1D1', '#E4DAD1', '#EEEADD', '#FFFFFF' );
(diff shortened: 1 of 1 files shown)


classes/class-twenty-twenty-one-custom-colors.php
```diff
@@ -115,8 +130,12 @@ public function editor_custom_color_variables() {
 			array(),
 			(string) filemtime( get_theme_file_path( 'assets/css/custom-color-overrides.css' ) )
 		);
+
 		if ( 'd1e4dd' !== strtolower( get_theme_mod( 'background_color', 'D1E4DD' ) ) ) {
 			wp_add_inline_style( 'twenty-twenty-one-custom-color-overrides', $this->generate_custom_color_variables( 'editor' ) );
+		} elseif ( $this->is_default_palette() ) {
+			// Add dark mode variable overrides.
+			wp_add_inline_style( 'twenty-twenty-one-custom-color-overrides', '@media (prefers-color-scheme: dark) { :root .editor-styles-wrapper { --global--color-background: var(--global--color-dark-gray); --global--color-primary: var(--global--color-light-gray); --global--color-secondary: var(--global--color-light-gray); } }' );
 		}
```
ryelle: The first conditional is catching when a different light background is chosen (ex, #e4dad1), so the editor shows the light background colors instead.

I think you want two separate conditionals to merge the two CSS outputs, so that the custom color overrides can be loaded in with the dark-mode  media query.

jffng: Fixed, thank you!

ryelle (approved):
Looks good in my testing 👍

PR #599, opened 2020-10-19
Colors: Add an option to disable the "dark mode" functionality

ryelle:
See #592 

## Summary

The "prefers-color-scheme"/dark mode feature might be unexpected for some site content & customizations, so this PR adds a way to disable it. This also adds a place for us to explain that this feature exists, and how to use it.

## Screenshots

Screen Shot 2020-10-19 at 3 14 56 PM

This text is very, very placeholder — would love some help with the longer description, and what, if any link we should provide for more info. Ideally something that would explain how to turn on dark mode yourself to test it out.

Text out of the screenshot:
> **Support visitor's color scheme preference**
> Show your site in dark mode if a visitor to your site requests it. More info? Accessibility warning?

## Test instructions

1. Open the customizer -> Colors
2. Make sure the background is set to a light color
3. Turn on dark mode, the background should flip to a dark color
4. Unselect the new checkbox, save
5. The background should flip back to the light color
6. Select a dark background, it should save & display correctly

## Quality assurance
* [x] I have checked that this code does not affect the accessibility negatively — I think we're good since it's on by default

### DIFF
--- a/classes/class-twenty-twenty-one-custom-colors.php
+++ b/classes/class-twenty-twenty-one-custom-colors.php
@@ -65,7 +65,7 @@ class Twenty_Twenty_One_Custom_Colors {
 
 		$theme_css        = 'editor' === $context ? ':root .editor-styles-wrapper{' : ':root{';
 		$background_color = get_theme_mod( 'background_color', 'D1E4DD' );
-		
+
 		if ( 'd1e4dd' !== strtolower( $background_color ) ) {
 			$theme_css .= '--global--color-background: #' . $background_color . ';';
 			$theme_css .= '--global--color-primary: ' . $this->custom_get_readable_color( $background_color ) . ';';
@@ -170,8 +170,9 @@ class Twenty_Twenty_One_Custom_Colors {
 			$classes[] = 'is-background-light';
 		}
 
+		$should_respect_color_scheme  = get_theme_mod( 'respect_user_color_preference', true );
 		$light_colors_default_palette = array( '#D1E4DD', '#D1DFE4', '#D1D1E4', '#E4D1D1', '#E4DAD1', '#EEEADD', '#FFFFFF' );
-		if ( in_array( strtoupper( '#' . ltrim( $background_color, '#' ) ), $light_colors_default_palette, true ) ) {
+		if ( $should_respect_color_scheme && in_array( strtoupper( '#' . ltrim( $background_color, '#' ) ), $light_colors_default_palette, true ) ) {
 			$classes[] = 'has-default-light-palette-background';
 		}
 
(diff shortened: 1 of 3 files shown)


classes/class-twenty-twenty-one-custom-colors.php
```diff
@@ -170,8 +170,9 @@ public function body_class( $classes ) {
 			$classes[] = 'is-background-light';
 		}
 
+		$should_respect_color_scheme  = get_theme_mod( 'respect_user_color_preference', true );
```
ryelle, line 173: Looks like this is flagging the phpstan check, but we do the same thing with `get_theme_mod( 'display_title_and_tagline', true )` — is this safe to `@phpstan-ignore-line`?

classes/class-twenty-twenty-one-customize.php
```diff
@@ -146,6 +146,27 @@ public function register( $wp_customize ) {
...
+					},
+				)
+			);
+
+			$wp_customize->add_control(
+				'respect_user_color_preference',
+				array(
+					'type'        => 'checkbox',
+					'section'     => 'colors',
+					'label'       => esc_html__( 'Support visitor\'s color scheme preference', 'twentytwentyone' ),
```
jffng, line 166: I feel if the wording were reversed, it would be more clear:

**Label**
Support Dark Mode

**Description**
By checking this box, your site will use a dark color palette if the visitor's system appearance is set to prefer dark: _link to more info about accessibility / system appearance settings_

jffng (commented):
Works in my testing!

Will need to update #597 as a follow up.

classes/class-twenty-twenty-one-custom-colors.php
aristath: Yes, it's safe to ignore :+1:

classes/class-twenty-twenty-one-customize.php
aristath: I'd like to suggest a slightly different label for this setting... `Respect visitor's device light/dark mode settings`.
* It makes it easier to understand that _"hey, users have their preferences, perhaps you should respect them"_.
* The use of "device" in there immediately makes me think of my phone, and it will be easier for someone non-technical to make the connection with the light/dark setting they have on their own device.

carolinan:
This is tricky, but we have until the soft string freeze in beta 4, november 10, to improve the description of the option.

But since it is not a bug it needs to be commited today.

classes/class-twenty-twenty-one-customize.php
carolinan: I think we at least need to keep it at this shorter length. "By checking this box" it sounds a little bit like a users agreement =D

jffng:
In the interest of time I opened #615 as I can't commit to @ryelle's branch, hope that's ok!

PR #600, opened 2020-10-19
Colors: Use the "dark editor style" when using a dark background color

ryelle:
When a user's selected a dark background for their site, the editor should use the "dark styles", so that the placeholder content displays.

| Before | After |
|--------|------|
| Screen Shot 2020-10-19 at 7 27 01 PM | Screen Shot 2020-10-19 at 7 27 16 PM |

## Test instructions

1. Go to Customizer -> Colors, select a dark background (from palette or custom)
2. See the text flips to white on the frontend
3. Go to create a new post or page
4. The placeholders should also be light colored text

Optionally test that this doesn't regress light backgrounds 👍 

## Quality assurance

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -126,6 +126,10 @@ if ( ! function_exists( 'twenty_twenty_one_setup' ) ) {
 
 		// Add support for editor styles.
 		add_theme_support( 'editor-styles' );
+		$background_color = get_theme_mod( 'background_color', 'D1E4DD' );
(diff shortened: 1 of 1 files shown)


aristath (approved):

aristath:
Note: this PR was causing a fatal error because the `get_relative_luminance_from_hex` is not static and is called statically.
Fixed in https://github.com/WordPress/twentytwentyone/commit/a1f5991d0f7941171585de22a14af4163e0bb304

PR #615, opened 2020-10-20
Colors: Add an option to disable the "dark mode" functionality (continuation of #599)

jffng:
This PR continues #599. 

---

See #592 

## Summary

The "prefers-color-scheme"/dark mode feature might be unexpected for some site content & customizations, so this PR adds a way to disable it. This also adds a place for us to explain that this feature exists, and how to use it.

## Screenshots

Screen Shot 2020-10-19 at 3 14 56 PM

This text is very, very placeholder — would love some help with the longer description, and what, if any link we should provide for more info. Ideally something that would explain how to turn on dark mode yourself to test it out.

Text out of the screenshot:
> **Support visitor's color scheme preference**
> Show your site in dark mode if a visitor to your site requests it. More info? Accessibility warning?

## Test instructions

1. Open the customizer -> Colors
2. Make sure the background is set to a light color
3. Turn on dark mode, the background should flip to a dark color
4. Unselect the new checkbox, save
5. The background should flip back to the light color
6. Select a dark background, it should save & display correctly

## Quality assurance
* [x] I have checked that this code does not affect the accessibility negatively — I think we're good since it's on by default

### DIFF
--- a/classes/class-twenty-twenty-one-customize.php
+++ b/classes/class-twenty-twenty-one-customize.php
@@ -163,8 +163,8 @@ if ( ! class_exists( 'Twenty_Twenty_One_Customize' ) ) {
 				array(
 					'type'            => 'checkbox',
 					'section'         => 'colors',
-					'label'           => esc_html__( 'Respect visitor\'s device light/dark mode settings', 'twentytwentyone' ),
-					'description'     => __( 'Show your site in dark mode if a visitor to your site requests it.', 'twentytwentyone' ),
+					'label'           => esc_html__( 'Respect visitor\'s device dark mode settings', 'twentytwentyone' ),
+					'description'     => __( 'Dark mode is a device setting. If a visitor to your site requests it, your site will be shown with a dark background and light text.', 'twentytwentyone' ),
 					'active_callback' => function( $value ) {
 						return 127 < Twenty_Twenty_One_Custom_Colors::get_relative_luminance_from_hex( get_theme_mod( 'background_color', 'D1E4DD' ) );
 					},
--- a/classes/class-twenty-twenty-one-customize.php
+++ b/classes/class-twenty-twenty-one-customize.php
@@ -161,10 +161,13 @@ if ( ! class_exists( 'Twenty_Twenty_One_Customize' ) ) {
 			$wp_customize->add_control(
 				'respect_user_color_preference',
 				array(
-					'type'        => 'checkbox',
-					'section'     => 'colors',
-					'label'       => esc_html__( 'Respect visitor\'s device light/dark mode settings', 'twentytwentyone' ),
-					'description' => __( 'Show your site in dark mode if a visitor to your site requests it.', 'twentytwentyone' ),
+					'type'            => 'checkbox',
+					'section'         => 'colors',
+					'label'           => esc_html__( 'Respect visitor\'s device light/dark mode settings', 'twentytwentyone' ),
+					'description'     => __( 'Show your site in dark mode if a visitor to your site requests it.', 'twentytwentyone' ),
+					'active_callback' => function( $value ) {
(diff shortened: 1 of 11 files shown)


aristath:
I was thinking.... If we add that option, we no longer need to check if the color is part of the default color-palette, right?
It should be available if the selected color has a luminosity > 127 - in which case the background color is light.

jffng:
This is ready for another review. 

Dark mode should be respected in the editor and front-end. If you have chosen a dark background color (luminosity < 127), the checkbox is hidden and dark mode is ignored.

aristath (approved):
Code looks good to me :+1: 
Tested and confirmed the functionality :partying_face:

functions.php
```diff
@@ -524,16 +524,52 @@ function twenty_twenty_one_non_latin_languages() {
  * @return void
  */
 function twentytwentyone_customize_preview_init() {
+	wp_enqueue_script(
+		'twentytwentyone-customize-helpers',
+		get_theme_file_uri( '/assets/js/customize-helpers.js' ),
+		array(),
+		get_theme_file_path( 'assets/js/customize-helpers.js' ),
```
ryelle: This should be a version, not a file path— looks like the other scripts use `wp_get_theme()->get( 'Version' ),`

functions.php
```diff
@@ -524,16 +524,52 @@ function twenty_twenty_one_non_latin_languages() {
...
+ *
+ * @return void
+ */
+function twentytwentyone_customize_controls_enqueue_scripts() {
+
+	wp_enqueue_script(
+		'twentytwentyone-customize-helpers',
+		get_theme_file_uri( '/assets/js/customize-helpers.js' ),
+		array(),
+		get_theme_file_path( 'assets/js/customize-helpers.js' ),
```
ryelle: Same version issue here, and in the `wp_enqueue_script` below.

classes/class-twenty-twenty-one-customize.php
```diff
@@ -146,6 +146,30 @@ public function register( $wp_customize ) {
...
+				)
+			);
+
+			$wp_customize->add_control(
+				'respect_user_color_preference',
+				array(
+					'type'            => 'checkbox',
+					'section'         => 'colors',
+					'label'           => esc_html__( 'Respect visitor\'s device light/dark mode settings', 'twentytwentyone' ),
+					'description'     => __( 'Show your site in dark mode if a visitor to your site requests it.', 'twentytwentyone' ),
```
ryelle: I think this could be more descriptive — it says basically the same thing as the label, and doesn't give the site owner any info on how to enable it themselves. Thinking on this more, but I don't want to hold up my review.

ryelle (changes requested):
Working well in my testing, just the issue with script versions needs to be updated.

functions.php
aristath: fixed

functions.php
aristath: fixed

classes/class-twenty-twenty-one-customize.php
ryelle:
```suggestion
					'label'           => esc_html__( 'Respect visitor\'s device dark mode settings', 'twentytwentyone' ),
					'description'     => __( 'Dark mode is a device setting. If a visitor to your site requests it, your site will be shown with a dark background and light text.', 'twentytwentyone' ),
```

label: Removed the "light/" — technically we don't respect a "light mode", if the user prefers light we don't change dark backgrounds.
description: Made it a little more clear what happens, and re-emphasized the device-setting nature of dark mode.

aristath: Makes sense :+1:

jffng:
I'm going to merge this because we've gotten two reviews, and there appears to be consensus that this improves the existing dark mode functionality.

To summarize the changes, we:

- Added a checkbox option in the customizer to toggle dark mode support.
- Added some context + info in the customizer for what this feature does.
- Activated dark mode by checking luminosity of the background color, rather than default color palette values.

Thanks everyone!

Issue #621, opened 2020-10-20
Dark Mode: Should we turn it off in the admin?


melchoyce:
**Describe the bug**
It can be confusing for your site to appear in dark mode while you're editing it in the admin, either in the editor, or in the Customizer. This is made extra confusing, IMO, because wp-admin itself doesn't support dark mode yet. Would it be less confusing if we didn't support dark mode within the editor and Customizer?

**To Reproduce**
1. Turn on dark mode in your OS or browser.
2. Visit your site's admin.
3. Either enter the editor or the Customizer.
4. You see the dark mode version of your site.

**Expected behavior**
When I'm editing my site, I expect to see the "canonical" version of it.

carolinan:
Even if we had user testing to learn what the majority wants, that would still be the majority.

That is why the option to turn it on and off is better.

asathoor:
Sometimes a uger might want to use WordPress in Dark Mode and your OS in another contrasting mode. That's why I'd like to have a button somewhere where it's easy to find.

carolinan:
If dark mode is off on the OS, a dark color option can still be chosen in the customizer.

hedgefield:
Perfect. I made a dark mode for my site recently, and in doing so I read a lot that it's good practice to offer a way for someone to toggle it on and off regardless of system setting. Especially if you want to author the dark theme during the day or vice versa, having a manual toggle right in WP seems ideal to me.

aristath:
During Friday's meeting it was decided to remove everything related to dark-mode and instead put that feature in a plugin as a safety precaution. dark-mode was removed in https://github.com/WordPress/twentytwentyone/pull/690 and moved in a plugin, with the hope that it can be re-integrated in the theme if we're satisfied with its implementation.
Closing this issue.

PR #622, opened 2020-10-21
Add a dark-mode toggle

aristath:
Came up in various tickets that we need a toggle for dark/light mode. Some users use dark mode on their OS but prefer light for webpages. Others prefer dark mode due to a11y. Whatever the reason, having a toggle makes things easier for end-users.

This PR is an example of how we can do it. It's pretty simple and effective... 
The one I went with here is MIT-licenced from https://codepen.io/waynedunkley/pen/YybgGo but there are many to choose from, or we can create one specifically for this theme (would be preferable).

Peek 2020-10-21 07-58

The switch used in this PR feels a bit out of context and doesn't match the overall design, @melchoyce could we create something for toggling between dark mode and light mode?

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -6782,46 +6782,12 @@ footer {
 #night-day-toggle {
 	cursor: pointer;
 	position: fixed;
-	bottom: 1em;
-	right: 1em;
-	padding: 24px;
-}
-
-#night-day-toggle label {
-	border-radius: 50%;
-	width: 36px;
-	height: 36px;
-	position: absolute;
-	top: 5px;
-	left: 5px;
-	box-shadow: inset 16px -16px 0 0 #28303d;
-	transform: scale(1) rotate(-2deg);
-	transition: box-shadow .5s ease 0s, transform .4s ease .1s;
-}
-
-#night-day-toggle label:before {
-	content: "";
-	width: inherit;
-	height: inherit;
-	border-radius: inherit;
-	position: absolute;
-	left: 0;
-	top: 0;
-	transition: background .3s ease;
-}
-
-#night-day-toggle label:after {
-	content: "";
-	width: 8px;
-	height: 8px;
-	border-radius: 50%;
-	margin: -4px 0 0 -4px;
-	position: absolute;
-	top: 50%;
-	left: 50%;
-	box-shadow: 0 -23px 0 #fff, 0 23px 0 #fff, 23px 0 0 #fff, -23px 0 0 #fff, 15px 15px 0 #fff, -15px 15px 0 #fff, 15px -15px 0 #fff, -15px -15px 0 #fff;
-	transform: scale(0);
-	transition: all .3s ease;
+	bottom: 5px;
+	right: 5px;
+	padding: 5px;
+	display: flex;
+	align-items: center;
+	justify-content: center;
 }
 
 #night-day-toggle input {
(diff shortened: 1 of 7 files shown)


aristath:
Update: This now matches the overall theme styles a lot better, but can further be improved & customized.

Peek 2020-10-21 09-18

hedgefield:
I love this! Beautiful icon choice, and as I mentioned in https://github.com/WordPress/twentytwentyone/issues/621 it's great that we're offering this option.

beafialho:
Let me jump in the conversation and show a very rough prototype I did of the toggle design. The way I see it is having little to no ornamentation. The moon shape may not even be needed since there's a color shift.

darkmode-toggle

Captura de ecrã 2020-10-21, às 16 41 05
Captura de ecrã 2020-10-21, às 16 41 14

melchoyce:
Thanks for the mockups, @beafialho!

What would it look like if we added a 1px opposite-color border to the toggle? Worried about it being invisible if it happens to be on top of a custom color that's a similar shade.

ryelle:
A few accessibility notes
- the bottom-right floating icon makes it hard to know where in the page's tab order it is
- this is in the footer of the site, so you need to tab through the whole page to get to it
- it doesn't seem to be focusable, but i'm not 100% sure on that, it will need a focus style
- it's being `aria-hidden` right now, but a screen reader user might want to be able to control this (sighted people use screen readers too)

Moving the toggle to the top of the page would help with the first two issues, but I'm not sure if there would be other design considerations.

melchoyce:
re: positioning, what about...

| Desktop | Tablet | Phablet | Small Phone |
| ----- | ----- | ----- | ----- |
| image | image | image |  image |

(The toggle is always 44px tall to comply with AAA sizing guidelines)

melchoyce:
One a11y question: Will the toggle need to have a label?

aristath:
* Made it focusable
* Added focus-styles
* Removed `aria-hidden`
* Added a label (screen-reader-only)

hedgefield:
To me, a standard toggle is less clear than the sun/moon, I feel like that's an established pattern for dark mode now. I wouldn't know what a regular toggle does presented in that spot without context, but adding a label would make it too prominent maybe. I do like the subtle moon shape that the toggle knob has in dark mode, maybe we should at least reverse that? So that's visible in light mode, indicating it will transition to dark mode as the toggle action?

aristath:
I agree that a standard toggle is a bit weird... But nevertheless my last commit implements it since from what I understood @melchoyce seemed to like the idea. It currently looks like this:

Peek 2020-10-22 13-09

I'd very much like to see something like the ones shown in these examples (posting for inspiration):

* https://codepen.io/bheberer/pen/BaNZKmq
* https://codepen.io/lucas-labs/pen/wvBvEZq
* https://codepen.io/sandeshsapkota/pen/xxVmMpe
* https://codepen.io/z-/pen/BaBxgro
* https://codepen.io/avstorm/pen/XWrNNOE
* https://codepen.io/sheelah/pen/dEmyqx

melchoyce:
The toggle is still referencing sun/moon, but let me check out those CodePen examples and see if one would work okay for the theme.

aristath:
Closing this one. It was ported to the (temporary) plugin repo in https://github.com/aristath/tt1-dark-mode/commit/45531aff8bfd0aebb1f573c5f743f02faa00339d

PR #624, opened 2020-10-21
pre tag is not preserving whitespace

### LINKED ISSUE #553: pre tag is not preserving whitespace
The `pre` tag is supposed to preserve whitespace, but it is using 
`pre {white-space: pre-wrap}`
which does not.
The font size should probably be reduced a bit, since it's monospace.

**Screenshots**
It looks like this mess:
pre-tag

If the `pre {white-space: pre-wrap}` rule is disabled, it looks better, but has a scrollbar. So if you use 
`pre {overflow-x: auto}`
it looks like this:
pre-tag-overflow

SResok:
Fixes #553

## Summary

New 
- Added a `preformatted` block
- editor preformatted block font size now matches with the front

Changes
- the `code` block its color to `currentColor` in the editor
- misc.scss: pre tag inside a `html` block will preserve whitespace too

## Test instructions

1. Add a `preformatted` and `code` block to a post/page
1. Add some content to those blocks to confirm it's preserving whitespace

## Quality assurance

### DIFF
--- a/assets/sass/04-elements/misc.scss
+++ b/assets/sass/04-elements/misc.scss
@@ -12,5 +12,6 @@ i {
 }
 
 pre {
-	white-space: pre-wrap;
+	white-space: pre;
+	overflow-x: auto;
 }
--- a/assets/sass/05-blocks/blocks-editor.scss
+++ b/assets/sass/05-blocks/blocks-editor.scss
@@ -20,6 +20,7 @@
 @import "media-text/editor";
 @import "navigation/editor";
 @import "paragraph/editor";
+@import "preformatted/editor";
 @import "pullquote/editor";
 @import "quote/editor";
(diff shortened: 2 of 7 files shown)


carolinan:
The font size needs to be applied to the front too, otherwise it looks good to me.
(On the front' it falls back to 1em set by normalize.)

carolinan:
> The font size needs to be applied to the front too, otherwise it looks good to me.
> (On the front' it falls back to 1em set by normalize.)

So that turned out not to be so easy because the browsers handles it differently. Lets keep it the way it is in this PR.

In my opinion this is ready to be merged, unless @ryelle  found improvements that can be made.

ryelle (approved):
Looks good to me too 👍

PR #625, opened 2020-10-21
Update the background color of input fields

### LINKED ISSUE #623: Placeholders in input fields are not readable when a dark background is used.
**Describe the bug**

Regular input fields has a background color that is the same as the body background color, but with lower opacity.

When a dark body background color is used, the _placeholder_ text inside form input fields does not have a high enough contrast against the background.

**To Reproduce**
1. Change the background color
2. Add an input field with a placeholder. _-In my case, I was looking at a coupon field of a WooCommerce install._
But I was able to reproduce the same, by adding a search block with a placeholder in the editor.
3. See error

**Expected behavior**
The text needs to be readable.

**Screenshots**

Search block placeholder in the editor and front:
search

Example placeholder with WooCommerce installed:
placeholder

- WordPress version: (5.6-beta1-49262)
- If the Gutenberg plugin is installed, which version is it? 9.2
(WooCommerce 4.6.0)

 - OS: Windows 10
 - Browser Chrome Version 86.0.4240.75

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/623

## Summary

Regular input fields has a background color that is the same as the body background color, but with lower opacity.

This PR changes the background color of an input that has a placeholder text to always be white, no matter what the global body background color is.

**It has the side effect that the input fields will look different from other input fields, that does have a background color.**

By changing the input background color, the result is that the placeholder text color is always readable, with very little effort.

Another alternative would be to calculate a text color that has a high enough contrast compared to the fields background color. But changing the text color would mean that the inputs text color would/may be inconsistent depending on if a field is actively being edited or not.

## Test instructions

1. Change the body background color to dark
1. Add a search block with a placeholder, view its colors.
1. View the search block on the front.
1. Test the input by focusing on it (tabbing to it), and by editing its contents.

## Screenshots

If this is a visual change please include screenshots of the change at various screen sizes.

Before and after:
search placeholder before and after

## Quality assurance

### DIFF
--- a/assets/sass/04-elements/forms.scss
+++ b/assets/sass/04-elements/forms.scss
@@ -50,6 +50,10 @@ input[type="color"] {
 	height: calc(4 * var(--form--spacing-unit));
 }
 
+input:placeholder-shown {
+	background: var(--global--color-white);
+}
+
 select {
 	border: var(--form--border-width) solid var(--form--border-color);
 	color: var(--form--color-text);
--- a/assets/sass/04-elements/forms.scss
+++ b/assets/sass/04-elements/forms.scss
@@ -299,3 +299,7 @@ fieldset {
 		}
 	}
 }
+
+::-moz-placeholder { // Firefox 19+
+	opacity: 1;
+}
--- a/assets/sass/04-elements/forms.scss
+++ b/assets/sass/04-elements/forms.scss
@@ -16,18 +16,23 @@ input[type="color"],
 	border: var(--form--border-width) solid var(--form--border-color);
 	border-radius: var(--form--border-radius);
 	color: var(--form--color-text);
-	background: var(--global--color-white-50);
 	line-height: var(--global--line-height-body);
 	padding: var(--form--spacing-unit);
+	// Gives a little more space for the outline offset.
+	margin: 0 2px;
 
 	&:focus {
 		color: var(--form--color-text);
-		outline: 1px solid var(--form--border-color);
-		background: var(--global--color-white);
+		outline-offset: 2px;
+		outline: 2px dotted var(--form--border-color);
 	}
 
 	&:disabled {
 		opacity: 0.7;
+	}
+
+	.is-background-dark & {
+		background: var(--global--color-white-90);
 	}
 
 }
@@ -36,22 +41,17 @@ input[type="color"],
 input[type="search"] {
 
 	&:focus {
-		outline-offset: 0;
+		outline-offset: -7px;
+
+		.is-background-dark & {
+			outline-color: var(--global--color-background);
+		}
 	}
 }
 
-input[type="date"] {
-	background: var(--global--color-white);
-}
-
 input[type="color"] {
-	background: var(--global--color-white);
 	padding: calc(var(--form--spacing-unit) / 2);
 	height: calc(4 * var(--form--spacing-unit));
-}
-
-input:placeholder-shown {
-	background: var(--global--color-white);
 }
 
 input[type="email"],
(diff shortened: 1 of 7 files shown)


ryelle:
This also affects the light mode inputs:

Screen Shot 2020-10-21 at 12 29 17 PM

I think this needs some Accessibility feedback too, since the focus state on these inputs is not very perceptible anymore.

assets/sass/04-elements/forms.scss
```diff
@@ -50,6 +50,10 @@ input[type="color"] {
 	height: calc(4 * var(--form--spacing-unit));
 }
 
+input:placeholder-shown {
```
ryelle: In Firefox this was affecting inputs with empty placeholders too — this should work in a more cross-browser friendly way:
```suggestion
input[placeholder]:not([placeholder=""]) {
```

carolinan:
Yeah the accessibility impacts here collide, the opacity makes it tricky.
The readability and color contrast is also a problem when a field is edited, and then focus is moved away from that field.

carolinan:
@melchoyce I don't want to ask this because they are a feature of the theme, but I think the best solution is that the inputs always have the browser default white background, (or a much lighter but consistent off white) with dark text.

On the plus side:
The fields are consistent and easy to recognize. 
This is an easy change to make

On the down side:
Less color variation in the theme. Even if this variation is subtle.
A new focus style needs to be added since the growing white border outside the white background is almost invisible.

melchoyce:
If you think that's going to be the more straightforward approach, I'm all for it 👍 

Instead of growing the border on focus, maybe we could take a similar approach to buttons:

**Idea 1: Focus on inside**
image

**Idea 2: Focus on outside**
image

There isn't enough room for radios and checkboxes to have inside focuses, so they'd have to have outer focuses:
image

Is it possible to add the focus to both the checkbox/radio and label combined?
image

carolinan:
I will try to see if there is a simple way to soften the white color when the dark palette is selected, since that is something that has been discussed for the other white elements.

carolinan:
Light colors:
form-focus

Dark colors:
dark-search

Search results, prefilled input:
searchresult

Placeholder:
placeholder

form-focus-dark

carolinan:
Here, I realized that the opacity change that I had put on the input fields are affecting the buttons, so I made some more changes to avoid that.

afercia:
Worth reminding that the Firefox user agent stylesheet applies some opacity to the placeholder pseudo element so whatever the color of the background/placeholder is, the placeholder text will look lighter in Firefox (and likely not meet a sufficient contrast ratio). To reset the opacity set by Firefox:

```
::-moz-placeholder {
	opacity: 1;
}
```

carolinan:
Firefox placeholder text before and after:

firefox-placeholder-opacity

assets/sass/04-elements/forms.scss
```diff
@@ -16,36 +16,40 @@ input[type="color"],
...
-		background: var(--global--color-white);
+		outline-offset: 2px;
+		outline: 2px dotted var(--form--border-color);
 	}
 
 	&:disabled {
 		opacity: 0.7;
 	}
 
+	.is-background-dark & {
```
carolinan: Note to self this class has been updated to has-background-dark in trunk.

kjellr:
It seems a little unfortunate to me that the field border is hard to discern from the field background in the latest version of this: 



Since the default version of the field background is derived from white for all light-colored backgrounds, have we considered just swapping that out for black for dark-colored backgrounds? Here's a quick mockup: 

Light Backgrounds|Dark Backgrounds
---|---
Frame 65|Frame 66

carolinan:
I think that is a bad idea and it was avoided on purpose because inputs are recognized as being white with dark text. 
Secondly because we would need to change the text color of multiple elements, which we do not need to do currently.

carolinan:
I do not feel that redoing this a 4th time would be time and effort well spent. Every thing we do during this period needs to be weighed against testing and all other issues.

melchoyce:
LGTM 👍

aristath (approved):

carolinan:
Did a quick test with 5.6-beta1-49314 and it still works.

PR #630, opened 2020-10-21
rename searchform label to aria_label argument.

### LINKED ISSUE #628: Correct the parameter used with the search form
**Describe the bug**

In searchform.php, the parameter/argument that is used is named `$args['label']`, the correct name for use with 
get_search_form() is `$args['aria_label']` so this needs to be updated.
See https://developer.wordpress.org/reference/functions/get_search_form/#parameters

<del>The theme calls get_search_form() in 3 places, but the aria_label arg is not used.</del>

**To Reproduce**
1. View the markup of a search form, for example the search widget or the 404 page.
2. See that no ARIA label is used.

**Expected behavior**
Each form should have a relevant ARIA label, since it is possible to have more than one search form on the same page.

aristath:
Fixes #628

Discussed with @carolinan and @afercia and came to the conclusion that the searchforms don't need a specific aria-label. Using an aria-label to differentiate searchforms would only make sense if they were performing different search queries (example: one for posts, one for products via a plugin). Since all existing forms in the theme perform the default search, making them different is not necessary.
This PR swill simply fix the argument name from `label` to `aria_label`.

## Quality assurance

### DIFF
--- a/searchform.php
+++ b/searchform.php
@@ -18,7 +18,7 @@
  */
 $twentytwentyone_unique_id = wp_unique_id( 'search-form-' );
 
-$twentytwentyone_aria_label = ! empty( $args['label'] ) ? 'aria-label="' . esc_attr( $args['label'] ) . '"' : '';
+$twentytwentyone_aria_label = ! empty( $args['aria_label'] ) ? 'aria-label="' . esc_attr( $args['aria_label'] ) . '"' : '';
 ?>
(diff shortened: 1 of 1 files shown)


ryelle:
This is to standardize the theme's behavior with the parameters documented in core for `get_search_form`, is that right?

aristath:
> This is to standardize the theme's behavior with the parameters documented in core for get_search_form, is that right?

Correct :+1:

ryelle (approved):

PR #634, opened 2020-10-21
Enable dark-mode in the editor when needed

aristath:
When a light color is selected as background AND the user has chosen to respect the visitors color-scheme AND a post-author is using dark-mode in their OS, the color-scheme in the editor is dark but the editor itself is not. That can lead to some things being unreadable.
Example: The post-title placeholder.
Screenshot_2020-10-21 Edit Post ‹ My Blog — WordPress(4)

This PR will check if we're in the editor, if the `respect_user_color_preference` option is enabled, the default background is light, and if all of the above is true then it adds a `twentytwentyone-supports-dark-theme` to `<body>` on the admin.
Then, via JS we check if the class exists in the body and if the client's browser has dark-mode enabled. If yes, then we add the `is-dark-theme` class to `<body>` which changes the editor UI to dark mode.

Screenshot_2020-10-21 Edit Post ‹ My Blog — WordPress(3)

## Quality assurance

### DIFF
--- a/assets/js/editor-dark-mode-support.js
+++ b/assets/js/editor-dark-mode-support.js
@@ -1,0 +1,6 @@
+if (
+	document.body.classList.contains( 'twentytwentyone-supports-dark-theme' ) &&
+	window.matchMedia( '(prefers-color-scheme: dark)' ).matches
+) {
+	document.body.classList.add( 'is-dark-theme' );
+}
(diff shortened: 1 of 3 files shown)


assets/js/editor-dark-mode-support.js
```diff
@@ -0,0 +1,6 @@
+if (
```
ryelle, line 1: It's possible this could change once the user's loaded the page (Macs have an "auto" setting which will switch to dark in the evening). I played around with `matchMedia().addListener()` locally & got it working, mind if I push that to this branch?

aristath: Of course, feel free to push :heart:

aristath:
Closing this one. Ported to the (temporary) plugin repo in https://github.com/aristath/tt1-dark-mode/commit/716bb36d9b42560d0fa16863a0c9503d653ec95d

PR #635, opened 2020-10-21
Fix aria-expanded on nav hover

### LINKED ISSUE #633: Update the menu toggle button aria-expanded value also on hover
**Describe the bug**
In the navigation menu, the buttons to toggle the sub-menus use an aria-expanded attribute to indicate the sub-menu collapsed/expanded state. The aria-expanded value is updated on click. So far so good. However, the sub-menus open also on mouse hover. In this case, the aria-expanded value is _not_ updated.

The aria-expanded value should always reflect the actual state of the sub-menu regardless of the action taken to open it. 

A few use cases:
- sometimes blind screen reader users do use the mouse to read what is under the mouse cursor
- in any case, they might have the mouse cursor that is coincidentally placed hover a toggle button
- not all screen reader users are blind: some low vision users do use screen readers as an aid but they still may use the mouse or a combination of mouse and keyboard

In all these cases, when hovering the toggle button, a screen reader would report the sub-menu state as "collapsed", which isn't correct.

**To Reproduce**
- inspect a menu toggle button in your browser's dev tools inspector
- hover the toggle button with a mouse
- the sub-menu opens
- see in the dev tools the button aria-expanded value is "false"

**Expected behavior**
The aria-expanded value should update also on hover.

aristath:
Fixes #633 

## Summary
Changes the `aria-expanded` attribute in nav dropdown-buttons when the menu-item is hovered.
Since the attribute now changes correctly I removed the specific styles for the `.hover` class, they were no longer necessary.

## Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -5769,7 +5769,7 @@ h1.page-title {
 	.primary-navigation #toggle-menu {
 		display: none;
 	}
-	.primary-navigation > .primary-menu-container ul > li:not(.hover) .sub-menu-toggle[aria-expanded="false"] ~ ul {
+	.primary-navigation > .primary-menu-container ul > li .sub-menu-toggle[aria-expanded="false"] ~ ul {
 		display: none;
 	}
 	.admin-bar .primary-navigation {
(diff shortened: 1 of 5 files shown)


ryelle (approved):
Works in testing — checked mobile & desktop sizes, and tested with VoiceOver.

PR #637, opened 2020-10-21
TRY: Button colors

### LINKED ISSUE #612: Button Block on hover: contrast ratio issue when setting custom colors
**Describe the bug**

While the hover effect (fill to outline and vice versa) works for the default buttons, as soon as the user tries to customize the colors of the button block, the hover effect makes it difficult to have a button with a good contrast ratio.

### Fill Button Examples

- Button 1: custom background-color, default text color. On hover, the text becomes black, contrast ratio too low.
- Button 2: custom background-color, text color white (from the color palette). On hover, the text still becomes black, contrast ratio too low.
- Button 3: custom background-color, custom text color. The hover state is the same as the “default” state.

"Default" state
button
Hover state
button-hover

### Outline Buttons Examples

- Button 1: text color gray (from the color palette): works inside the editor, doesn’t work on the front. All the colors from the palette don’t work on the front, they are overriden.
- Button 2: custom text color. On hover, the background becomes black, contrast ratio too low.

Default state
button-outline
Hover state
button-outline-hover

A solution might be to keep the effect for the default buttons, but not apply the hover effect when the user specifies a text color or a background color.

**To reproduce**

Here's the code:

```
<!-- wp:buttons -->
<div class="wp-block-buttons"><!-- wp:button {"style":{"color":{"background":"#007c89"}},"className":"is-style-fill"} -->
<div class="wp-block-button is-style-fill"><a class="wp-block-button__link has-background" href="#" style="background-color:#007c89">Button 1</a></div>
<!-- /wp:button -->

<!-- wp:button {"style":{"color":{"background":"#0070c9"}},"textColor":"white","className":"is-style-fill"} -->
<div class="wp-block-button is-style-fill"><a class="wp-block-button__link has-white-color has-text-color has-background" href="#" style="background-color:#0070c9">Button 2</a></div>
<!-- /wp:button -->

<!-- wp:button {"style":{"color":{"text":"#eeeeee","background":"#0070c9"}},"className":"is-style-fill"} -->
<div class="wp-block-button is-style-fill"><a class="wp-block-button__link has-text-color has-background" style="background-color:#0070c9;color:#eeeeee">Button 3</a></div>
<!-- /wp:button --></div>
<!-- /wp:buttons -->

<!-- wp:buttons -->
<div class="wp-block-buttons"><!-- wp:button {"textColor":"gray","className":"is-style-outline"} -->
<div class="wp-block-button is-style-outline"><a class="wp-block-button__link has-gray-color has-text-color" href="#">Button 1</a></div>
<!-- /wp:button -->

<!-- wp:button {"style":{"color":{"text":"#0070c9"}},"className":"is-style-outline"} -->
<div class="wp-block-button is-style-outline"><a class="wp-block-button__link has-text-color" href="#" style="color:#0070c9">Button 2</a></div>
<!-- /wp:button --></div>
<!-- /wp:buttons -->
```

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/612

## Summary

I will be the first to say that this PR is not great. But if nothing else, it highlights the issues one might run into when trying to solve the issue with the custom button colors.

**End result achieved:**
Buttons with a custom text color or custom background color keeps their colors. 
The hover style is now the same as the focus style for the outlined button: The 7px inset dotted outline.

The colors will not always be accessible. But I did not want to leave the buttons without _any_ hover style.

Buttons with border radius was not tested because I ran out of time(!).

To respect the button blocks custom text and background color on hover, active, and focus states,
I needed to move the hover and active state out of the _extends_.
That meant that I had to add these states back for two other button types: File download link, and search.

A default button is _filled_ but it does not actually use the "is-filled" class.
So to target both the default and the filled styles, I used `:not(.is-style-outline)`.

When built, the end result was unreadable, so I removed styles that were duplicated and that,
after discussion, were believed to not be needed:

`.has-focus`
` &.wp-block-button__link` (originally this was  `&.wp-block-button__link,  .wp-block-button__link {`)

I also removed focus styles from the blocks editor CSS, to not interfere with the editing.

## Test instructions

1. Add one of each button: filled, outlined, with and without custom text color and background color.
1. Remember to add links to the buttons to be able to test focus styles.
1. Test the hover state in the editor
1. Test the hover, active and focus state on the front.
1. Test that other buttons are not negatively affected.

## Screenshots

buttons

## Quality assurance


aristath:
After a lot of testing this seems to get the job done nicely.
I'm not a fan of all the `:not` selectors, but I don't see a better way to handle things. An alternative would be to remove the `:not` selectors and then add lots of overrides, but that would defeat the purpose... Instead of cleaning things up we'd end up with something a lot dirtier.

assets/sass/05-blocks/button/_editor.scss
```diff
@@ -1,36 +1,92 @@
...
-		border: var(--button--border-width) solid currentColor;
+/**
+ * Block Options
+ */
+.wp-block-button {
 
-		&:visited {
-			color: var(--button--color-background);
-		}
+	&:not(.is-style-outline) {
```
jffng, line 25:
```suggestion
	// Target the default and filled button states. 
	&:not(.is-style-outline) {
```

assets/sass/05-blocks/button/_editor.scss
```diff
@@ -1,36 +1,92 @@
...
+
+		.wp-block-button__link {
+
+			&:not(.has-background):not(.has-text-color) {
```
jffng, line 30:
```suggestion
			// Apply default button styles when custom text and background colors are not selected.
			&:not(.has-background):not(.has-text-color) {
```

jffng (commented):
In general, this looks good to me and I agree with @aristath — there isn't a clearly better way to handle this while implementing the theme design. I think your use of the not selectors is judicious and clever, even if the resulting CSS looks like garbage. 

Still needs a design review on the decision to apply the same styles as the focus states to the hover for buttons with customizations. 

Here's a screen cap with some buttons with a border radius set:

Kapture 2020-10-22 at 10 30 51

assets/sass/05-blocks/button/_editor.scss
jffng: There might be a clearer way to word this. But in general I think it would be useful to add some comments about what the intention is.

ryelle:
It looks like the palette text colors are being overridden on Outline buttons, for example I set the left button here to be a white color, which shows correctly in the editor, but black on the frontend — and then any palette color when hovered goes invisible.

Editor
Screen Shot 2020-10-22 at 10 35 34 AM

Frontend
Screen Shot 2020-10-22 at 10 35 28 AM

Frontend, both `:hover`
Screen Shot 2020-10-22 at 10 36 24 AM

Is it also expected that the default filled button looks different that filled buttons with a background color? this one doesn't have the dotted outline & the background goes transparent.

carolinan:
> Is it also expected that the default filled button looks different that filled buttons with a background color? this one doesn't have the dotted outline & the background goes transparent.

No I think the default and filled is meant the be one style.

carolinan:
Test content: https://gist.github.com/carolinan/3068a2ec3408dc51feda386ccb4e381f

carolinan:
The button block, submit and search buttons work well on the default background color, but when the background color is dark, the text is invisible when you hover over/activate the button.

The border is never visible, but that is less of a problem.

scruffian:
In Seedlet we have changed from using  `@extend %button-style` to a mixin which has several benefits:
- the amount of code it generates is less
- overriding styles is easier

melchoyce:
Does this still need any design input, or are we good to go?

carolinan:
> Does this still need any design input, or are we good to go?

It needs testing, because every time someone looks at it, new issues are found ;) 

The design question was for the hover style for the buttons with custom colors, which was changed compared to the original and figma.

melchoyce:
This is another place where Gutenberg makes it harder on theme authors to handle interactions. The focus styles (dotted outline) on hover feels weird in action, but I'm not sure how else to approach this. Could we use the same hover colors as the default buttons, regardless of what custom colors are chosen?

image

carolinan:
I think that was the original issues raised, that the buttons did not respect the users colors. 🤔

aristath:
Merged in #706

PR #640, opened 2020-10-22
Fix RTL issue

### LINKED ISSUE #638: Incorrect placement of 'Edit' link in RTL locales
**Describe the bug**

On RTL locales, such as Hebrew or Arabic, the untranslated 'Edit' text in 'edit-link' span gets concatenated with the year since there is no space between them and both span elements are inline elements.

**To Reproduce**
1. Go to Settings->General.
2. Switch the site language to an RTL locale (he_IL-עִבְרִית or ar-العربية).
3. Go to the front page of the website.
4. Observe the date and edit link at the bottom of the default post.

**Expected behavior**

The Edit link would be separated from the date.

**Screenshots**

image

**Possible solutions**
1. Make sure that the Edit text is translated for all RTL locales so that it won't be appended to the previous LTR directional sequence.
2. Add a space either before 'Edit' or after the date
3. Switch to using `display:inline-block`

aristath:
fixes #638

The problem was that the edit link was going between the date parts. Fixed by adding `display:inline-block`.

Before:

Screenshot_2020-10-22 My Blog(1)

After:

Screenshot_2020-10-22 My Blog

## Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -5022,6 +5022,10 @@ h1.entry-title {
 	display: block;
 }
 
+.entry-footer > span {
+	display: inline-block;
+}
+
 .entry-footer a {
 	color: currentColor;
 }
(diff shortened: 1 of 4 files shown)


carolinan (approved):
Looks good to me.

PR #648, opened 2020-10-22
IE: Visible sub menu items

### LINKED ISSUE #636: Internet Explorer 11: navigation sub-menus not visible
**Describe the bug**
With Internet Explorer 11, the navigation menu sub-menus aren't visible.  

**To Reproduce**
- set as primary menu a menu that does have sub-menus
- use Internet Explorer 11
- on the front end, either hover a top level menu item with sub-menus or click on it
- the sub-menu doesn't get visible
- in some cases, the following top-level items disappear (layout issue)

Animated GIF to better illustrate:

ie menu

While layout quirks are allowed to happen in IE, I guess the theme should ensure the navigation menu is still usable.

Not sure if there are specific reasons for the following `overflow` CSS properties, but seems they're the culprit and maybe they should be removed:

https://github.com/WordPress/twentytwentyone/blob/4337f375aaf679ac4471cb01a236eb08eff6d92e/assets/css/ie.css#L5703-L5704

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/636

## Summary

Moves the overflow rules to mobile-only.
The solution suggest in the comment works well when I test it in IE and I don't see any other side effects.

## Test instructions

1. Add a menu with submenu items.
1. View and test the menu with Internet Explorer 11
1. View and test the menu with other browsers to see that it does not affect the menu negatively.

## Screenshots
An animation of the visible sub-menu items in internet explorer

## Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -5747,8 +5747,6 @@ h1.page-title {
 	padding-right: 20px;
 	padding-bottom: 25px;
 	background-color: #d1e4dd;
-	overflow-x: hidden;
-	overflow-y: auto;
 	transition: all .15s ease-in-out;
 	transform: translateY(30px);
 }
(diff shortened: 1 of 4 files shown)


aristath (approved):
Code looks good to me, but I don't have IE to test it.
If anyone on a Windows PC can confirm then we can merge it :+1:

aristath:
Tested in IE11, works fine 👍

PR #650, opened 2020-10-22
Add min-width & min-height to pagination links

### LINKED ISSUE #533: Pagination links: the clickable area is too small

aristath:
Fixes #533

## Summary
Make sure the clickable area is at least 44x44px. 

## Quality assurance

### DIFF
--- a/assets/sass/06-components/pagination.scss
+++ b/assets/sass/06-components/pagination.scss
@@ -25,6 +25,11 @@
 
 	.nav-links {
 
+		> * {
+			min-width: 44px;
+			min-height: 44px;
+		}
+
 		.nav-next a,
 		.nav-previous a {
 			display: flex;
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -4924,9 +4924,11 @@ a.custom-logo-link {
 }
 
 .page-links .post-page-numbers {
+	display: inline-block;
 	margin-left: 13px;
 	margin-right: 13px;
-	padding: 10px 0;
+	min-width: 44px;
+	min-height: 44px;
 }
 
 .page-links .post-page-numbers:first-child {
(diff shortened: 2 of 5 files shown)


melchoyce (approved):
LGTM 👍

carolinan:
This improvement needs to be added to the columns pagination and content pagination (.page-links for wp_link_pages() ).

aristath:
> This improvement needs to be added to the columns pagination and content pagination (.page-links for wp_link_pages() ).

Done.
These used to have a vertical padding which made them 42.8px-tall.
The padding was removed and replaced with max-height & max-width.

carolinan:
Sorry for the fatal typo here: ` columns pagination` 🤣 🤦  , I have confirmed that the comment pagination is also the correct size.

assets/sass/06-components/pagination.scss
```diff
@@ -25,6 +25,11 @@
 
 	.nav-links {
 
+		> * {
```
afercia, line 28: Just a quick question: why the direct children + universal selector pattern `> *` is used so widely across the theme's CSS? Not sure in this case there's the need to use it. Seems to me all the targeted elements do have a specific CSS class.

afercia: Update: quickly discussed this on Slack, agreed to consider later refinements. Not a blocker.

PR #666, opened 2020-10-23
Change text color to white in IE if using a dark background

aristath:
## Summary
In IE11, when using a dark background-color the text doesn't switch to white and is unreadable.
This PR detects if the user is on IE and if they are using a dark background-color then it switches the text-color to white.

## Quality assurance

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -91,27 +91,27 @@
 }
 
 .wp-block-button__link:before {
-	margin-bottom: -calc(1em - 0);
+	margin-bottom: -calc(1em + 0);
 }
 
 .wp-block-file .wp-block-file__button:before {
-	margin-bottom: -calc(1em - 0);
+	margin-bottom: -calc(1em + 0);
 }
 
 .wp-block-search .wp-block-search__button:before {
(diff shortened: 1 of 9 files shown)


carolinan:
This is so much better, but there are still elements where the specificity overrides the colors.
<del>I think this specificity can be removed but it needs to be tested.</del>
_Or it can't be removed of course, because it allows us to use the variables to change these._

Example:
.entry-title , .site-title when not linked

All the text and links in the footer:
`.footer-navigation-wrapper li a:link`
`.site-footer > .site-info a:visited`

aristath:
Added more selectors, I don't see anything else missing now 👍

carolinan:
The responsive menu has the mint green background color when it is opened, do we open a separate issue for this?
responsive-ie

carolinan:
Skip link is not readable when focused:

ie-skiplink

carolinan:
The button block, submit and search buttons work well on the default background color.
When the background color is dark, the text is invisible when you hover over/activate the button.

The dark color I had selected from the palette, is the dark color that is also the default text color 🤦

aristath:
> The responsive menu has the mint green background color when it is opened, do we open a separate issue for this?

Fixed

PR #677, opened 2020-10-23
Button styles: refactor extend to mixin

jffng:
This is a PR based on #637 and @scruffian's [comment](https://github.com/WordPress/twentytwentyone/pull/637#issuecomment-715268433) about using an argument-less mixin as opposed to an extend to apply the button styles.

**Context**
Extends in Sass work by trying together selectors, and this can result in undesirable side effects. More on that [here](https://csswizardry.com/2014/11/when-to-use-extend-when-to-use-a-mixin/) if you're curious.

**Result**
In this case, using a mixin actually generated more lines of code. But the benefit is the resulting styles are easier to read + overwrite, and unexpected behaviors between elements that use a shared button style will be avoided in the future.

Nothing should be changed visually about the button styles; this is a code quality PR that can either be merged to #637 or as a follow up.


scruffian:
LGTM

PR #690, opened 2020-10-26
Remove all things dark-mode

aristath:
## Summary
Removes dark-mode from the theme. A plugin will be created instead, so that we can experiment on iterations easier without blocking the overall theme development. 
Depending on the progress and time, we may re-integrate the plugin inside the theme.

Note: The new repo for dark-mode is on https://github.com/WordPress/twentytwentyone-dark-mode

## Quality assurance


assets/js/customize-preview.js
```diff
@@ -43,15 +43,6 @@
 				textColor = ! isDark ? 'var(--global--color-dark-gray)' : 'var(--global--color-light-gray)',
 				tableColor = ! isDark ? 'var(--global--color-light-gray)' : 'var(--global--color-dark-gray)';
 
-			// Modify the body class depending on whether this is a dark background or not.
```
carolinan: This class is used for other color settings, like changing the background of the input fields, so it needs to be kept.

assets/sass/07-utilities/ie.scss
```diff
@@ -1,45 +0,0 @@
-.is-IE {
```
carolinan: Why is this removed? I thought this was for any dark background, not specifically for the device dark mode

classes/class-twenty-twenty-one-custom-colors.php
```diff
@@ -160,26 +151,4 @@ public static function get_relative_luminance_from_hex( $hex ) {
 		$lum = ( 0.2126 * $red ) + ( 0.7152 * $green ) + ( 0.0722 * $blue );
 		return (int) round( $lum );
 	}
-
```
carolinan: This body class is used to soften the color of the input field background so needs to be kept.

aristath: :thinking: where? Ah, I assume it's on another PR that hasn't been merged yet? 'Cause there's nothing like that on `trunk`

aristath: Added back :+1:

assets/js/customize-preview.js
aristath: Added back

assets/sass/07-utilities/ie.scss
aristath: You're right. added back :+1:

PR #692, opened 2020-10-26
Fix layout shifts for featured images

aristath:
## Summary
Google pagespeed reports a layout shift: When a page/post has a featured image, the image initially has zero height which gets re-calculated when the image loads. This is caused by the `img{height:auto;}` styles we have. Unfortunately that style can't be removed because it handles scaling images without distorting their ratios in smaller screen-sizes.
This PR tries to fix that shift by adding a height to the image.
After some research and trying many implementations, this one is the simplest, most robust and least complicated. Alternative solutions included adding calculated paddings to the wrapper element, changing the position of the image to absolute and other hacks with css-variables, pseudo-elements, or even client-side JS.
Adding inline styles to the element is not the perfect solution, but it's better than the alternatives, and definitely worth it. With this change we're fixing Cumulative Layout Shift (CLS) at scale and we do it server-side so it's a lot easier.

## Quality assurance

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -361,9 +361,8 @@ figcaption {
 	padding: 0;
 }
 
-.post-thumbnail img.attachment-post-thumbnail {
-	width: 100%;
-	height: calc(attr(width)/attr(height)*1%);
+img.has-calculated-ratio {
+	height: undefined;
 }
 
 select {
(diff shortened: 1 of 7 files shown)


carolinan:
I am not sure how to test this personally, but it has been tested locally and on a live site by @aristath, @jdevalk and @jono-alderson, so I will merge it.

Issue #680, opened 2020-10-23
Navigation Block: Style submenus


melchoyce:
**Describe the bug**
The submenu styling for the Navigation Block inherits core Gutenberg styles, which don't match the aesthetics of the theme.

**To Reproduce**
1. Create a new page or post.
2. Add a navigation block and select a menu that contains a submenu.
3. Hover over the submenu.
4. See the drop-shadow on the submenu.

**Expected behavior**
The submenu styling should match the design of the header navigation.

**Screenshots**
![image](https://user-images.githubusercontent.com/2846578/97048339-42f34580-1548-11eb-8098-c96be17ed4cc.png)
![image](https://user-images.githubusercontent.com/2846578/97048369-4d154400-1548-11eb-96d8-c961fbcfb2af.png)

**Editor version (please complete the following information):**
- WordPress version: 5.5.1
- Gutenberg version: 9.2.1

**Desktop (please complete the following information):**
 - OS: macOS 10.15.6
- Browser: Chrome 86.0.4240.80 (Official Build) (x86_64)

**Additional context**
See also #660 for a discussion on submenu styling.

carolinan:
Do we also want to change the styling in the editor or will this affect the editing experience / usability when setting up the menu?

JustinyAhin:
> Do we also want to change the styling in the editor or will this affect the editing experience / usability when setting up the menu?

I don't believe we should do that. For the navigation block, the way it works in the editor is slightly different. For instance for with menu item that has a submenu, you need to click on the submenu `<ul>` item to add other sub submenus.

JustinyAhin:
Closing this as it was fixed in #695

PR #695, opened 2020-10-26
Style submenus container

### LINKED ISSUE #660: Improve menu visibility
The menu can be difficult to use since there is little indication of change on hover and the text looks the same as the rest of the page. I get lost.
Can you tell there is a menu open here?
menu

What if there were some box-shadow or something to distinguish the menu from the other parts of the page? Or it could be the border that is used for keyboard navigation.
menu-with-shadow

JustinyAhin:
Fixes #660 & #680.

## Summary
Style primary submenu container.

No

## Test instructions

This PR can be tested by following these 

1. Add a menu to the theme which contain submenus
2. Hover over the menus items with ths "+" symbol

## Screenshots
Capture d’écran du 2020-10-26 09-59-12
Capture d’écran du 2020-10-26 09-56-10
Capture d’écran du 2020-10-26 09-55-29

## Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -3471,7 +3471,7 @@ dd {
 	padding: 0;
 	position: absolute;
 	top: 100%;
-	border: 1px solid #28303d;
+	border: 1px solid #39414d;
 }
 
 .wp-block-navigation:not(.has-background) .wp-block-navigation__container {
@@ -6035,7 +6035,7 @@ h1.page-title {
 	padding: 0;
 	list-style: none;
 	margin-left: 13px;
-	border: 1px solid #28303d;
+	border: 1px solid #39414d;
 }
 
 @media only screen and (min-width: 482px) {
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -6034,6 +6034,7 @@ h1.page-title {
 	padding: 0;
 	list-style: none;
 	margin-left: 13px;
+	border: 1px solid #39414d;
 }
 
 @media only screen and (min-width: 482px) {
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -3466,12 +3466,40 @@ dd {
 
 .wp-block-navigation > .wp-block-navigation__container > .has-child > .wp-block-navigation__container {
 	background: #d1e4dd;
-	box-shadow: 1px 1px 3px 0 rgba(0, 0, 0, 0.2);
 	margin: 0;
 	padding: 0;
 	position: absolute;
 	top: 100%;
-	border: 1px solid #39414d;
+	border: 1px solid #28303d;
+}
+
+.wp-block-navigation > .wp-block-navigation__container > .has-child > .wp-block-navigation__container:before {
+	content: "";
+	display: block;
+	position: absolute;
+	width: 0;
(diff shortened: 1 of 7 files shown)


JustinyAhin:
Ping @joyously

joyously:
Looks better to me! but it doesn't match the [mockup](https://github.com/WordPress/twentytwentyone/issues/660#issuecomment-715451116) I saw, which had the border around the parent item also (I don't know if that matters). 
In the dark version the border seems barely discernible.
Can you show a Navigation block screenshot? (your test instructions are not coherent)
That [mockup](https://github.com/WordPress/twentytwentyone/issues/680#issue-728488907) seems different.

JustinyAhin:
>  In the dark version the border seems barely discernible.

I've updated the fix, and added it to the navigation block too.

What I did basically to both primary menus and navigation blocks that have submenus is adding a 1px solid light-gray (#39414d) border to them.

It looks fine for both with the default background color, but could be more visible on darker background colors (I'm not sure we can handle this before correctly supporting dark mode in the theme -- cc @aristath ).

Primary menu on default light background
Capture d’écran du 2020-10-26 15-08-13

Primary menu on dark background
Capture d’écran du 2020-10-26 15-07-11

Navigation block on default light background
Capture d’écran du 2020-10-26 15-08-30


Navigation block on dark background
Capture d’écran du 2020-10-26 15-07-31

melchoyce:
Thanks for working on this, @JustinyAhin. Can you update the design to look like this?

image
[Figma Mockup](https://www.figma.com/file/OEtsCj8oD8JPVx4OvsS0U2/Twenty-Twenty-One?node-id=921%3A38)

JustinyAhin:
@melchoyce best I think can be done is something like this:
Capture d’écran du 2020-10-26 15-56-11

Because the submenu `<ul>` is larger than the menu item, it's a bit tricky to make the border-top disappear, while leaving a small part for the overflow.

There is also a slight offset to the right of the submenu, but I can find out where it comes from, and undo it.

aristath:
Thinking out loud:
Normally I'd suggest removing the bottom border from the parent, then changing its z-index so it's above the submenu, and finally moving the submenu up by 1px. That _should_ make it appear as if they are one like in the mockup.
But then we may have a problem if the submenu is not as wide as the parent element. Unless we also add min-width:100% to the submenu? :thinking: 
Should be possible to accomplish though I haven't tried it yet so there may be unforeseen complications with all of the above I described

melchoyce:
All good points. Would something like this be any less complicated?

image

ryelle:
I've pushed an attempt at the arrow design to this PR. I created a new variable for the border color which inherits from the global color (so it'll flip to white when a dark background is picked). The border only applies to the 2nd level menu, so we don't have nested borders.

Screen Shot 2020-10-26 at 6 44 11 PM
Screen Shot 2020-10-26 at 6 43 15 PM

Navigation block:

It should match the primary nav now. I removed the shadow, and added in the hover style (which was missing). I didn't change anything for the editor style, I think that can be done in another PR if necessary.

Screen Shot 2020-10-26 at 6 44 32 PM

carolinan:
Both the menu and block looks good to me.

PR #696, opened 2020-10-26
Add has-background to the text color utility styles

carolinan:
## Summary

_This description only applies to blocks that has no custom text color._

When the body background is changed to dark, the primary text color is changed to white.
When a block is placed inside a block and that block has a background color, 
changing the text color to white automatically can make the text unreadable.

A style was created to make sure that if the parent blocks background color is black, dark grey or grey, the text color is white. 
And if any other palette color is selected, the text color is dark grey.

-This PR adds the has-background class to this style, so that if a **custom** background color is selected,
not from the palette but with the picker, the text color is dark grey.

_This does not solve the problem where the dark grey has a low contrast with the background, for that scenario, the user has to select a readable text color._

## Test instructions

1. Change the body background to dark
1. Add a group block with a background color not from the palette
1. Add some blocks inside of it, for example quote, verse, table
1. Duplicate this and change the background color to one from the palette.
1. Compare the editor and front.

## Screenshots

Before:
Left, editor, right, front.

before

After:
Left, editor, right, front.
after

## Quality assurance

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -2712,6 +2712,10 @@ a {
 	color: #fff;
 }
 
+:not(.has-text-color).has-background {
+	color: #28303d;
+}
+
 :not(.has-text-color).has-green-background-color[class] {
 	color: #28303d;
 }
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -6705,6 +6705,10 @@ h1.page-title {
 	color: #fff;
 }
 
+:not(.has-text-color).has-background {
+	color: #28303d;
+}
+
 :not(.has-text-color).has-green-background-color[class] {
 	color: #28303d;
 }
--- a/assets/css/style-editor.css
+++ b/assets/css/style-editor.css
@@ -1935,7 +1935,7 @@ a {
 	color: var(--global--color-white);
 }
 
-:not(.has-text-color).has-green-background-color[class], :not(.has-text-color).has-blue-background-color[class], :not(.has-text-color).has-purple-background-color[class], :not(.has-text-color).has-red-background-color[class], :not(.has-text-color).has-orange-background-color[class], :not(.has-text-color).has-yellow-background-color[class], :not(.has-text-color).has-white-background-color[class] {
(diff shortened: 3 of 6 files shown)


aristath (approved):
Since we already have the exclusions for dark backgrounds above, I think it makes sense :+1:

carolinan:
No matter if the text is black or white, there will be color issues since the user does not select a custom text color.

ryelle:
Is this actually an issue in the verse block, not correctly inheriting the color? The editor/frontend seems the same for paragraphs, quotes, tables:

| Editor | Frontend |
|-------|----------|
| Screen Shot 2020-10-26 at 12 08 35 PM | Screen Shot 2020-10-26 at 12 08 40 PM |

ryelle:
I created #700 to fix the issue I noticed above.

For this issue, I'm not sure it's necessary to add an extra condition— as you say, this doesn't automatically detect contrast, so we're really exchanging one problem for another. If a user adds a light background color to their content, they can see in the editor that it's unreadable, and they should change the text color.

carolinan:
The verse block was not selected deliberately for this PR. The problem is that not all blocks are tested so we don't know which blocks work and not.

carolinan:
There is no logic in changing the color for the palettes, but not the custom colors either. 
This is why the color was brought up (at the very very end) of the monday meeting on slack.

ryelle:
We change the text color with the palette colors because we know what those colors are. For custom colors, there's no way of knowing whether the user chose a dark or light color, so we have no way of setting a readable color 100% of the time. If we set this to a dark text color, it'll be low-contrast when the user picks a dark background color. If we leave it as is, it's low-contrast when the user picks a light background color (and vice-versa for light background sites).

I'm suggesting that we not add this extra class, because the user will see that the text is not readable & they need to set a text color too.

carolinan:
The theme only changes the text color because the body background is dark.
When a block is added, and the background is not dark, it also makes no sense that the color would be white there.
It is the white color that is the addition, not the reverting of the white color to the default.
But there is no good solution, so lets close it.

Issue #688, opened 2020-10-25
Better handling of "Archive" pages. 


bph:
One of the first things, we do with every theme is to remove the words "Categories" or "Archive" from the category or tag pages, and add the display of the descriptions field. 

For SEO purposes there is no value to have those words on the page (it might hurt the page rankings)  but it definitely would help if there is a way to add additional text to the category pages in the description field. 

I am aware that this might not be core territory. However, there were recent changes, like XML Sitemaps that made it into core. And this idea seems to be much less complicated and would go a long way for normal users. 

It's past Beta stage, so having a new feature like a toggle switch to switch on and off the described behavior would be for a future version. For now, it should just be default behavior of a theme.

bph:
It' probably to distracting for the default theme. On second thought I close this...

PR #697, opened 2020-10-26
Add missing archive descriptions

aristath:
I realized we're missing archive descriptions while reading https://github.com/WordPress/twentytwentyone/issues/688

This PR adds the descriptions in archives. They should be there out of the box in default themes for many reasons - including SEO. By not showing them we are hiding content that the user has entered and obviously expects to be used (otherwise why enter a description in the first place). To me this is a bug...

Before:

Screenshot_2020-10-26 Block – My Blog(1)

After:

Screenshot_2020-10-26 Block – My Blog

## Quality assurance

### DIFF
--- a/archive.php
+++ b/archive.php
@@ -16,6 +16,9 @@ get_header();
 
 	<header class="page-header alignwide">
 		<?php the_archive_title( '<h1 class="page-title">', '</h1>' ); ?>
+		<div class="archive-description">
+			<?php echo wp_kses_post( wpautop( get_the_archive_description() ) ); ?>
+		</div>
 	</header><!-- .page-header -->
 
 	<?php while ( have_posts() ) : ?>
--- a/archive.php
+++ b/archive.php
@@ -10,15 +10,17 @@
  */
 
 get_header();
+
+$description = get_the_archive_description();
 ?>
 
 <?php if ( have_posts() ) : ?>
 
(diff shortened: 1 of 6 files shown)


carolinan (approved):

melchoyce:
I think the description size should increase. I tried two options:

LG:
image

XL:
image

Both options use the heading line-height variable.

Large _feels_ more reasonable, but the line-length is too long. XL's line-length ends up being within an acceptable readability range. 

We could also limit the max-width for the description, and have the edge align with the right edge of the content width:

image

I'm leaning towards XL and full-width. @kjellr, can I get your input here?

kjellr:
Yeah, I like XL and full width too. 👍

melchoyce:
Aside, any way to drop the "Category Archives:" part of the heading? It always feels so unnecessarily meta to me.

aristath:
Per the suggestions above, I changed the font-size to `--global--font-size-xl` and the line-height to `--global--line-height-heading`. Since these were approved I'll go ahead and merge the PR.

> Aside, any way to drop the "Category Archives:" part of the heading? It always feels so unnecessarily meta to me.

I created a new ticket on https://github.com/WordPress/twentytwentyone/issues/705
This one will require some discussion to see what labels we want to show so it's better if there's a separate PR and ticket for it 👍

PR #698, opened 2020-10-26
let overlapping columns use the outer block background color

### LINKED ISSUE #682: Overlapping Columns: Paragraphs should inherit container background
**Describe the bug**
When you create an overlapping columns block, and add a paragraph to the column that is overlapping, it includes a background color so the text can always be read:

image

**To Reproduce**
1. Create a new post or page.
2. Add a cover block, and select a background color that isn't green.
2. Add a columns block inside of the cover block, and select the "overlapping" block style.
3. Add something like an image to the left column, and a paragraph to the right column.
4. See that the paragraph's background color is green, not the color you chose for your cover block.

**Expected behavior**
I think we should inherit that color from the container's background, instead of always using green.

- WordPress version: 5.5.1
- Gutenberg version: 9.2.1

- OS: macOS 10.15.6
- Browser: Chrome 86.0.4240.80 (Official Build) (x86_64)

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/682

Partial fix only. In my opinion, this style is **huge** and I don't think it is worth adding, it is a lot of CSS for a small win.

## Summary

Since the paragraph can not inherit the great grandparents background color,
add the same background color to the paragraph, _if the palette color is used._

## Test instructions

1. Add a group block and cover block with a background color setting
1. Add a two column block inside, and select the overlapping style.
1. Add content like an image to the left, and a paragraph to the second, overlapping column.
1. Compare the background color of the outer block and the paragraph.

## Quality assurance

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -656,6 +656,7 @@ a:hover {
 		margin-left: -50px;
 		margin-top: 63px;
 		z-index: 2;
+		background: inherit;
 	}
 	.wp-block-columns.is-style-twentytwentyone-columns-overlap .wp-block-column:nth-child(2n) > p:not(.has-background) {
 		background-color: #d1e4dd;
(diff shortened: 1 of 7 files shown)


carolinan:
This is not going to be an acceptable solution, closing.

PR #701, opened 2020-10-26
Increase button focus offset

melchoyce:
## Summary
Increase the `outline-offset` on buttons from 4px to 6px, so the focus style is never flush against the border of the button.

## Test instructions

1. Go to any page that has a button present, such as the 404 page.
2. Test the button's focus styles.

## Screenshots

Before | After
----- | -----
 | 
 |


aristath (approved):
LGTM 👍

PR #708, opened 2020-10-27
Make sure submenus don't go off-screen horizontally

### LINKED ISSUE #703: Navigation: Dropdown can be pushed off the screen
**Describe the bug**
If you have a long menu and a narrow browser viewport, it's possible for a dropdown towards the end of the menu to fly off screen:

**To Reproduce**
1. Create a long navigation menu and assign it to the primary navigation location.
2. Next some subpages underneath the right-most navigation menu item.
3. If your viewport isn't narrow enough for the menu to be close to the edge, narrow it further.
4. Hover over the right-most menu item and see the dropdown fly off screen.

**Expected behavior**
Ideally, we'd push the menu left, instead of pushing it right off screen.

aristath:
Fixes #703

## Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -6125,6 +6125,42 @@ h1.page-title {
 	.primary-navigation > div > .menu-wrapper > li > .sub-menu li {
 		background: #d1e4dd;
 	}
+	.primary-navigation > div > .menu-wrapper > li > .sub-menu.submenu-reposition-left {
+		/* rtl:ignore */
+		left: 0;
+		/* rtl:ignore */
+		right: auto;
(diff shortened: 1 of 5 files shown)


carolinan:
I don't see any problems when there is only one sub menu level
but if I enable the "testing menu" from the theme unit test which has 10 levels, 
the sub menu slides to the left but the last levels does not fit.

menu-left

aristath:
Refactored, should be OK now 👍

carolinan:
This works much better, thank you. Tested and confirmed on RTL as well. Using Chrome and Firefox on PC.

PR #718, opened 2020-10-28
Fix starter content url

mukeshpanchal27:


carolinan:
Hi, what is the reason and background for this?
The permalink for the article that is being worked on, does include the dashes.

PR #722, opened 2020-10-28
Fix separator alignment issue in the editor

### LINKED ISSUE #714: Block editor: Separator is full width inside the editor while it’s centered on the front
**Describe the bug**
When you insert a separator block inside the editor, it appears full width. On the front, it appears centered.

**To Reproduce**
Inside the code editor, insert a separator block or paste the following code:

```
<!-- wp:separator -->
<hr class="wp-block-separator"/>
<!-- /wp:separator -->
```

**Screenshots**

| Editor  | Front |
| ------------- | ------------- |
| separator-editor | separator-front |

JustinyAhin:
Fixes #714.

## Summary
Fix separator alignment issue in the editor.

No

## Test instructions

1. Add a separator block in the editor
2. Switch between the different width settings
3. The block in the editor should match with it in frontend

## Quality assurance

### DIFF
--- a/assets/sass/05-blocks/separator/_editor.scss
+++ b/assets/sass/05-blocks/separator/_editor.scss
@@ -10,7 +10,7 @@ hr {
 	}
 
 	&:not(.is-style-dots) {
-		max-width: inherit;
+		max-width: var(--responsive--aligndefault-width);
 	}
 
 	&.is-style-twentytwentyone-separator-thick {
@@ -18,7 +18,6 @@ hr {
 	}
 
 	&.is-style-dots {
-
 		border-bottom: none;
 
 		&.has-background,
@@ -40,5 +39,17 @@ hr {
 	[style*="background-color"]:not(.has-background-background-color) &,
 	.wp-block-cover[style*="background-image"] & {
 		border-color: currentColor;
+	}
+}
+
+.wp-block[data-align="full"],
+.wp-block[data-align="wide"] {
+
+	.wp-block-separator,
+	hr {
+
+		&:not(.is-style-dots) {
+			max-width: inherit;
+		}
 	}
 }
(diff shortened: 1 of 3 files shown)


assets/sass/05-blocks/separator/_editor.scss
```diff
@@ -42,3 +41,15 @@ hr {
 		border-color: currentColor;
 	}
 }
+
+.wp-block[data-align="full"],
```
carolinan: This needs to be inside the .wp-block-separator, otherwise the default separator will be full width when placed inside another full width block like a group.

JustinyAhin: > This needs to be inside the .wp-block-separator, otherwise the default separator will be full width when placed inside another full width block like a group.

It wouldn't work this way since the element with the class `.wp-block[data-align=""]` always wrap the separator (when set as full or wide width) as parent.

What adding instead is adding the `>` selector before the separator class. Then this 
```
.wp-block[data-align="wide"] {

	> .wp-block-separator,
	> hr {

		&:not(.is-style-dots) {
			max-width: inherit;
		}
	}
}
```
will only applies if the separator is the direct children of an element with the class `.wp-block[data-align=""]`.

carolinan: I have been placing it inside the .wp-block-name class, and then used the `&` as a parent selector, but I did not know sass before this project started, so I don't know best practices.

https://sass-lang.com/documentation/style-rules/parent-selector

ryelle: I don't think there are any set standards around this, but we do use the `&` a lot to avoid repetition.

I'd suggest moving this up into the `.wp-block-separator, hr` block, and in fact you don't need the `:not(.is-style-dots)` specificity hack, so it can be simplified to:

```scss
[data-align="full"] > &,
[data-align="wide"] > & {
	max-width: inherit;
}
```

JustinyAhin: Should be fine now.

carolinan (approved):

PR #725, opened 2020-10-28
Portfolio List block pattern: fix image dimensions

kjellr:
The Portfolio List pattern was [originally designed to be wide-width](https://github.com/WordPress/twentytwentyone/issues/714#issuecomment-717885639), but when it was built in #299, wide width separator blocks were not possible yet. This PR changes the pattern to be wide width so that it better matches the original design. 

Also, when the images were swapped out to these ones, the dimensions were not updated. This PR makes that change too, ensuring that the ragged-left edge of the images is restored. 

Before|After
---|---
Screen Shot 2020-10-28 at 8 09 54 AM|Screen Shot 2020-10-28 at 8 19 08 AM


kjellr:
One thing to discuss: since wide/full separators are not supported in WP 5.5, this PR will display like this in the editor. 

Screen Shot 2020-10-28 at 8 37 51 AM

The separators do have the `alignwide` and `align:wide` class/attribute, but the older version of Gutenberg appears to strip them out when it displays them in the editor.

carolinan:
Do we want to open a Gutenberg issue for that? I guess it has a purpose, like keeping the code valid.

I would suggest keeping the pattern the regular width. The before image.

The right aligned images in the last image, above this comment, looks off.

kjellr:
> The separators do have the alignwide and align:wide class/attribute, but the older version of Gutenberg appears to strip them out when it displays them in the editor.

On second look, it _does_ still have the `alignwide` class, but since Gutenberg doesn't wrap the block in a `<div class="wp-block" data-align="right">` wrapper, it doesn't appear as the wide width.

melchoyce:
If we can upstream this, that would be cool — if not, I'm okay with a _little_ awkwardness in the editor for older versions of WordPress. The pattern looks best when wide-width IMO.

kjellr:
I took a closer look at this, and I'm not sure that we'll be able to get a better treatment here for 5.5. Because of the way the block is structured in that release, it requires a `<div class="wp-block" data-align="wide">` wrapper, and we can't get that added without an update that would.. which would defeat the purpose here. 😄

I think it probably makes the most sense to just keep the default alignment, since the alternative [looks broken](https://github.com/WordPress/twentytwentyone/pull/725#issuecomment-717907372) in 5.5. 

If that sounds good to everyone else, I can strip that part out of this PR. We'll still want the enhancements to the image sizes that this PR adds.

kjellr:
Ok, the wide alignment stuff is removed. This PR now only adjusts the image sizes so they are proportionally correct. 

Before|After
---|---
Screen Shot 2020-10-29 at 3 06 06 PM|Screen Shot 2020-10-29 at 3 05 41 PM

melchoyce:
I still plan on keeping it wide in the demo: https://2021.wordpress.net/work/

Do you think that's okay?

kjellr:
I think that's ok. 👍

PR #726, opened 2020-10-28
Update margin for the first social link

### LINKED ISSUE #716: Block editor: Social icons aren’t correctly aligned
**Describe the bug**
When trying to create a list with 2 social icons using the Social Icons block, the icons aren’t aligned inside the editor. 

**To Reproduce**
Inside the code editor, paste the following code:

```
<!-- wp:social-links -->
<ul class="wp-block-social-links"><!-- wp:social-link {"url":"#","service":"wordpress"} /-->

<!-- wp:social-link {"url":"#","service":"github"} /--></ul>
<!-- /wp:social-links -->
```

**Screenshots**

Block not focused:
social-icons

Block focused:
social-icons-focus

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/716

## Summary

Replaces `margin-top: auto` with `margin-top: var(--global--spacing-vertical);` to override the generic top margin for the first child inside a block.

## Test instructions

1. Add a social icons block with two icons, no more, no less.
1. Confirm that the icons are aligned next to each other in the editor.
1.

## Quality assurance

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -2137,8 +2137,9 @@ hr {
 	border-color: currentColor;
 }
 
-.wp-block-social-links li.wp-block-social-link:first-child {
-	margin-top: 30px;
+.wp-block-social-links [data-block] {
+	margin-top: 0;
+	margin-bottom: 0;
 }
 
 .wp-block-social-links.is-style-twentytwentyone-social-icons-color button {
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -2138,7 +2138,7 @@ hr {
 }
 
 .wp-block-social-links li.wp-block-social-link:first-child {
-	margin-top: auto;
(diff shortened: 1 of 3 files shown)


carolinan:
It would be better to find a new solution for https://github.com/WordPress/twentytwentyone/blob/trunk/assets/sass/05-blocks/utilities/_editor.scss#L81, one that does not require overriding this style for individual blocks.
It also affects https://github.com/WordPress/twentytwentyone/issues/727

ryelle:
I think we can clear the margins off these link items instead of overriding the `utilities/_editor.scss` style. This matches the frontend better (on trunk the editor has too much space).

> It would be better to find a new solution […] that does not require overriding this style for individual blocks.

I _think_ this does work for groups & columns and other basic nested blocks, and these exceptions will each have specific solutions, so continuing this way makes sense to me.

PR #731, opened 2020-10-28
Fix margins on Media & Text with vertical alignment

### LINKED ISSUE #727: Media & Text vertical align text position is not the same in the editor and the front.
**Describe the bug**
When the vertical top alignment is chosen for the text part of the Media & text block,
the vertical spacing is not the same in the editor and the front.

**To Reproduce**
1. Add a media and text block.
2. Select the _Vertically Align Top_ option.
3. See the position of the text. Save.
4. Compare the position in the editor with the front.

The difference is easier to see with the borders block style.

**Expected behavior**
The editor and front should match.

**Screenshots**

Editor:
editor

Front:
front

- WordPress version: 5.6-beta2-49352
- Does the website has Gutenberg plugin installed, or is it using the block editor that comes by default? "default"

ryelle:
Fixes #727

> When the vertical top alignment is chosen for the text part of the Media & text block, the vertical spacing is not the same in the editor and the front.

## Test instructions

1. Add a Media & Text block, with some content in the text part
2. Set the vertical alignment to "top"
3. The content should move upward, but there's still some space above the text
4. View the frontend, it should look the same

## Screenshots
Editor
Screen Shot 2020-10-28 at 5 31 57 PM

Frontend
Screen Shot 2020-10-28 at 5 32 04 PM

## Quality assurance

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -1541,9 +1541,16 @@ dt {
 	font-weight: bold;
 }
 
-.wp-block-media-text .block-editor-inner-blocks {
-	padding-right: 25px;
-	padding-left: 25px;
+.wp-block-media-text .wp-block-media-text__content {
+	padding: 25px;
+}
+
+.wp-block-media-text .wp-block-media-text__content [data-block]:first-child {
+	margin-top: 0;
+}
+
+.wp-block-media-text .wp-block-media-text__content [data-block]:last-child {
+	margin-bottom: 0;
 }
 
 .wp-block-media-text.is-style-twentytwentyone-border {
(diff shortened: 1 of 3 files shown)


carolinan:
I did not understand why the `&.has-child-selected` was needed.
For the Media & Text block, there is a shift in the editor when text content is selected, both with and without this addition 🤔

ryelle:
> I did not understand why the &.has-child-selected was needed.

Ah, that shouldn't've made it into this branch - that rule should only apply when the appender is visible, and it's only visible when `has-child-selected` … but that's unrelated to this fix, so I've dropped it.

PR #733, opened 2020-10-29
Add a variable for admin-bar height

aristath:
Adds a `--global--admin-bar--height`variable which can be used consistently everywhere instead of having to add hardcoded values like 32px, 46px etc depending on the viewport.


carolinan:
LGTM

PR #734, opened 2020-10-29
Split navigation to separate template-part

aristath:
## Summary
Splits the nav template to a separate template-part.
Makes it easier to override the nav without having to rewrite the header.

## Quality assurance


carolinan:
Makes sense, and no issues found during testing.

PR #740, opened 2020-10-29
Update the position of the latest comments block

### LINKED ISSUE #736: Latest comments block is positioned to the left in the editor
**Describe the bug**
When a comments block without alignment (default) is placed in the editor,
it is positioned all the way to the left instead of in the middle.

On the front, the block is in the middle.

**To Reproduce**
1. Add a latest comments block
2. See the position, save.
3. Compare the position with the position on the front.

**Expected behavior**
I expect the block to be positioned in the middle in the editor.
I expect the front and editor to match.

**Screenshots**
latest-comments-editor

WordPress version: 5.6-beta2-49360
Does the website has Gutenberg plugin installed, or is it using the block editor that comes by default?
"default"

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/736

## Summary

Deleted the blocks editor style, since the CSS inside it caused the incorrect positioning of the block.

While testing this solution, I noticed that the block was also positioned to the left on the front
when it was set to wide align. 
Removing `margin-left: 0;` from` .wp-block-latest-comments` solved this issue.

## Test instructions

1. Add a latest comments block. 
1. Clone this block multiple times to test all different alignments.
1. Compare the positions in the editor and the front.

</details>

## Quality assurance

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -1301,10 +1301,6 @@ h6 {
 	padding: 20px;
 }
 
-.wp-block-latest-comments {
-	margin-left: 0;
-}
-
 .wp-block-latest-posts {
 	padding-left: 0;
 }
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -3217,10 +3217,6 @@ img {
 	padding: 20px;
 }
 
-.wp-block-latest-comments {
-	margin-left: 0;
-}
-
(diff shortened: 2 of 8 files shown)


kjellr (approved):
Looks good, thank you! 

I think that `margin-left: 0` was originally supposed to eliminate this padding, but I'll open up another PR to take care of that: 

Screen Shot 2020-10-29 at 2 41 09 PM

PR #742, opened 2020-10-29
Tidy up full-width group block children when the group has padding

### LINKED ISSUE #729: Full width headings inside group block ignores padding
**Describe the bug**
When a group block has a background color, it has some additional padding:
```
.editor-styles-wrapper .wp-block-group.has-background {
    padding: var(--global--spacing-vertical);
}
```
When a heading is added inside the group block, and is set to be _full width_,
the heading ignores this padding and moves right up to the edge of the group block.

**To Reproduce**
1. Add a group block with a background color.
2. Add a heading block.
3. Set the heading to full width.

**Expected behavior**
Perhaps this is working as intended? I don't think it looks as good when text ignores the padding.

**Screenshots**

full-width-heading

WordPress version: 5.6-beta2-49352
Does the website has Gutenberg plugin installed, or is it using the block editor that comes by default? "default"

kjellr:
Fixes https://github.com/WordPress/twentytwentyone/issues/729

Group blocks get inner padding when they have a background color assigned, or when Twenty Twenty-One's "Frame" block style is active. When this is the case, full-width children need width and margin adjustments to compensate and remain full-width. 

To test: 

1. Add a Group block
2. Assign it a background color _or_ assign it the "Frame" block style.
3. Add a block inside the Group block, make it full width.
4. Ensure that it stretches all the way to the edges of the group block in the front end and editor. 

## Screenshots

Front End

Before|After
---|---
Screen Shot 2020-10-29 at 2 10 24 PM|Screen Shot 2020-10-29 at 2 09 00 PM

Editor

Before|After
---|---
Screen Shot 2020-10-29 at 2 10 37 PM|Screen Shot 2020-10-29 at 2 08 10 PM

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -825,6 +825,12 @@ a:hover {
 	padding: 30px;
 }
 
+.wp-block-group.is-style-twentytwentyone-border .wp-block-group__inner-container > [data-align="full"] {
+	max-width: calc(100% + 60px);
+	width: calc(100% + 60px);
+	margin-left: -30px;
+}
+
 .wp-block-group .wp-block-group__inner-container > *:last-child {
 	margin-bottom: 0;
 }
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -357,11 +357,11 @@ a:hover {
 }
 
 .wp-block-button__link:before {
-	margin-bottom: -calc(1em - 0);
+	margin-bottom: -calc(1em + 0);
 }
 
 .wp-block-button__link:after {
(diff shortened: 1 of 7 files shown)


carolinan:
Will this PR place text elements up against the edge of the screen again?

carolinan:
It does not work with a full width separator,
this style is overriding the new max width calculation:
```
hr.wp-block-separator:not(.is-style-dots):not(.alignwide).alignfull {
    max-width: var(--responsive--alignfull-width);
}
```

kjellr:
Good catch, thanks @carolinan. Should be all set now.

PR #743, opened 2020-10-29
Remove left padding from latest comments block.

kjellr:
Followup from #740. Removes the left padding from the Latest Comments block. 

**Editor**

Before|After
---|---
Screen Shot 2020-10-29 at 2 48 29 PM|Screen Shot 2020-10-29 at 2 47 00 PM

**Front End**

Before|After
---|---
Screen Shot 2020-10-29 at 2 48 35 PM|Screen Shot 2020-10-29 at 2 47 16 PM


melchoyce (approved):
LGTM 👍

PR #749, opened 2020-10-30
Override the link focus style specificity for buttons

### LINKED ISSUE #747: The link focus style background color is affecting the button block links
**Describe the bug**
After the change of the background color for focused links, the new background color also affects button blocks that has a background color.

**To Reproduce**
1. Add a button block with a link
2. Save and view the front
3. Focus on the link in the button

**Expected behavior**
The buttons should keep their focus style according to the design.

**Screenshots**
button-background-color

- WordPress version:   5.6-beta2-49392
- Does the website have Gutenberg plugin installed, or is it using the block editor that comes by default?  "default"

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/747

## Summary

The link focus style uses .site a:focus, this specificity is used to make sure that the themes link styles does not affect
the WordPress admin bar.

Because of this specificity, the background color for the focus state was affecting the button block links.
Adding .site to the button block works but it's hacky.

If there is a better solution where the .site is not needed at all, that would be preferred.

## Test instructions

1. Add a button with a link and a background color. It does not need to be a custom background color, so the filled button will work. Save.
1. View the front and focus on the button link.
1. Confirm that the background color is correct.
1. Change the body background color. 
1. Focus on the link.
1. Confirm that the background color is still correct.

## Quality assurance


luminuu:
When I use this PR and test it, my button get's a dark background and does not change on focus - I'm not sure if it's because this PR is already 3 days old. Maybe you can merge the current trunk into this PR @poena?

kjellr:
I sorted the merge conflicts and I think I'm seeing the same thing that @luminuu noted above: Preset buttons are not using the correct background color on the front end: 

Editor|Front End
---|---
Screen Shot 2020-11-02 at 11 04 20 AM|Screen Shot 2020-11-02 at 11 04 09 AM

When the button styles have the extra `.site` specificity, the background style rule there overrides our color preset rules. We could add `.site` specificity there, but as noted above, it would be great to figure out a way to do this without that specificity. I'll take a closer look now.

luminuu:
Closing this PR as @kjellr found a nice way to implement this in #776.

Issue #763, opened 2020-10-31
Bug quote block


naunha:
When inserting a quote block, the quotation mark is in the wrong position, overlapping the text
![Captura de Tela (134)](https://user-images.githubusercontent.com/5017547/97779423-af7bcf00-1b5c-11eb-956f-276dc59c1d2a.png)

aristath:
Closing as duplicate of #759

PR #766, opened 2020-10-31
Pullquote: open quote position and i18n

### LINKED ISSUE #759: Pullquote: Quotation mark positioning regression
**Describe the bug**
The text overlaps the quotation mark in the Pullquote block now:

image

**To Reproduce**
1. Make a new post or page.
2. Add a pullquote block.
3. See that the quotation mark overlaps the text now.

**Expected behavior**
It should look like this:

image
https://2021.wordpress.net/2020/10/16/a-love-of-nature/

- WordPress version: 5.6-beta2-49360
- Gutenberg: 9.2.2

felipeelia:
Fixes #759 and #763 

## Summary

By default, `blockquote::before` will have a position absolute. This PR changes that behavior for blockquotes inside the Pullquote block.

It also provides a way to translate the open quote sign.

The function that adds the Custom CSS could be placed somewhere else or merged with the existent one (if we change its name). As it is tricky to escape quotes, I'm calling simply `__()` instead of `esc_attr__()`, for example. Perhaps that could be changed.

## Test instructions

1. Change the translation of the character using
```
add_filter(
	'gettext',
	function( $translation, $text ) {
		if ( '“' === $text ) {
			return '«';
		}
		return $translation;
	},
	10,
	2
);
```
2. Check how it changes

## Quality assurance
* [ ] I have thought about any security implications this code might add.

### DIFF
--- a/assets/css/style-editor.css
+++ b/assets/css/style-editor.css
@@ -1247,6 +1247,7 @@ pre.wp-block-preformatted {
 	font-size: 3rem;
 	font-weight: 500;
 	line-height: 1;
+	position: static;
 }
 
 .wp-block-pullquote p {
--- a/assets/sass/05-blocks/pullquote/_editor.scss
+++ b/assets/sass/05-blocks/pullquote/_editor.scss
@@ -15,6 +15,7 @@
 		font-size: 3rem;
 		font-weight: 500;
 		line-height: 1;
+		position: static;
 	}
 
 	p {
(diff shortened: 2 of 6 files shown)


carolinan:
This works well, but we need a solution for both the pullquote, quote block, and legacy quotes

felipeelia:
Oh, sure thing @carolinan !

I can expand the code to deal with those scenarios too. I wasn't sure if this approach was something you'd like to use (or if you have something else in mind). I'll try to send the code later today or tomorrow.

By "legacy quotes" I guess you mean quotes created in the classic editor, right? Thanks

carolinan:
> By "legacy quotes" I guess you mean quotes created in the classic editor, right? Thanks

Yes

aristath:
Fixed in #787

PR #771, opened 2020-11-02
Set the menu fallback to false

### LINKED ISSUE #770: Primary navigation: Style the page menu
**Describe the bug**
If a site has a fall back page menu, and then switches to Twenty Twenty-One, the page menu is still in use.
Because of this, the theme needs to style the page menu even though it can't be activated when using the theme.

Currently the responsive menu is broken and there is a JavaScript error:
```
primary-navigation.js?ver=1.0.0:44 Uncaught TypeError: Cannot read property 'getBoundingClientRect' of null
    at twentytwentyoneSubmenuPosition (primary-navigation.js?ver=1.0.0:44)
    at twentytwentyoneToggleAriaExpanded (primary-navigation.js?ver=1.0.0:17)
    at HTMLButtonElement.mobileButton.onclick (primary-navigation.js?ver=1.0.0:96)
```

And the wide menu is displayed vertically instead of horizontally.

**To Reproduce**
1. Activate a different theme 
2. Make sure that no menu is set to the menu locations.
3. Switch to Twenty Twenty-One

**Expected behavior**
The fallback page menu should have the same style as the selected menu.

**Screenshots**
pagemenu

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/770

## Summary

Set the menu fallback call back to false

## Test instructions

1. Make sure that no menu is set to the menu locations.
2. Activate a different theme
3. Make sure that no menu is set to the menu locations.
4. Switch to Twenty Twenty-One
5. Confirm that no fallback menu is showing.

## Quality assurance

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -32,6 +32,7 @@
 							'depth'          => 1,
 							'link_before'    => '<span>',
 							'link_after'     => '</span>',
+							'fallback_cb'    => 'false',
 						)
 					);
 					?>
--- a/template-parts/header/site-nav.php
+++ b/template-parts/header/site-nav.php
@@ -28,6 +28,7 @@
 				'menu_class'      => 'menu-wrapper',
 				'container_class' => 'primary-menu-container',
 				'items_wrap'      => '<ul id="primary-menu-list" class="%2$s">%3$s</ul>',
+				'fallback_cb'     => 'false',
 			)
 		);
 		?>


aristath (approved):

sabernhardt:
The menus correctly do not have the fallback or container element with the code as it is, but the `false` values for both the `'fallback_cb'` and `'container'` args should be better without the quotes.
https://wordpress.stackexchange.com/questions/13564/wp-nav-menu-always-falls-back-to-a-menu#answer-13579

PR #774, opened 2020-11-02
Update post pagination spacing

### LINKED ISSUE #659: posts pagination links on various screen widths
On an archive with multiple pages, the links at the bottom can be difficult to discern, depending on the screen width, because they are much longer now that each has "Page" in front of the number.  And at every width, there is a large white space below the links, which could be used.
(And I wonder if it's a worse problem in a language that has a longer word for "Page".)

On the narrowest width, (this one is 562px), the Page links are not visible at all, so the user has fewer navigation options.
Post Formats – Page 2 – Local Current WP-45

When the screen is wide enough for the links to appear, they are sort of jumbled together with the other links. (this one is 598px)
Post Formats – Page 2 – Local Current WP-04

As the screen gets wider, the links look more like a sentence, which is quite confusing. (this one is 735px)
Post Formats – Page 2 – Local Current WP-18

It is similar at 834px.
Post Formats – Page 2 – Local Current WP-33

This one is 1008px, where it is finally wide enough to have the Page links separate from the other links.
Post Formats – Page 2 – Local Current WP-48

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/659

## Summary

Hide the page numbers on screens up to 822px using `desktop-only`
Added a top margin to increase the vertical spacing when the pagination wraps.
Centered the dots.

## Test instructions

1. View an archive with pagination
1. Reduce the browser window width
1.

## Screenshots
Before:
pagination-before

After:
pagination

## Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -6486,6 +6486,10 @@ h1.page-title {
 	flex-direction: column;
 }
 
+.navigation .nav-links .dots {
+	text-align: center;
+}
+
 @media only screen and (min-width: 592px) {
 	.navigation .nav-links {
 		display: flex;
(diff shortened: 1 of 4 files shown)


aristath (approved):
LGTM :+1:

kjellr (approved):
Thanks, this seems like a solid, simple approach for now. 👍

PR #775, opened 2020-11-02
Do not apply the image size filter to the logo

### LINKED ISSUE #760: my site logo is stretched
My logo looked fine the other day when I added it, but something changed since then (I just downloaded today's version of the theme zip).
The PNG file is 160px by 142px, no cropping done.

stretched-logo

In the Customizer, it looks fine when the logo is by itself.
stretched-logo

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/760

## Summary

Inside `twenty_twenty_one_get_attachment_image_attributes`, check if the image has the custom-logo CSS class.
If it does, return the attributes early without adding the filtered image size attribute.

## Test instructions

1. Add a custom logo, save and view the front
1. Check that the image is not stretched
1. View source to confirm that the inline size attribute is not filtered for the logo.
1. In the archive, view a post with a featured image.
1. View source to confirm that the filter is still applied to the featured image.

## Quality assurance

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -411,6 +411,11 @@ add_filter( 'the_password_form', 'twenty_twenty_one_password_form' );
  * @return array
  */
 function twenty_twenty_one_get_attachment_image_attributes( $attr, $attachment, $size ) {
+
+	if ( 'custom-logo' === $attr['class'] ) {
+		return $attr;
+	}
+
 	$width  = false;
 	$height = false;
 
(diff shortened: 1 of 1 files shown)


inc/template-functions.php
```diff
@@ -411,6 +411,11 @@ function twenty_twenty_one_password_form( $post = 0 ) {
  * @return array
  */
 function twenty_twenty_one_get_attachment_image_attributes( $attr, $attachment, $size ) {
+
+	if ( 'custom-logo' === $attr['class'] ) {
```
aristath: There may be other filters from plugins etc, adding more classes. I think it would be safer to check if `custom-logo` is _contained_ in the classes instead of it being the only class.

```suggestion
	if ( isset( $attr['class'] ) && false !== strpos( $attr['class'], 'custom-logo' ) ) {
```

aristath (changes requested):

PR #777, opened 2020-11-02
Try using "bolder" for bold text inside of blockquotes.

### LINKED ISSUE #762: Quote block: bold looks less bold than non-bold
I don't think this is an editor problem, but the default style, because it works correctly in my theme.
I made a word in a quote block bold, but that word looks normal and the rest looks bold.

**To Reproduce**
1. Go to the block editor.
2. Add a quote block and type some text.
3. Select a word and click on the Bold icon to make it bold.
4. See that it looks non-bold and the rest of the text looks bold.

**Expected behavior**
The bold text should look bolder than the non-bold text.

**Screenshots**
Front end with one bold word in two quote blocks.
Block Quote – Local Current WP-31

Editor, showing the B indicator is on when on the bold word.
bold-quotes

- WordPress version: 5.5.3
- no Gutenberg plugin

 - OS: Ubuntu Studio 20.04
 - Browser Firefox 82.0

kjellr:
Fixes #762 

Removes the reverse-bold treatment for `<strong>` text inside of blockquotes, instead relying on `bolder`. 

Could use some testing on various operating systems — if there is no system font equivalent that's stronger than a `700` weight, the browser should either fake it, or it'll appear the same. Will work just fine on MacOS, Windows, and Android, where an Extrabold/Heavy/Black system font variant should be provided. 

## Screenshots

Before|After
---|---
Screen Shot 2020-11-02 at 3 12 58 PM|Screen Shot 2020-11-02 at 3 06 34 PM


carolinan (approved):
Tested, works well for me on windows.

aristath (approved):

PR #778, opened 2020-11-02
Fix mobile menu button placement

kjellr:
The "Menu" and "Close" buttons on mobile are awkwardly spaced right now when the button outlines are not shown: 

Screen Shot 2020-11-02 at 3 50 06 PM|Screen Shot 2020-11-02 at 3 49 56 PM
---|---

There are two ways to solve for this. We'll want to trim the left and right button padding, and then: 

1.|2.
---|---
Align based on the entire button's size.|Align the button content visually.
Screen Shot 2020-11-02 at 3 46 04 PM|Screen Shot 2020-11-02 at 3 44 52 PM
Screen Shot 2020-11-02 at 3 46 14 PM|Screen Shot 2020-11-02 at 3 45 00 PM

This PR currently takes the second approach, since when things are not focused (which is most of the time) I think the tighter right margin feels more similar to the space around the site title on the left. Open to more design feedback on that though.

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -369,11 +369,11 @@ a:hover {
 }
 
 .wp-block-button__link:before {
-	margin-bottom: -calc(1em - 0);
+	margin-bottom: -calc(1em + 0);
 }
 
 .wp-block-button__link:after {
-	margin-top: -calc(1em - 0);
+	margin-top: -calc(1em + 0);
 }
 
 .wp-block-button__link:focus {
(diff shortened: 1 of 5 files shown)


kellychoffman:
Let's align the button content visually. We should prioritize for the default state. Meaning, people's site should look good (spacing consistent, etc) 99% of the time, when there is no outline present.

kjellr:
Sounds good, thank you! That's what I was leaning towards too. 🙌

carolinan (approved):
Looks good to me

PR #780, opened 2020-11-03
Take the post-type label into account on post-navigation links

### LINKED ISSUE #761: navigation links for custom post types
I think the theme should provide navigation if its template is loaded. It shouldn't say post though. Using 'Next' and 'Previous' would work fine.

_Originally posted by @joyously in https://github.com/WordPress/twentytwentyone/issues/566#issuecomment-710735196_

The user shouldn't have to have a custom template to show custom post types.
I tested removing the `if ( is_singular( 'post' ) ) ` from single.php and it works fine on posts, pages, attachments, and custom post types. The only thing needed is to remove the word Post from the Next and Previous texts.

aristath:
Fixes #761 

## Summary
Uses the post-type singular label if it exists. If a label does not exist then it will fallback to the WordPress defaults and use `Post` for non-hierarchical post-types, and `Page` for hierarchical post-types.

## Quality assurance

### DIFF
--- a/single.php
+++ b/single.php
@@ -36,10 +36,27 @@ while ( have_posts() ) :
 		// Previous/next post navigation.
 		$twentytwentyone_next = is_rtl() ? twenty_twenty_one_get_icon_svg( 'ui', 'arrow_left' ) : twenty_twenty_one_get_icon_svg( 'ui', 'arrow_right' );
 		$twentytwentyone_prev = is_rtl() ? twenty_twenty_one_get_icon_svg( 'ui', 'arrow_right' ) : twenty_twenty_one_get_icon_svg( 'ui', 'arrow_left' );
+
+		$twentytwentyone_post_type      = get_post_type_object( get_post_type() );
+		$twentytwentyone_post_type_name = '';
+		if (
+			is_object( $twentytwentyone_post_type ) &&
+			property_exists( $twentytwentyone_post_type, 'labels' ) &&
+			is_object( $twentytwentyone_post_type->labels ) &&
+			property_exists( $twentytwentyone_post_type->labels, 'singular_name' )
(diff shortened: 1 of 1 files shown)


carolinan:
Do you have a sample code for a post type to test this with?

carolinan:
I was able to make it work by removing the post type from `if ( is_singular( 'post' ) ) {`

I used this example post type:
https://developer.wordpress.org/reference/functions/register_post_type/#comment-3847

If there is no `singular_name`, it falls back to `name` and if name does not exist then it shows _Next Post_ and _Previous Post_
is that correct? It does not match the description above.

aristath:
> I was able to make it work by removing the post type from `if ( is_singular( 'post' ) ) {`

Ah right, I missed that in my commit :+1: 

> If there is no `singular_name`, it falls back to `name` and if name does not exist then it shows _Next Post_ and _Previous Post_
is that correct? It does not match the description above.

Correct. I updated the description of this PR. Thank you for the note, I had to track down what happens. :+1: 
It will fallback to the WordPress defaults and use `Post` for non-hierarchical post-types, and `Page` for hierarchical post-types.
The reason for that is because when someone registers a new post-type, they use the [`register_post_type`](https://developer.wordpress.org/reference/functions/register_post_type/) function, which creates a new `WP_Post_Type` object, which in turn calls the [`get_post_type_labels`](https://developer.wordpress.org/reference/functions/get_post_type_labels/) function and in there we have the default labels.
It will however fallback to an empty string (Next/Previous without "Post" etc) if there is an error when getting the post-type label. A post-type without labels though is extremely unlikely... And all the extra checks in there make sure there won't be an error thrown even if the post-type is registered completely incorrectly. :laughing:

PR #782, opened 2020-11-03
Replace node-sass with sass (dart-sass)

### LINKED ISSUE #746: Replace node-sass
node-sass relies on LibSass which is deprecated: https://sass-lang.com/blog/libsass-is-deprecated

The migration process is described on the link above as:

`If you’re a user of Node Sass, migrating to Dart Sass is straightforward: just replace node-sass in your package.json file with sass. Both packages expose the same JavaScript API.`

https://www.npmjs.com/package/sass

aristath:
Fixes #746 

## Summary
* Replaced the package
* Fixed the CLI commands, the new package uses different arguments
* Fixed a compiling error due to wrong selector
* re-compiled styles with the new package

The only difference is the indentation, there is no parameter I could find in dart-sass to select tabs instead of spaces for the generated file. But that's fine there's no difference.

## Quality assurance


carolinan:
This is peculiar because we do need to replace node-sass in order to future proof the theme.

WordPress coding standards does ask us to use tabs, are you sure this is going to pass all tests when the theme is added to core?

aristath:
It will pass now... I added an extra step and the CSS files will go through a CSS check/fix. So the tabs were fixes along with some other indentation & whitespace issues.

carolinan:
Tested to the best of my ability, no issues found.

Issue #536, opened 2020-10-16
Blocks vertical margins


briceduclos:
Currently, all blocks inside .entry-content have a top margin and a bottom margin.
```
.site-main > article > *, .site-main > .not-found > *, .entry-content > *, [class*="inner-container"] > *, .wp-block-template-part > * {
    margin-top: var(--global--spacing-vertical);
    margin-bottom: var(--global--spacing-vertical);
}
```

Is there a reason to add a top margin? (Previous default themes set the top margin to 0). I’m asking because when the user inserts a small spacer block (e.g 20px, the minimal value) between two blocks, the space is very large (20px + 2* spacing-vertical = 80px).

kjellr:
I don't have a strong opinion on this one. Twenty Nineteen used top and bottom margins. Twenty Twenty did not. 

There are a handful of Gutenberg blocks that don't ship bottom margins by default on the front end (like the Group block for example). This causes issues when a theme doesn't account for that. In fact, while looking into this, I just noticed that being an issue on the Twenty Twenty demo: 😅 

![Screen Shot 2020-10-16 at 9 15 18 AM](https://user-images.githubusercontent.com/1202812/96262675-1eb7c780-0f90-11eb-8c87-4e19becacd17.png)

For Twenty Twenty-One this point, I'd suggest taking a lighter touch — since we're so close to beta I'd prefer not to rock the boat with the global margins too much. But if someone would like to try it out and see, I'd be happy to take a look. 

Either way, I do think it'd be a good idea to remove the margins from the spacer block.

aristath:
> Either way, I do think it'd be a good idea to remove the margins from the spacer block.

I agree for the spacer block. Not for all others, but an exception for spacers should be made :+1:

pattonwebz:
Yup seems like a great call to make to remove margins from the spacer at least :)

kjellr:
I'll open a quick PR to handle that.

kjellr:
Actually — we do that already. 😂 

https://github.com/WordPress/twentytwentyone/blob/e0486a21ce723aee3c99d4bdca1331abd31134e6/assets/sass/05-blocks/spacer/_style.scss#L4-L5

pattonwebz:
Nice, seems someone already considerd this before us. That person was smart :)

briceduclos:
[...]

PR #783, opened 2020-11-03
Update vertical margins for full width blocks

carolinan:
Related to https://github.com/WordPress/twentytwentyone/issues/536
And discussed on WordPress Slack https://wordpress.slack.com/archives/C02RP4VMP/p1604327710441600 (Account needed).

## Summary

1. Removes an editor style that removed the vertical margin from _all_ full width blocks.
1. Removes the vertical margin from the full width cover block and full width media & text block
in the editor and the front.

This way we can place two full width cover blocks or media & text blocks after each other without spacing, creating layouts that is not possible with spacing in between them.

(I also tested it with the columns block, but the result was not good.)

## Test instructions

1. Create a cover block and set it to full width. Duplicate the block.
1. Create a media and text block and set it to full width. Duplicate the block.
1. View the spacing in the editor.
1. Save and view the spacing on the front.
1. If you like, create a full width group block with a background color, duplicate it and compare. 
1. Imagine all the ways this can be combined to create layouts.

## Screenshots

vertical-front

## Quality assurance


kjellr (approved):
This looks great in my testing. Would it make sense to also implement this for the ~Cover~ Group block in this PR? In my testing, that block has margins in the editor, but not in the front end so far: 

Editor|Front End
---|---
Screen Shot 2020-11-04 at 11 39 52 AM|Screen Shot 2020-11-04 at 11 39 42 AM

If you want to leave that to a separate PR that's fine too — this one is good to go as is.

carolinan:
It was supposed to be fixed, not sure how I missed that 🙃

Ah, maybe I only tested a cover block with a background image.

kjellr:
Oh whoops, sorry — I wrote "Cover" above, but I meant "Group" block. 😅 

The Cover block is working great! I was wondering if we want to add the Group block to this PR too. No worries if not, that can be handled separately.

carolinan:
> Oh whoops, sorry — I wrote "Cover" above, but I meant "Group" block. 😅
> 
> The Cover block is working great! I was wondering if we want to add the Group block to this PR too. No worries if not, that can be handled separately.

Yes makes sense when the group has a background color.

PR #791, opened 2020-11-05
Improve the dark mode option message in the customizer

### LINKED ISSUE #790: Improve the dark mode instructions in the customizer
**Describe the bug**
The current instructions for the dark mode is the customizer:
- Label:  Respect visitor's device dark mode settings 
- Instructions: Dark mode is a device setting. If a visitor to your site requests it, your site will be shown with a dark background and light text.

These could be improved to be more clear to the users.

**To Reproduce**
1. Go to the customizer
2. Click on the "Colors" tab
3. See the instructions for the dark mode option

JustinyAhin:
Proposal to fix #790.

## Summary
Improve the dark mode option message in the customizer.

No

## Test instructions
1. Go to the customizer
2. Click on the "Colors" tab
3. Read the new instructions for the dark mode option

## Quality assurance


classes/class-twenty-twenty-one-dark-mode.php
```diff
@@ -164,8 +164,8 @@ public function customizer_controls( $wp_customize ) {
 			array(
 				'type'            => 'checkbox',
 				'section'         => 'colors',
-				'label'           => esc_html__( 'Respect visitor\'s device dark mode settings', 'twentytwentyone' ),
-				'description'     => __( 'Dark mode is a device setting. If a visitor to your site requests it, your site will be shown with a dark background and light text.', 'twentytwentyone' ),
+				'label'           => esc_html__( 'Dark Mode', 'twentytwentyone' ),
```
aristath: Clicking the checkbox doesn't enable dark-mode... Dark mode will only be automatically enabled if the visitor is using dark-mode in their OS settings.
So we can't call this setting "dark mode", it really is about respecting the visitor's OS settings, and allowing them to switch if they want.

JustinyAhin: My understanding and what I saw is that when clicking the checkbox, the dark mode button appears, and user can click to enable it.
Is that right?

carolinan: It does both and that is part of the difficulty

aristath: Yes. But if the checkbox is labeled "Dark Mode", it gives the impression that checking the checkbox will enable dark-mode, which is not what happens :smile:

carolinan:
I did a google image search for "enable dark mode" to get some suggestions. 
"Enable dark theme" is most commonly used. These settings have no further explanations to what this means. 
I think we are overcomplicating it.

aristath:
`Enable dark theme` would be appropriate if that's what the setting does. But that's not what it does... It will only enable dark mode if the OS is set to dark-mode. If not, it will allow users to enable dark-mode if they so desire - but dark mode will not be enabled when they click the checkbox. So if we change the checkbox's label to `Enable dark mode`, then enabling the checkbox looks broken because it won't enable dark mode.
The current label is verbose and long, but what exactly is the problem with it? Why is `dark mode` or `enable dark mode` preferable? Is it simply a matter of length?

JustinyAhin:
I think it's more a matter of understanding than anything else.

I agree with you Ari on the point that putting "Enable dark theme" doesn't reflect what the checkbox does.

I personally didn't understood it until you fully explain it now.

Let's keep the discussion open to have more feedbacks.

I'll ping Mary Baum on Slack later today (she's in US timezone). She may have proposals to rephrase all this.

carolinan:
Label: Dark Mode,  or, Dark Mode support
(bonus if we can make this bold)

Description:
Respect visitor's device Dark Mode settings.
Dark mode is a device setting. If a visitor to your site
requests it, your site will be shown with a dark 
background and light text.

Visitors can also turn Dark Mode on and off with a button.
 (link to support page)


----
_Visitors can also_
... Not only visitors can use the button, it is also for the site owner and content creator 🤔

carolinan:
As a reminder this text must be set by November 10. After that we can't update it.

classes/class-twenty-twenty-one-dark-mode.php
carolinan:
```suggestion
				'label'           => esc_html__( 'Dark Mode Support', 'twentytwentyone' ),
```

classes/class-twenty-twenty-one-dark-mode.php
```diff
@@ -164,8 +164,8 @@ public function customizer_controls( $wp_customize ) {
 			array(
 				'type'            => 'checkbox',
 				'section'         => 'colors',
-				'label'           => esc_html__( 'Respect visitor\'s device dark mode settings', 'twentytwentyone' ),
-				'description'     => __( 'Dark mode is a device setting. If a visitor to your site requests it, your site will be shown with a dark background and light text.', 'twentytwentyone' ),
+				'label'           => esc_html__( 'Dark Mode Support', 'twentytwentyone' ),
+				'description'     => __( 'By enabling this option, visitors to your site will be able to switch to Dark Mode. In this mode, the background color of the site will be dark, with light text.', 'twentytwentyone' ),
```
carolinan:
```suggestion
				'description'     => __( 'Respect visitor\'s device dark mode settings. <br>Dark mode is a device setting. If a visitor to your site requests it, your site will be shown with a dark background and light text.<br><br>Dark Mode can also be turned on and off with a button.<br><br><a href="
https://wordpress.org/support/article/twenty-twenty-one/">Learn more about Dark Mode.</a>', 'twentytwentyone' ),
```

aristath (approved):

carolinan:
Do we need to remove the HTML from the string in the description?

PR #792, opened 2020-11-05
Update text strings

carolinan:
# Summary

This is part of the text string review for the Beta 4 string freeze.

Reduces some text strings for the starter content and block patterns that were very similar.
Adds some additional translator contexts.
Renames the menu locations to improve consistency.

## Quality assurance


template-parts/content/content-none.php
```diff
@@ -27,7 +27,7 @@
 
 		<?php else : ?>
 
-			<h1 class="page-title"><?php esc_html_e( 'Nothing Found', 'twentytwentyone' ); ?></h1>
+			<h1 class="page-title"><?php esc_html_e( 'Nothing here', 'twentytwentyone' ); ?></h1>
```
aristath, line 30: Should this be `Nothing Here`? (question, not suggestion)

carolinan: I don't think so https://make.wordpress.org/docs/handbook/documentation-team-handbook/handbooks-grammar-guide/#capitalization

aristath (approved):

PR #794, opened 2020-11-05
Fix inconsistent use of classes for dark/light mode between editor & front

aristath:
The editor already has a `.is-dark-theme` class. This PR updates all other implementations to use the same class. Ensures consistent styling for dark-mode and manually selected dark background-colors.

## Quality assurance


carolinan:
We need to do this change to ensure that the colors work with dark mode on, and dark mode off but with a dark background. 
There are some side effects and it will require testing, but they can be fixed in another pull request.

Issue #784, opened 2020-10-30
Dark Mode Toggle Placement


kjellr:
I'd like revisit discussion about the toggle placement from WordPress/twentytwentyone-dark-mode#1. Currently, the dark mode toggle is placed here: 

![image](https://user-images.githubusercontent.com/1202812/97740888-4478bc80-1ab8-11eb-896b-da98039f3ec9.png)

I have a couple concerns with its current placement: 

## Prominence of the Button

This is a very strong visual element, and it is located above all other site content. Unless there's a prominent site logo, this button will likely be the first thing a user's eye is drawn to in the header. From a hierarchy perspective, I don't think that makes much sense: while we want to make this toggle feature available, it shouldn't be more important visually than the site title. 

This is especially an issue on mobile, where it takes up a larger proportion of the screen: 

![Screen Shot 2020-10-30 at 2 25 05 PM](https://user-images.githubusercontent.com/1202812/97743216-b9012a80-1abb-11eb-95ad-9a7e2c355cb7.png)

Does this need to exist in the header? I do not have data around this, but I'd guess that most users will not make use of this button at all. Does this need to be in the main header if that's the case?

## Position of the Button

Assuming this button does remain in the header, the current placement feels imbalanced. The header forms a weird "N" shape: 

![image](https://user-images.githubusercontent.com/1202812/97741977-0381a780-1aba-11eb-9e5c-35e1eec3ba2f.png)

If we do place the button in the header, I'd suggest we work to balance this shape out visually. Even something like this helps: 

![image](https://user-images.githubusercontent.com/1202812/97742094-362ba000-1aba-11eb-8eca-881ddbcbec89.png)

Moving it below the menu there would also improve the mobile situation, allowing the close button to once again appear in the same position as the open menu button. Currently, [they are in different spots](https://cloudup.com/cp1h9lIA1qE). 

It would be even more balanced to do something like this (without placing it in the actual `nav` element of course): 

![image2](https://user-images.githubusercontent.com/1202812/97742165-4a6f9d00-1aba-11eb-816e-6de12b4fae9a.png)

Are there concerns with this approach? Does anyone have any other ideas that I've missed?

carolinan:
I still have accessibility concerns for the placement of the button in the right corner, because focus moves from the upper left skip link to the bottom right. This causes problems if you are zooming in or using a magnifier.

[...]

PR #798, opened 2020-11-06
Hide dark-mode toggler on scroll

aristath:
Hides the dark-mode toggler on scroll-down, reveals it on scroll-up.
See https://github.com/WordPress/twentytwentyone/issues/784#issuecomment-722364092

## Quality assurance

### DIFF
--- a/assets/css/style-dark-mode.css
+++ b/assets/css/style-dark-mode.css
@@ -55,6 +55,11 @@
 	position: fixed;
 	bottom: 5px;
 	right: 5px;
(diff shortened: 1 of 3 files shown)


aristath:
The failing test has nothing to do with this PR, apparently `npm install` fails. Tested & linted on localhost, there are no issues.

carolinan:
Tested on desktop with up to 200 and 400 % zoom and on mobile, and this update makes it easier to read the content.

PR #799, opened 2020-11-06
Hide the toggler while in the editor.

aristath:
This is for a suggestion in https://github.com/WordPress/twentytwentyone/issues/784#issuecomment-722367594
Hides the toggler in the editor and simply respects the user's preferences from the frontend.

## Quality assurance

### DIFF
--- a/assets/css/style-dark-mode.css
+++ b/assets/css/style-dark-mode.css
@@ -46,9 +46,7 @@
 }
 
 .wp-admin #dark-mode-toggler {
-	z-index: 99999;
-
-	/* Necessary for the editor. */
+	display: none;
 }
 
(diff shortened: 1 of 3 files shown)


aristath:
The failing test has nothing to do with this PR, apparently `npm install` fails. Tested & linted on localhost, there are no issues.

carolinan:
> The failing test has nothing to do with this PR, apparently `npm install` fails. Tested & linted on localhost, there are no issues.

It's failing on all PR's but I don't know how to resolve it

aristath:
I don't think there's something we can do to solve it... I updated the package-lock file but that didn't fix it. Looks like this is an issue with GitHub right now and it's unable to get one of the sub-packages required.

PR #800, opened 2020-11-06
don't enlarge images

### LINKED ISSUE #563: Featured image is assumed to be large
When using a small featured image, the CSS uses 
`min-width: var(--responsive--aligndefault-width)` which makes the image blurry.
The image in this test is 165px wide. 

blurred-featured-image

Without that one rule, it looks like it should:
normal-featured-image

aristath:
Fixes #563

If the image is small this PR doesn't force it to be the width of its container and instead it keeps its original size.

Screenshot_2020-11-06 My Blog – Just another WordPress site

Screenshot_2020-11-06 My Blog – Page 3 – Just another WordPress site

## Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -5918,23 +5918,10 @@ body:not(.single) .site-main > article:last-of-type .entry-footer {
 .post-thumbnail .wp-post-image {
 	display: block;
 	width: auto;
-	min-width: calc(100vw - 30px);
 	max-width: 100%;
 	margin-left: auto;
 	margin-right: auto;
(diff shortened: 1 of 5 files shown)


carolinan:
Tested, works well for me with new images and with the theme unit test. But this is a design decision.

carolinan (approved):

aristath:
> Tested, works well for me with new images and with the theme unit test. But this is a design decision.

Agreed. Just getting things ready for when the decision comes :+1:

melchoyce (approved):

Issue #804, opened 2020-11-06
Allow outlined buttons to have custom background colors.


carolinan:
**Describe the bug**
The outlined button has a transparent background, even if the user selects a background color.

**To Reproduce**
1. Add a button block. 
2. Set the style to outline.
3. Add a background color.

**Expected behavior**
When the user selects a background color the button should use that color.

carolinan:
Fixed with #808, closing

PR #808, opened 2020-11-07
Restore the default colors for the button blocks hover and focus styles

### LINKED ISSUE #707: Try another hover style for buttons with custom colors
From @melchoyce  on #637 
_This is another place where Gutenberg makes it harder on theme authors to handle interactions. The focus styles (dotted outline) on hover feels weird in action, but I'm not sure how else to approach this. Could we use the same hover colors as the default buttons, regardless of what custom colors are chosen?_

97234701-c5ce0780-17b7-11eb-9d07-c99ede8d610b

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/707, https://github.com/WordPress/twentytwentyone/issues/804

## Summary

Previously, an issue was opened about the buttons not respecting the users custom colors on hover and focus.
This led to a series of attempts over several weeks to make the hover and focus styles use the custom colors.
-This was made difficult by the inline color styles, the text colors that are used together with the palette background colors,
and the lack of a way to inherit a background color.

This PR

1. Restores the buttons default colors on hover and focus.
1. Allows custom background colors for outlined buttons, **in the buttons default state**
1. Restores the original focus and hover styles.
1. Removes styles in the editor, that was previously needed to make sure that the button text is always visible while editing and typing.
1. Reduces the complexity and the amount of CSS needed.

In order to override the custom colors, `!important` was needed in a couple of places.

## Test instructions

1. Add buttons, one for each style and color variation. With color variation I mean text and background, not one of each palette.
1. View the buttons and test their default and hover style in the editor
1. View the buttons and test their default, hover and focus styles on the front.

## Screenshots

Original design in Figma:
buttons-style-guide

This PR:
Filled, hover:
filled-hover

Filled focus:
filled-focus

Outline hover:
outline-hover

Outline focus:
outline-focus

## Quality assurance


kjellr:
Thanks for giving this a try. I'm not a huge fan of using `!important` so many times here, though I understand the necessity. 

I do see a couple issues so far: 

1. Filled buttons with a custom (not preset) background color appear with invisible text on hover. This happens in both the front end and the editor: 

filled

2. Outlined buttons with a custom (not preset) background color and custom (not preset) text color appear incorrectly on the front end. They appear as expected in the editor though: 

Editor|Front End
---|---
outline|outline

carolinan:
@kjellr thank you for testing, <del>I am not able to reproduce this, did you perhaps choose not to add a link to the button text?</del>

Answered the question myself after removing the links.
I am only able to reproduce this for the outlined button that has both text color and background.

carolinan:
Test content: https://gist.github.com/carolinan/2952bb4ba1a66beace648a7362d185d3

kjellr:
The issues from above appear to be fixed now. 👍 

I am seeing some weirdness around the default (non-hover) states when using the preset colors though. Here's an example: 

Editor|Front End
---|---
Screen Shot 2020-11-09 at 2 51 44 PM|Screen Shot 2020-11-09 at 2 51 52 PM

In the first set, the first three buttons should have white text in the front end. And in the second set, all of the buttons look identical in the front end even though they're supposed to have custom text colors. 

Here's the sample content I used: https://gist.github.com/kjellr/1bb3687e6d9d9b64a0cb6ef9f25f736e

carolinan:
I am missing context, why are they supposed to have white text? Because of the utilities style?

kjellr:
When a user choose one of the theme's 3 dark preset colors, it usually changes the text to white automatically to make sure it's readable. This happens for buttons, but also for any block with a background color (paragraph, group, etc). 

So the editor is showing the intended behavior in my screenshots above. The front-end just doesn't match up.

carolinan:
Editor|Front End
---|---
dark-editor|dark-front

Focus, dark mode:
button-focus

Hover, default background:
button-hover

carolinan:
I am going to merge this so that more people can test and so that I can continue on the colors when the button is placed inside a block with a background color.

PR #812, opened 2020-11-09
Fix dark background-colors

### LINKED ISSUE #805: Dark mode: selecting a dark background-color in the customizer should turn off the feature completely
Currently dark colors are not applied because dark-mode is forced on (CSS classes conflict)

aristath:
Fixes #805

## Quality assurance


carolinan:
Looks good to me, tested with different browsers. 
There are some delays when I test the option in IE11 but that is expected.

PR #817, opened 2020-11-09
Load dark-mode button in the footer

aristath:
See #784 

## Quality assurance


carolinan:
Works well and I am confident that with the choices we have, this is the better, least disruptive solution.

PR #819, opened 2020-11-09
Show dark-mode toggler when scrolling to the very bottom of the page

aristath:
Came up in Friday's test-scrub.
When the visitor scrolls-down to the very bottom of the page, the toggler should appear and not require a scroll-up.

## Quality assurance


carolinan:
Tested on desktop and mobile without issues.

Issue #833, opened 2020-11-11
Inform user when Dark Mode feature unavailable because site background color is dark.


adamsilverstein:
**Describe the bug**
When testing the Dark Mode feature I noticed that if I set my theme's background color to a dark color, the option to enable the dark mode feature disappears:
 
While this makes sense functionally, I found it confusing as a user. I wanted to enable dark mode, after I set the background color. When I chose a dark color, then discovered the Dark Mode option was gone, I was confused.

**To Reproduce**
1. Test a dark background color.

**Expected behavior**
Perhaps when the background is too dark we can grey out/disable the Dark Mode feature? Or perhaps display a notice to the user "Choose a lighter background color to enable the Dark Mode feature"?

**Screenshots**
[Screencast](https://user-images.githubusercontent.com/2676022/98832250-d4dcc880-23f9-11eb-961b-ff5c18344c5c.gif)

**Editor version (please complete the following information):**
- WordPress version: [e.g: 5.5.1]
- Does the website has Gutenberg plugin installed, or is it using the block editor that comes by default? [e.g: "gutenberg plugin", "default"]
- If the Gutenberg plugin is installed, which version is it? [e.g., 8.9]

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

melchoyce:
FWIW, I think it would be more straightforward to remove this conditional and not worry about someone's chosen background color.

carolinan:
There is already a description above the option that says that you need to select a light color to access the Dark Mode settings.

I am OK with the solution Mel suggested if it can be done easilly.

carolinan:
@adamsilverstein I see the confusion now that I am more awake (good morning) The screencast is using an older version of the theme. 
Please test the current trunk version from this GitHub repository, or see https://make.wordpress.org/core/2020/11/10/twenty-twenty-one-dark-mode-update/

This is what the customizer option looks like currently:
![TT1-customizer-2](https://user-images.githubusercontent.com/7422055/98910665-c529b680-24c3-11eb-909f-000eded490ef.png)

aristath:
> I think it would be more straightforward to remove this conditional and not worry about someone's chosen background color.

Hmmm then what happens if I select a dark background-color and then enable the dak-mode checkbox? If I make a change to a control I expect to see a change in the preview, and in this case nothing would happen (assuming of course the selected background-color is dark)
[...]

PR #838, opened 2020-11-12
Change implementation for customizer conditions

aristath:
See https://github.com/WordPress/twentytwentyone/issues/833

This PR changes the behavior of control dependencies.
Now the checkbox is always shown, BUT if the background color is dark then the checkbox is disabled.

demo31

## Quality assurance

### DIFF
--- a/assets/js/customize.js
+++ b/assets/js/customize.js
@@ -3,18 +3,22 @@
 ( function() {
 	// Wait until the customizer has finished loading.
 	wp.customize.bind( 'ready', function() {
-		// Hide the "respect_user_color_preference" setting if the background-color is dark.
+		// Get the checkbox element.
+		var input = wp.customize.control( 'respect_user_color_preference' ).container.find( 'input' )[0];
+		// Disable the "respect_user_color_preference" checkbox if the background-color is dark.
 		if ( 127 > twentytwentyoneGetHexLum( wp.customize( 'background_color' ).get() ) ) {
-			wp.customize.control( 'respect_user_color_preference' ).deactivate();
+			input.setAttribute( 'disabled', 'disabled' );
 		}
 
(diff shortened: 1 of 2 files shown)


carolinan (approved):
Looks good to me, I think this will be less confusing.

melchoyce:
Honestly, this feels more complicated to me. Could we just move the "To access Dark Mode..." message below the color picker, and only show it when someone selects a dark background color?

image

carolinan:
@melchoyce I did not understand, can you describe again what shows when?

The feedback from the testing scrubs said that the temporary notice was confusing.

melchoyce:
Default / Light Color Selected | Dark Color Selected
----- | -----
image | image

carolinan:
The test scrub showed that people found the temporary notice and hiding of the checkbox confusing, that is why the previous change was made.

melchoyce:
Yeah, I'm wondering if part of the problem is that the "to access dark mode..." message is hard to spot. I think putting the message where the checkbox normally is will help, but what if we also wrapped the message in a notice and only showed it when someone has selected a dark background color? 

image

carolinan:
I think it is a bad idea to hide the support link, as in, only show it when the option already is disabled. 🤔 
(Also, the customizer has its specific way to show notices which is above the option not below.)

melchoyce:
Can we add the support link to the default description as well? 

> Respect visitor’s device dark mode settings.
> 
> Dark mode is a device setting. If a visitor to your site requests it, your site will be shown with a dark background and light text. [Learn more about Dark Mode](#).
> 
> Dark Mode can also be turned on and off with a button that you can find in the bottom right corner of the page.

aristath:
Closing in favor of #852

PR #839, opened 2020-11-12
Update the default pullquote color to gray

### LINKED ISSUE #837: Pull quote has the wrong default color on the front
**Describe the bug**
On the front, the pull quote block is inheriting the color `#555` from the block files.
In the editor, the color is the dark grey  `#28303D`.

The _solid style_ pull quote has the correct color.

**To Reproduce**
1. Add a pull quote block.
2. Compare the colors in the editor and front.

**Expected behavior**
I expect the color in the editor and front to match.

**Screenshot**

wrong-color

JustinyAhin:
Fixes #837.

## Summary
Update the default pullquote color to gray. What is done is to add `color: var(--global--color-gray)` to the `wp-block-pullquote` class both in frontend and in editor files.

No

## Test instructions

1. Add a pull quote block.
2. The color in the editor and in front should be gray (#39414d)
3. This color should also change in front and in editor when it's updated in the block settings

## Quality assurance

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -1780,7 +1780,7 @@ pre.wp-block-preformatted {
 	border-width: 3px;
 	border-bottom-style: solid;
 	border-top-style: solid;
-	color: currentColor;
+	color: #39414d;
 	border-color: currentColor;
 	position: relative;
 }
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -1780,7 +1780,7 @@ pre.wp-block-preformatted {
 	border-width: 3px;
 	border-bottom-style: solid;
 	border-top-style: solid;
-	color: #39414d;
+	color: currentColor;
 	border-color: currentColor;
 	position: relative;
 }
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -4087,6 +4087,7 @@ pre.wp-block-preformatted {
 }
 
 .wp-block-pullquote {
+	color: #39414d;
 	padding: 40px 0;
 	text-align: center;
 	border-width: 3px;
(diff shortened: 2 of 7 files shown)


carolinan:
I meant that when the body background is changed to dark, the block needs to use the variables that changes to white:

darkmode-pq

I also wonder if this color is a new change to Gutenberg / 5.6 🤔

ryelle:
> I also wonder if this color is a new change to Gutenberg / 5.6 🤔

It is, I believe it will be fixed by https://github.com/WordPress/gutenberg/pull/26684, but I haven't been following which fixes will make it into 5.6. It should already be fixed if you've got gutenberg master installed.

kjellr:
https://github.com/WordPress/gutenberg/pull/26684 is not slated to make it into 5.6, and I'm not sure it really qualifies as something that should be backported, since it wasn't actually meant to be a fix (cc @jasmussen for thoughts on that). 

If that's the case, I see no harm in fixing on the theme end today, so that this works at launch.

jasmussen:
That will be fixed by https://github.com/WordPress/gutenberg/pull/26684, correct, but I also don't know if that lands. 

If your CSS can carry it, I'd recommend just duplicating some of the updated styles that change the text and border to `currentColor` and adding them to the theme.

ryelle:
I've pushed a commit to use `currentColor` for text & borders, which fixes the issue:

Screen Shot 2020-11-20 at 10 57 52 AM

Screen Shot 2020-11-20 at 10 57 47 AM

I also tested with custom colors, and it still works as expected.

Screen Shot 2020-11-20 at 10 59 06 AM

kjellr (approved):
Works well in my testing. Thanks for the fix!

PR #847, opened 2020-11-12
Actions: Intentionally set the node version on lint check

ryelle:
This should prevent the linter check from erroring out. I'm sure we're moving off github soon, so to keep things functioning this just sets the node version to 12.x, rather than fixing whatever the issue is with node 14.x.


aristath (approved):
Thanks! LGTM :+1:

PR #846, opened 2020-11-12
Dark Mode: Show toggle button in the editor

ryelle:
## Summary
There's code to inject the Dark Mode toggle into the editor, but it's not currently showing up (or working correctly). This PR should fix the display, z-indexing, and JS to let authors cycle the dark mode on and off while writing.

## Relevant technical choices:
The toggle code in `dark-mode-toggler.js` now checks for `twentytwentyoneDarkModeEditorToggleEditorStyles` & runs that, to make sure we toggle the `is-dark-mode` class from `.editor__typewriter`. It's not the most straightforward method, but it lets us avoid adding a lot of wp-admin code into `dark-mode-toggler.js` or duplicating code into separate toggle functions.

## Test instructions

This PR can be tested by following these steps:
1. Edit a post/page
2. See the Dark Mode button in the bottom right corner
3. It should not be obstructed by any other element
4. Clicking it should toggle dark mode
5. Clicking it again should toggle dark mode again

Check that the frontend toggle still works too.

## Screenshots

Before, note no visible dark mode button:

before-no-sidebar
before-sidebar

After, button is visible with sidebar open or closed:

after-light-sidebar
after-dark-sidebar
after-light-no-sidebar

## Quality assurance
* [x] I have thought about any security implications this code might add.
* [x] I have checked that this code doesn't impact performance (greatly).
* [x] I have tested this code to the best of my abilities
* [x] I have checked that this code does not affect the accessibility negatively

PR #849, opened 2020-11-13
Dark Mode: Remove toggle code from the editor

ryelle:
See #846 — the dark mode toggle is still being injected into the editor, even though it was intentionally removed. This PR removes the code related to injected the toggle button into the editor.

Removes the JS file used in the editor, the ajax action that returns the HTML, and the style output function (which was only used in the ajax action). It also removes the `display: none` since the button doesn't exist anymore.

## Test instructions

1. Open the editor
2. There should be no toggle button
3. Inspect element, look for the div with `editor-styles-wrapper`; the first child item should not be a button
4. Switch back to the frontend
5. That toggle should still work as expected

## Quality assurance


aristath:
Even though the button doesn't technically exist (or at least it's not shown), the HTML & JS still get injected because we need to figure out what the preferred color-scheme for the user is, and modify the classes so that the dark/light mode gets applied to the editor.
We can't just remove these...

ryelle:
> we need to figure out what the preferred color-scheme for the user is, and modify the classes so that the dark/light mode gets applied to the editor.

Aha, fair enough, but we don't need an entire ajax request for that. I brought back the editor script & updated it to use a new `twentytwentyoneIsDarkMode` check. If it's dark mode, it waits until `.block-editor__typewriter` exists, and adds the class.

aristath (approved):
Great! LGTM :+1:

PR #852, opened 2020-11-16
Add custom notice pseudo-control

aristath:
This is an alternative to https://github.com/WordPress/twentytwentyone/pull/838

## Summary
* Adds a notice between the colorpicker and checkbox
* Removes the section description

In WP notices are always above their controls. Since we want to keep the checkbox dependency (we don't want it always shown) we can't add the notice to that control (it will be hidden when the checkbox is also hidden). As a result, a new control-type had to be introduced. This PR adds a new `Twenty_Twenty_One_Customize_Notice_Control` class which just prints a notice with our hardcoded message. For the purposes of this theme and an one-off control there's no point in abstracting the logic.

Peek 2020-11-16 16-59

## Quality assurance

### DIFF
--- a/assets/js/customize.js
+++ b/assets/js/customize.js
@@ -6,6 +6,7 @@
 		// Hide the "respect_user_color_preference" setting if the background-color is dark.
 		if ( 127 > twentytwentyoneGetHexLum( wp.customize( 'background_color' ).get() ) ) {
 			wp.customize.control( 'respect_user_color_preference' ).deactivate();
+			wp.customize.control( 'respect_user_color_preference_notice' ).deactivate();
 		}
 
 		// Handle changes to the background-color.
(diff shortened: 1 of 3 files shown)


aristath:
Discussed some more on Slack, added an active_callback:

Peek 2020-11-16 17-11

aristath:
As requested, a link to the theme docs was added to the checkbox's description:

Screenshot_2020-11-16 Customize My Blog – Just another WordPress site(2)

melchoyce (commented):
Works great! Thanks Ari 🌟 

Noticed the copy formatting looks a little weird at my screen size: 

image

Could we remove the `<br>`s in that first section?

image

aristath:
Done

Screenshot_2020-11-16 Customize My Blog – Just another WordPress site(3)

melchoyce:
LGTM!

PR #855, opened 2020-11-17
Fix the casing of "Dark Mode."

desrosj:
Looks like "Dark Mode" is consistently used throughout the theme. Let's change this instance. This change was included in the latest sync in WordPress `trunk`.

It also feels like the link does not belong within the translation string. I know that support documentation can have localized URLs, but I'm not sure the best practice for this. 

The [internationalization best practices say](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#best-practices-for-writing-strings):
```
- Do not put unnecessary HTML markup into the translated string
- Do not leave URLs for translation, unless they could have a version in another language.
```

Defer to the team's interpretation and experience here (I know there are polyglots here), but wanted to raise this!

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -5653,6 +5653,12 @@ a.custom-logo-link {
 .no-widgets .site-footer {
 	margin-top: 180px;
 }
+@media only screen and (max-width: 481px) {
+
+	.no-widgets .site-footer {
+		margin-top: 30px;
+	}
+}
 
 .site-footer > .site-info {
 	padding-top: 30px;
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -7395,7 +7395,7 @@ h1.page-title {
 @media only screen and (max-width: 481px) {
 
 	.widget-area {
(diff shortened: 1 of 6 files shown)


melchoyce:
+1 to "Dark Mode."

carolinan:
The support pages does come in different languages:

https://fr.wordpress.org/support/article/twenty-twenty/

While the Swedish and german version redirects to the forum:

https://sv.wordpress.org/support/article/twenty-twenty/
redirects to: https://sv.wordpress.org/support/topic/twenty-twenty-blogg/

https://de.wordpress.org/support/article/twenty-twenty/

desrosj:
I looked through Core a bit. It seems that the predominant pattern is to leave the `<a>` tag within the string, extract the URL out into a separate translatable string, and use `sprintf`/`printf` to stitch them together.

I updated to follow this pattern. It's a bit more verbose, but would better match the current patterns in Core.

aristath (approved):

PR #861, opened 2020-11-20
Decrease range input thumb size on Firefox

### LINKED ISSUE #860: Range input thumb has different sizes depending on the browser
**To Reproduce**
1. Create a post and add an HTML block with these contents:
```HTML
<input type="range" min="0" max="100" />
```
2. Save the post and preview the end result in Chrome and Firefox.
3. Notice the thumb is larger in Firefox than in Chrome.

**Expected behavior**
I would expect `<input type="range" />` to render the same independently of the browser.

**Screenshots**
_Firefox:_
imatge

_Chrome:_
imatge

- WordPress version: 5.6-beta4
- Does the website has Gutenberg plugin installed, or is it using the block editor that comes by default? "default"

Aljullu:
Fixes #860.

## Summary
This PR sets `box-sizing: border-box` to the range input thumb on Firefox, so it has the same size as Chrome.

## Test instructions

1. Create a post and add an HTML block with these contents:
```HTML
<input type="range" min="0" max="100" />
```
2. Save the post and preview the end result in Chrome and Firefox.
3. Verify the thumb has the same size in Firefox and Chrome.

## Screenshots
_Chrome:_
imatge

_Firefox (before):_
imatge

_Firefox (after):_
imatge

I also tested Gnome Web (which uses WebKit like Safari), and it looks like Chrome. IE11 seems to be between both sizes, so I left it untouched.

## Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -1978,6 +1978,7 @@ input[type=radio] + label {
 		border-radius: 50%;
 		background: #d1e4dd;
 		cursor: pointer;
+		box-sizing: border-box;
 	}
 }
 
--- a/assets/sass/04-elements/forms.scss
+++ b/assets/sass/04-elements/forms.scss
@@ -230,6 +230,7 @@ input[type="radio"] + label {
 		border-radius: 50%;
 		background: var(--global--color-background);
 		cursor: pointer;
+		box-sizing: border-box;
 	}
 }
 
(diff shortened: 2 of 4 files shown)


carolinan:
This got me thinking, do we have to make the thumb at least 44x44 px? It's currently 25x25px.
Reference https://www.w3.org/WAI/WCAG21/Understanding/target-size.html

pattonwebz:
Good point, I recall in a11y chats that 44x44 was the minimum size suggestion for interaction items.

Issue #865, opened 2020-11-22
Spacing issue inside the editor


briceduclos:
**Describe the bug**
I created a common layout: a full-width group that contains a main heading and two columns with text and image. Inside the editor, the heading and paragraph blocks inside the columns have unwanted side paddings. And the main heading has no margin-bottom (related to https://github.com/WordPress/twentytwentyone/pull/859). 

 
![editor-group-layout](https://user-images.githubusercontent.com/56188815/99906071-2a0bba80-2cd5-11eb-9ee3-8de264ce3b3b.PNG)

**To Reproduce**
Inside the code editor, paste the following code:
```
<!-- wp:group {"align":"full","backgroundColor":"blue"} -->
<div class="wp-block-group alignfull has-blue-background-color has-background"><div class="wp-block-group__inner-container"><!-- wp:heading {"textAlign":"center"} -->
<h2 class="has-text-align-center">Group Title</h2>
<!-- /wp:heading -->

<!-- wp:columns {"align":"wide"} -->
<div class="wp-block-columns alignwide"><!-- wp:column -->
<div class="wp-block-column"><!-- wp:image {"sizeSlug":"large"} -->
<figure class="wp-block-image size-large"><img src="https://raw.githubusercontent.com/WordPress/twentytwentyone/trunk/assets/images/Reading.jpg" alt=""/></figure>
<!-- /wp:image -->

<!-- wp:heading {"level":3} -->
<h3>Column 1</h3>
<!-- /wp:heading -->

<!-- wp:paragraph -->
<p>Lorem ipsum dolor sit amet, commodo erat adipiscing elit. Sed do eiusmod ut tempor incididunt ut labore et dolore.</p>
<!-- /wp:paragraph --></div>
<!-- /wp:column -->

<!-- wp:column -->
<div class="wp-block-column"><!-- wp:image {"sizeSlug":"large"} -->
<figure class="wp-block-image size-large"><img src="https://raw.githubusercontent.com/WordPress/twentytwentyone/trunk/assets/images/Reading.jpg" alt=""/></figure>
<!-- /wp:image -->

<!-- wp:heading {"level":3} -->
<h3>Column 2</h3>
<!-- /wp:heading -->

<!-- wp:paragraph -->
<p>Integer enim risus, suscipit eu iaculis sed, ullamcorper at metus. Venenatis nec convallis magna, eu congue velit.</p>
<!-- /wp:paragraph --></div>
<!-- /wp:column --></div>
[...]

PR #870, opened 2020-11-23
Editor: Remove padding when columns are placed inside groups

carolinan:
Partial fix for https://github.com/WordPress/twentytwentyone/issues/865

## Summary

I tried several different ways to limit the padding in the columns editor style.
But because `.wp-block[data-align="full"]` can wrap any block and not only the nested column block, 
I was not able to make it work.

Because of that, I added an exception to the group block style instead. I'm not really happy about it, but it does remove the padding. Feel free to improve it.

## Test instructions

1. Copy the test content from the issue.
1. View it in the editor.
1. Confirm that there is no padding for the text that is inside the columns that are place in the group block.

1. Add a column outside the group block, and set it to full width.
1. Add text content.
1. Confirm that there is still padding around the text, which prevents the text from going up to the edge of the browser window, keeping it readable.

## Screenshots

After:
columns-after

## Quality assurance

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -975,6 +975,17 @@ a:hover {
 	margin-bottom: 0;
 }
 
+.wp-block[data-align=full] .wp-block-group .wp-block-columns p:not(.has-background),
+.wp-block[data-align=full] .wp-block-group .wp-block-columns h1:not(.has-background),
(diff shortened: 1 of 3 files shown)


kjellr:
The issue here was just with the specificity of the extra padding rules for the column block children. I opened #874 to fix. 

If that one works for everyone, this one can be closed.

PR #871, opened 2020-11-23
Increase range input thumb size

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/pull/861#issuecomment-731628911

## Summary

Increases the clickable size of the thumb in range input elements to the minimum size 44x44px.

## Test instructions

1. In the editor, add a range input element:
```
<!-- wp:html -->
<input type="range">
<!-- /wp:html -->

```
1. View the element on the front.
1. Test to make sure that the size is correct in different browsers.

## Screenshots

Top: Chrome
Bottom: Firefox
range

## Quality assurance


kjellr (approved):
👍 Makes sense, works well.

PR #875, opened 2020-11-23
TwentyTwentyOne: Add bottom-border to hover states in pagination links

### LINKED ISSUE #868: Post pagination current item style
**Describe the bug**

The styling for the current item in the post pagination (new/ older posts) does not match the hover style.
The left is the current page, right is a link when hovered.

current-page

allancole:
Fixes: #868

## Summary

This PR simply adds bottom-borders to the hover state on pagination links to make them match the current page bottom-border. 

## Test instructions

1. Visit any post archive page (search, category, etc.)
2. Scroll to the bottom and hover over the pagination links.
3. You should see a thin underline beneath the hovered link that matches the bottom-border for the _current_ page text. (see screenshots below)

## Screenshots
Before: 

After: 

## Quality assurance

### DIFF
--- a/assets/sass/06-components/pagination.scss
+++ b/assets/sass/06-components/pagination.scss
@@ -163,8 +163,10 @@
 		margin-left: calc(0.66 * var(--global--spacing-unit));
 		margin-right: calc(0.66 * var(--global--spacing-unit));
 
-		&.current {
+		&.current,
+		&:not(.dots):hover {
 			border-bottom: 1px solid var(--pagination--color-text);
+			text-decoration: none;
 		}
 
 		&:first-child {
--- a/assets/sass/06-components/pagination.scss
+++ b/assets/sass/06-components/pagination.scss
@@ -11,6 +11,7 @@
 		&:hover {
 			color: var(--global--color-primary-hover);
 			text-decoration: underline;
+			text-decoration-style: dotted;
 		}
 
 		&:focus {
@@ -167,7 +168,7 @@
 			text-decoration: underline;
 		}
 
-		&:not(.dots):hover {
+		&:not(.dots):not(.current):hover {
 			text-decoration-style: dotted;
 		}
 
--- a/assets/sass/06-components/pagination.scss
+++ b/assets/sass/06-components/pagination.scss
@@ -167,7 +167,7 @@
 			text-decoration: underline;
 		}
 
-		&:not(.dots):hover {
+		&:not(.dots):not(.current):hover {
 			text-decoration-style: dotted;
 		}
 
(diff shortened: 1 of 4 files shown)


melchoyce:
LGTM 👍

carolinan:
I don't think these borders would match any other link hover styles? I don't think it looks bad but it will be inconsistent?

melchoyce:
Oh, you're right — I had them reversed in my head 😵 💫  Ideally the post navigation styles would match the menu styles: solid action, dotted hover:

image

allancole:
Ahhh, I didn’t catch that! I’ll reverse this so it matches the menu styles.

allancole:
Should be good to go now @melchoyce & @carolinan: 

Header Nav:



Pagination:

allancole:
Also just noticed the font-sizes don’t match between the header navigation and pagination links:

20px header nav| 24px pagination
------------|------------
image|image

Since the hover actions are the same, should the font-sizes also the same @melchoyce ?

melchoyce:
I think I want to leave them as-is. Thanks for checking!

ryelle:
I noticed that the current item has the hover state too, but it's not a link. That could be confusing, I'm not sure if it's technically an a11y issue or not. Could `.current` also be excluded from the hover state?

assets/sass/06-components/pagination.scss
```diff
@@ -164,7 +164,11 @@
 		margin-right: calc(0.66 * var(--global--spacing-unit));
 
 		&.current {
-			border-bottom: 1px solid var(--pagination--color-text);
+			text-decoration: underline;
+		}
+
+		&:not(.dots):hover {
```
carolinan:
```suggestion
		&:not(.dots):not(.current):hover {
```

carolinan:
Just a note that with this change we have solid underlines for next and previous posts (single view), but dotted underlines for the newer/older post pagination. @melchoyce Is that what you want?

allancole:
> Could .current also be excluded from the hover state?

Ahh, good call. I’ll correct that :-)

allancole:
> Just a note that with this change we have solid underlines for next and previous posts (single view), but dotted underlines for the newer/older post pagination. @melchoyce Is that what you want?

I went ahead and corrected this in that last batch of commits. Feels more consistent this way but happy to change it back if you prefer, @melchoyce. Cc: @carolinan

melchoyce:
Can you add another gif for comparison @allancole?

allancole:
No prob, @melchoyce. See below:

### Archive View
Before: 


After:


### Singular View
Before: 


After:

pattonwebz:
The _after_ gifs look real nice to me :heart:

melchoyce:
Same, thanks @allancole 👍

allancole:
I think this is good to merge then, @carolinan—if you’d like to do the honors ;-)

carolinan:
I am very confused about why one of the after gifs have both a dotted and solid underline?

PR #879, opened 2020-11-24
Override cover block link colors

### LINKED ISSUE #858: Button links in cover block inherit the wrong link colors
When creating a cover block with a call to action button, the `color: currentColor` style has the wrong priority and ends up overwriting the intended color.

This seems to be the same issue as with #209, and originates from styles which accounts for regular text links, but not buttons, found at at https://github.com/WordPress/twentytwentyone/blob/d4c6fecb56d7e8fbfa859129057b8d1444e03063/style.css#L2039-L2046

Example:
Cover block with a background image and a button.
The button has a blue background, and white text chosen from the available color palette, this applies the `has-white-color` class to the button link.

The result:
dark button with dark text

Expected result:
dark button with white text

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/858

## Summary

The cover block sets the links to currentColor so that the text and link color has a high contrast against a chosen palette color.
This style was overriding the button block and file block link colors.

## Test instructions

1. Add the sample content from the issue.
1. Add a file block with a download button.
1. Save and view the front
1. Compare the colors and confirm that all the text is readable.
1. Test hover and focus styles.
1. Turn Dark Mode on or off, test both body background colors.

## Screenshots

After:
Dark Mode
light-body-color 

Light body background:

dark-body-color

## Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -2848,12 +2848,12 @@ input[type=reset]:hover {
 	margin-bottom: 30px;
 }
 
-.wp-block-cover .wp-block-cover__inner-container a,
-.wp-block-cover .wp-block-cover-image-text a,
-.wp-block-cover .wp-block-cover-text a,
-.wp-block-cover-image .wp-block-cover__inner-container a,
-.wp-block-cover-image .wp-block-cover-image-text a,
-.wp-block-cover-image .wp-block-cover-text a {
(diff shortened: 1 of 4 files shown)


ryelle (approved):
I think I've tested the combinations described in the issue, and a few more (outline button style, other blocks). It's working as expected, definitely better than trunk.

PR #882, opened 2020-11-24
Editor: Update button CSS variables for dark backgrounds.

### LINKED ISSUE #880: Editor: With Dark Mode enabled, buttons are only white if the selected body background is the default color
**Describe the bug**
If the user selects or keeps the mint green body background color, and enables dark mode,
the button block color is changed to white in the editor and the front.

If the user selects another light body background color from the palette or color picker, and enables dark mode,
the button block color remains dark grey in the editor, but is changed white on the front.

If Dark Mode is disabled, and the user selects a dark body background, the button block color is changed to white in the editor and the front.

**To Reproduce**
1. Enable Dark Mode
2. Change the body background color to a light palette color, but not mint.
3. _-It does not matter in which order you enable dark mode or change the color_
4. Add a default button block in the editor.
5. See that the editor background color is grey and the button is the default dark grey, not the white color.
6. Add some text in the content and confirm that the text color is white.

**Expected behavior**
With Dark Mode, the buttons colors should be adjusted to improve readability, even if the color selected in the customizer option is light.

**Screenshots**

- WordPress version: 5.6-RC1-49687, with theme version from lastest trunk on GitHub
- Does the website has Gutenberg plugin installed, or is it using the block editor that comes by default?  Default

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/880

## Summary

Adds the CSS variables from sass\style-dark-mode.scss into classes\class-twenty-twenty-one-dark-mode.php.
This adds the button CSS variables to the editor.

We primarily want the button related variables, but I added the `--global--color-border` just in case.

## Test instructions

1. Enable Dark Mode
1. Change the body background color to a light palette color, but not mint.
1. -It does not matter in which order you enable dark mode or change the color
1. Add a default button block in the editor.
1. See that the editor background color is grey and the button is white.
1. Add some text in the content and confirm that the text color is white.

## Quality assurance

### DIFF
--- a/classes/class-twenty-twenty-one-dark-mode.php
+++ b/classes/class-twenty-twenty-one-dark-mode.php
@@ -65,7 +65,7 @@ class Twenty_Twenty_One_Dark_Mode {
 			// Add Dark Mode variable overrides.
 			wp_add_inline_style(
 				'twenty-twenty-one-custom-color-overrides',
-				'.is-dark-theme.is-dark-theme .editor-styles-wrapper { --global--color-background: var(--global--color-dark-gray); --global--color-primary: var(--global--color-light-gray); --global--color-secondary: var(--global--color-light-gray); }'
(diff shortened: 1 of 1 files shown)


ryelle (approved):
Looks good 👍 

> I added the `--global--color-border` just in case.

Good call, this is needed to make the editor & frontend match — before the editor was using pure white borders while the frontend uses this grey color.

PR #884, opened 2020-11-24
Social Links Block: Fix focus on default block style

ryelle:
The white focus background on the Social Links block makes the white icons functionally invisible. This PR updates the focus color to `--global--color-primary`, which is dark grey and has contrast against the white background.

## Screenshots

Before:
Screen Shot 2020-11-24 at 12 30 27 PM

After:
Screen Shot 2020-11-24 at 12 30 35 PM

(The `li`'s background is slightly visible through the semi-transparent `a:focus` background, but I don't _think_ we need to fix anything there.)

## Test instructions

1. Add a Social Links block and fill in some links, either default or "pill" style
2. View the post/page, not in dark mode
3. Tab through to focus on the social links
4. You should be able to see the social icon against the background

Check that there are no regressions with the other styles/dark mode.

## Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -4997,6 +4997,10 @@ hr.wp-block-separator.is-style-dots:before {
 	border-color: currentColor;
 }
 
+.wp-block-social-links a:focus {
+	color: #28303d;
+}
+
 .wp-block-social-links.is-style-twentytwentyone-social-icons-color a {
 	color: #28303d;
 }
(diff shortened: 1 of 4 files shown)


carolinan:
It might be worth exploring a different focus style than the very "blocky" background color for this block and for the social icons for the links in the footer menu. Perhaps just reuse the size change that is the default for :hover  🤔 .

But I am torn because we are so late in the release schedule now. It does not feel like a new style is critical enough to be added in RC2, while this PR solves a real problem.

PR #877, opened 2020-11-24
Use black for focused links over a light body background

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/873

## Summary

Increases the readability of focused links over light body backgrounds, by using a black link background color that has a higher contrast ratio against the body background.

Removes the ` has-background-white ` CSS class and implements the previously unused `is-light-theme` class instead.
`is-dark-theme` is still used for dark backgrounds.

## Relevant technical choices:

The theme has CSS variables for link text colors for example for the primary navigation, pagination and footer area.
To override this, the `is-*-theme` body classes are used to increase the specificity.

## Test instructions

This PR can be tested by following these steps:
1. Add a transparent logo.
1. In the Customizer, change the body background color.
1. Focus on a link: Preferably, tab through an entire page and test all links.
1.  Change color and repeat.
1. Confirm that the colors are readable
1. Confirm that the colors do not affect the transparent logo

Also pay attention to footer credit links and social icons in the footer menu.

## Screenshots

After:
focus-after

## Quality assurance
* [x] I have thought about any security implications this code might add.
* [x] I have checked that this code doesn't impact performance (greatly).
* [x] I have tested this code to the best of my abilities
* [x] I have checked that this code does not affect the accessibility negatively

PR #885, opened 2020-11-24
Try: Use luminance of color to check for "white" background

### LINKED ISSUE #873: With an off-white page background, text focus styles are hard to see
**Describe the bug**

If the page background is set to an off-white color (I used `#f9f9f9` in my testing), the focus styles for buttons and links is hard to see: 

**Expected behavior**

I expected something more akin to the dark focus background used when the background is set to white. 

**Screenshots**

Navigation:
Screen Shot 2020-11-23 at 3 11 48 PM

Post title:
Screen Shot 2020-11-23 at 3 14 48 PM

Post meta:
Screen Shot 2020-11-23 at 3 14 55 PM

ryelle:
This PR is an alternate solution to #873 (other solution is #877). This expands what colors are considered "white" for the higher-contrast black focus state.

This addresses @melchoyce's comment,

> Is there any way to keep the white focus styles in the colorful schemes?

Fixes #873

## Summary
Instead of checking for exactly `#ffffff`, this checks whether the luminance of the color is greater than 225. This ensures most of the default background colors have a white focus color, while custom light-grey (like `#eee`) will trigger the black focus style. The default yellow color also has a high-enough luminance that it triggers the black focus.

In the interest of not changing the CSS, I'm sticking with the class name `has-background-white`; I'm on the fence about whether it's too unclear though.

## Test instructions

1. Play with a number of background colors. The focus style should be visible.

## Screenshots

I've got a lot of screenshots, each one the same page with the title focused.

<details>
<summary>View screenshots</summary>

black

grey

blue

red

yellow

white

Dark Mode:
dark-mode

Custom color: `#eeeeee`
custom-eee

Custom color: `#eef4d7`
custom-eef4d7

</details>

## Quality assurance

### DIFF
--- a/assets/js/customize-preview.js
+++ b/assets/js/customize-preview.js
@@ -15,17 +15,21 @@
 			if ( isDark ) {
 				document.body.classList.add( 'is-dark-theme' );
 				document.documentElement.classList.add( 'is-dark-theme' );
+				document.body.classList.remove( 'is-light-theme' );
+				document.documentElement.classList.remove( 'is-light-theme' );
 				document.documentElement.classList.remove( 'respect-color-scheme-preference' );
 			} else {
 				document.body.classList.remove( 'is-dark-theme' );
 				document.documentElement.classList.remove( 'is-dark-theme' );
+				document.body.classList.add( 'is-light-theme' );
+				document.documentElement.classList.add( 'is-light-theme' );
 				if ( wp.customize( 'respect_user_color_preference' ).get() ) {
 					document.documentElement.classList.add( 'respect-color-scheme-preference' );
 				}
 			}
 
 			// Toggle the white background class.
-			if ( '#ffffff' === to.toLowerCase() ) {
+			if ( 225 <= lum ) {
 				document.body.classList.add( 'has-background-white' );
 			} else {
 				document.body.classList.remove( 'has-background-white' );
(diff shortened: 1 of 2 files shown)


kjellr (approved):
This works well in my testing! Fixes https://github.com/WordPress/twentytwentyone/issues/873, while also keeping the white hover color in the theme's suggested color schemes as @melchoyce requested in https://github.com/WordPress/twentytwentyone/pull/877#issuecomment-733081496.

carolinan:
I understand why we want to keep the white background, but it turns out it is not enough on it's own, added the contrast ratios here:
https://github.com/WordPress/twentytwentyone/issues/873#issuecomment-733451555

For some links the fix is very easy, change or remove the style of the underline. With a text decoration change that not relies on color, we can use the white.

assets/js/customize-preview.js
```diff
@@ -15,17 +15,21 @@
 			if ( isDark ) {
 				document.body.classList.add( 'is-dark-theme' );
 				document.documentElement.classList.add( 'is-dark-theme' );
+				document.body.classList.remove( 'is-light-theme' );
```
carolinan, line 18: We no longer need the is-light-theme class here, unless we want to add it for other people to use?

aristath: Yeah, I'd rather keep it even if we don't use it... It will make customizations easier for those who will want to write custom css (`body.is-light-theme` is easier than `body:not(.is-dark-theme)`)

aristath (approved):
LGTM :+1:

PR #889, opened 2020-11-25
Try: Add a dotted underline to focused links

carolinan:
Follow up on https://github.com/WordPress/twentytwentyone/issues/873

## Summary

Adds the dotted underline (which is already used on hover) to focused links.
This _additional_ text decoration change does not rely on color, so the contrast ratio between the body light background and the white link focus background is less important.

## Test instructions

1. View the front of the site
1. Place focus on links for example using tab.
1. See the dotted underline.

## Screenshots

Before:
before-solid

After:
dotted

## Quality assurance

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -338,6 +338,8 @@ a:hover {
 
 	/* Only visible in Windows High Contrast mode */
 	outline: 2px solid transparent;
+	text-decoration-style: dotted;
+	text-decoration-skip-ink: none;
 	background: rgba(255, 255, 255, 0.9);
 }
 
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -338,7 +338,7 @@ a:hover {
 
 	/* Only visible in Windows High Contrast mode */
 	outline: 2px solid transparent;
-	text-decoration: underline dotted currentColor;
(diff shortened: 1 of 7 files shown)


ryelle:
Working great when the site has a light background 👍 

Should this also be the case in dark mode? [this rule in `style-dark-mode.scss`](https://github.com/WordPress/twentytwentyone/blob/49b3ca31e39d792793be0b2da3817d27ebc6dbba/assets/sass/style-dark-mode.scss#L19) is overriding the underline.

Screen Shot 2020-11-30 at 11 44 55 AM

Also, when the site background is dark (not dark mode), the text color is the background color — so this has an underline, but it's not visible since its the same color as the background. Since the contrast is high, maybe this is okay?

Screen Shot 2020-11-30 at 11 42 57 AM

carolinan:
Hmm I think regular dark color and dark mode should have the same focus styles.

carolinan:
I remember now that we set the dark mode focus style to black on purpose because white is too bright to work well for dark mode.

carolinan:
Yes on dark backgrounds (not dark mode) it is sometimes visible, also depending on the underline offset, but the contrast is high enough.

dark-background

carolinan:
Dark mode:
dark-mode

aristath (approved):
LGTM 👍

PR #892, opened 2020-11-25
CSS: Remove `has-background-background-color` selector

ryelle:
The `has-background-background-color` is leftover from the original theme, which had [a color called `background`](https://github.com/Automattic/themes/blob/e878ddb802a713e66660b9ae4d1eddb48a59489d/seedlet/functions.php#L187-L191) in the palette. This rule was intended to exclude any block with that background color, but this color doesn't exist in Twenty Twenty-One, so this class is never applied anywhere. It should be safe to remove, with no effect on the theme.

## Test instructions

1. Make sure blocks with background colors look correct — there should be no visual changes between this branch & trunk.

## Quality assurance

### DIFF
--- a/assets/sass/07-utilities/color-palette.scss
+++ b/assets/sass/07-utilities/color-palette.scss
@@ -43,7 +43,7 @@
 // Gutenberg background-color options
 .has-background {
 
-	&:not(.has-background-background-color) a:not(.wp-block-button__link),
+	a,
 	p,
 	h1,
 	h2,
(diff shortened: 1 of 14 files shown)


assets/sass/07-utilities/color-palette.scss
```diff
@@ -43,7 +43,7 @@
 // Gutenberg background-color options
 .has-background {
 
-	&:not(.has-background-background-color) a:not(.wp-block-button__link),
```
ryelle, line 46: Removing the `:not(.wp-block-button__link)` is intentional too — this seems to be unnecessary, since the button link styles have a higher specificity.

carolinan:
Yay for cleaning up

PR #896, opened 2020-11-27
Fix button cursor by removing the crop-text mixin

### LINKED ISSUE #878: Bug: Cursor is not visible when inserting a new button.
**NB! Please skip straight to this comment:** 
https://github.com/WordPress/twentytwentyone/issues/878#issuecomment-732961808

------------------

**Describe the bug**
Using the Buttons block it looks like this in the backend:
Screen Shot 2020-11-24 at 11 38 14

In comparison with theme Twenty Twenty.

Screen Shot 2020-11-24 at 11 29 22

**To Reproduce**
1. Add a Buttons block.
2. Add a few buttons.
3. Notice the difference between button one and button two.

**Expected behavior**
All buttons in one nice row.

- WordPress version: 5.5.3
- Gutenberg version: 9.4.1

No other plugin is activated.

Running locally through Desktop Server. 
Browser: Chrome
Macbook Pro: OSX 10.15.7

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/878

## Summary

When a button block is added, the cursor needs to be visible to indicate that you can start entering the button text. 

If my understanding is correct, the crop-text mixin sets the content width to 0, and it means that the cursor is not visible.
https://github.com/WordPress/twentytwentyone/blob/trunk/assets/sass/02-tools/mixins.scss#L33

The crop-text mixin is only used for the buttons block.
The code comment does not seem to be accurate, or may have been accurate in a previous version of the theme:
"- Sets a fixed-width on content **within alignwide and alignfull blocks**"

I do not see any unexpected side effects in 5.5.3 or 5.6 RC when removing the mixin.
Other options would be:
**-Change the width of the content from 0 so that the cursor is shown correctly. I was not able to determine if this was viable because I do not know what problem the mixin is meant to solve.
-Only apply the mixin on the front, not the editor.**

## Test instructions

1. Add a button block
1. Confirm that the cursor is visible immediately.
1.

## Screenshots
cursor

## Quality assurance

### DIFF
--- a/assets/css/ie-editor.css
+++ b/assets/css/ie-editor.css
@@ -396,22 +396,6 @@ a:hover {
 	padding: 15px 30px;
 }
 
-.wp-block-button__link:before,
-.wp-block-button__link:after {
-	content: "";
-	display: block;
-	height: 0;
-	width: 0;
-}
-
-.wp-block-button__link:before {
-	margin-bottom: -calc(1em - 0);
-}
-
-.wp-block-button__link:after {
-	margin-top: -calc(1em - 0);
-}
-
 .wp-block-button__link:focus {
 	background: transparent;
 	outline-offset: -6px;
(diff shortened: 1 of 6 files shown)


kjellr:
Weird! If that mixin is only used here, it seems fine to remove. @allancole do you happen to know the origin of the `crop-text` mixin?

PR #897, opened 2020-11-28
Add 'after_page_number' to paginations

qtxh:
## Summary
I added 'after_page_number' for pagination, in order to fit more languages.

When I translate the theme on translate.wordpress.org, I've found out that there is no relevant original string for pagination expressions that puts "page" or similar words in other languages after the page number. For example, in Chinese we say "第1页" 
 (lit. "No. 1 Page"), instead of "页1" (lit. "Page 1") — which is no Chinese expression.

## Quality assurance
* [ ] I have thought about any security implications this code might add.


carolinan:
Hi!
Please add translator contexts for the new text string and make sure that the code follows the coding standards; see the PHPCS error in the checks.

qtxh:
OK, I will withdraw my PR and revise the patch until it meets the coding standards. Thanks.

PR #899, opened 2020-11-30
Update sub menu toggle position

### LINKED ISSUE #898: Submenu toggle is too close to the menu item text
**Describe the bug**
When focus is on the sub menu toggle in the primary menu, there is no space between the menu item text and the button outline.

This outline is also 2px, reducing it to 1px would match the line around the expanded sub menu.

Top: Chrome
Bottom: Firefox

submenu-toggle

thickness

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/898

## Summary

Moves the button 5px to the right.
Reduces the outline thickness from 2 to 1px.

## Test instructions

1. Add a menu with sub menus to the primary menu location
1. focus on the sub menu toggle button
1. click on the sub menu toggle button
1. Confirm that the text is readable

## Screenshots
Before:
submenu-toggle

After
toggle-after3
toggle-after2
toggle-after

## Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -6812,7 +6812,7 @@ h1.page-title {
 }
 
 .primary-navigation > div > .menu-wrapper .sub-menu-toggle:focus {
-	outline: 1px solid #28303d;
+	outline: 2px solid #28303d;
 }
 @media only screen and (max-width: 481px) {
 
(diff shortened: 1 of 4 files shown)


melchoyce:
Do we have 1px borders anywhere else? Honestly wouldn't mind keeping them 2px and just moving a tiny bit away from the text, especially since the further away the plus gets, the more it feels like it's not a part of the menu item.

carolinan:
I reverted it so that the border is 2px. I kept the distance at 5px, so that the actual distance between the letter and the border is 3px.

Result: 
final

And just a comparison of the thin and thick outlines overlapping so that we can see the difference.
overlap

PR #904, opened 2020-11-30
Gradients: Add position values to gradients

### LINKED ISSUE #903: Bug: gradient presets shows linear input selections when radial is selected
**Describe the bug**
When using a group block and a paragraph inside I stumbled upon a bug in the ”Background Color”-section in the ”Color Settings”-panel in the editor. When selecting a gradient preset, I can select Radial in the dropdown menu. When doing so the input fields for linear manipulation is still visible, even though they should be hidden. (Compare to other default Twenty-themes.) When changing the values in the input-fields, the block crashes.

**To Reproduce**
1. Create a new post
2. Add a group block
3. Add a paragraph inside the group block, add dummy text.
4. Select group block and open the Block settings in the right sidebar.
5. Open panel ”Color settings”
6. Select ”Gradient” for background color.
7. Select a Gradient-preset
8. Change Type to ”Radial”
9. Change the value 160 to 120 or any other number
10. Block crashes. ”This block has encountered an error and cannot be previewed.”

**Expected behavior**
The input fields from Linear-gradient should be hidden when selecting Radial.

**Screenshots**
image
image

- WordPress version: 5.6 RC1
- Does the website has Gutenberg plugin installed, or is it using the block editor that comes by default? default
- If the Gutenberg plugin is installed, which version is it? [e.g., 8.9]

 - OS: MacOS 10.14.6
 - Browser: Firefox
 - Version 81.0.2

**Additional context**

ryelle:
While not required by CSS, the editor uses the position values in `linear-gradient` to set the control points in the gradient picker. This PR adds the default control points (0%, 100%) explicitly, so that the editor doesn't crash trying to use them. 

(an upstream fix could be made to understand these implicit defaults, but it wouldn't ship in 5.6.)

Fixes #903.

## Test instructions

1. Add any block that supports gradient background colors (for example, Group)
2. Set a gradient background
3. Try editing the gradient- move a control point, add another, change the angle…
4. You should be able to do all of the above without the block crashing

## Screenshots

| Before (no control points) | After (gradient editor highlighted) |
|---------------------------|--------------------|
| Screen Shot 2020-11-30 at 11 27 17 AM | Screen Shot 2020-11-30 at 11 24 15 AM |

## Quality assurance

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -272,42 +272,42 @@ if ( ! function_exists( 'twenty_twenty_one_setup' ) ) {
 			array(
 				array(
 					'name'     => esc_html__( 'Purple to yellow', 'twentytwentyone' ),
-					'gradient' => 'linear-gradient(160deg, ' . $purple . ', ' . $yellow . ')',
+					'gradient' => 'linear-gradient(160deg, ' . $purple . ' 0%, ' . $yellow . ' 100%)',
 					'slug'     => 'purple-to-yellow',
 				),
 				array(
 					'name'     => esc_html__( 'Yellow to purple', 'twentytwentyone' ),
-					'gradient' => 'linear-gradient(160deg, ' . $yellow . ', ' . $purple . ')',
+					'gradient' => 'linear-gradient(160deg, ' . $yellow . ' 0%, ' . $purple . ' 100%)',
 					'slug'     => 'yellow-to-purple',
 				),
 				array(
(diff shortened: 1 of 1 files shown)


kjellr (approved):
Good fix. Thanks for sorting this out!

I've opened an issue here to track an upstream fix: https://github.com/WordPress/gutenberg/issues/27382

carolinan:
Looks good to me, thank you.

PR #886, opened 2020-11-24
TwentyTwentyOne: Restructure button style cascade to improve button color options.

allancole:
Fixes #858
_Possibly_ fixes: #880

## Summary
Continues with @carolinan work in #879 so im borrowing the summary from there. The cover block sets the links to currentColor so that the text and link color has a high contrast against a chosen palette color. This style was overriding the button block and file block link colors. 

Additionally, this PR attempts to address all Blocks that use a button and use the same border-width style consistently. 
## Test instructions

This PR can be tested by following these steps:

1. Create a post with this [sample content](https://cloudup.com/cXTzNhag6EV)
2. Save and view the front
3. Compare the colors and confirm that all the text is readable.
4. Test hover and focus styles.
5. Turn Dark Mode on or off, test both body background colors.
6. Repeat step 5 with the Block editor.

## Screenshots (UPDATED: 2020-11-25)

Before | After | Before (darkmode) | After (darkmode) 
------------|------------|------------|------------
image|image|image|image

## Todo

- [x] Finish solution for nested button colors on a lighter background color.
- [ ] Test and apply this fix to the editor.

## Quality assurance
* [x] I have thought about any security implications this code might add.
* [x] I have checked that this code doesn't impact performance (greatly).
* [x] I have tested this code to the best of my abilities
* [ ] I have checked that this code does not affect the accessibility negatively

PR #909, opened 2020-12-01
Update search block focus style

### LINKED ISSUE #908: Search block: When the button is inside, the input field has no visible focus indicator
**Describe the bug**

When a search block is added and the button is placed inside, there is no visible focus indicator. 
-A blinking cursor is not enough to indicate focus.

The block editor sets the outline to none, so the themes focus outline is not visible.

```
.wp-block-search.wp-block-search__button-inside .wp-block-search__inside-wrapper .wp-block-search__input:focus {
    outline: none;
}
```

**To Reproduce**
1. Add a default search block. Duplicate it.
2. Select the button inside option
3. Duplicate the block and select the icon only, button inside options.
4. Go to the front.
5. focus on the blocks. See that there is a difference in the focus style for the 3 blocks.

**Expected behavior**

There needs to be a visible focus indicator.

**Screenshots**
search-focus-indicator

- WordPress version: 5.6-RC1-49715
- Does the website has Gutenberg plugin installed, or is it using the block editor that comes by default?  Default.

 - OS: Windows 10
 - Browser: Chrome, version 87.0.4280.66

carolinan:
Fixes https://github.com/WordPress/twentytwentyone/issues/908

## Summary

When the button is placed inside:
Overrides the `outline: none`. Adjusts the offset from 7 to 5 so that it matches the button, and takes the border change in 
https://github.com/WordPress/twentytwentyone/pull/886 in consideration.

The padding(?) here makes this style less than ideal ( **OK, its really ugly** ), but there needs to be a visible focus indicator.

## Test instructions

1. Add a default search block. Duplicate it.
1. Select the button inside option
1. Duplicate the block and select the icon only, button inside options.
1. Go to the front.
1. Place focus on the blocks. Compare the focus style for the 3 blocks.

## Screenshots

Top: Default, with focus styles. 
Bottom: Button inside, with focus styles.

Both the input and the button is in focus here using browser dev tools, to show how the dotted outline aligns.

search-focus-indicator-after gif

## Quality assurance

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -4778,6 +4778,11 @@ pre.wp-block-preformatted {
 	color: #39414d;
 }
 
+.wp-block-search.wp-block-search__button-inside .wp-block-search__inside-wrapper .wp-block-search__input:focus {
+	outline: 2px dotted #39414d;
+	outline-offset: -5px;
+}
+
 .wp-block-search.wp-block-search__button-inside.wp-block-search__text-button button.wp-block-search__button {
 	padding: 15px 30px;
 }
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -4778,6 +4778,10 @@ pre.wp-block-preformatted {
 	color: #39414d;
 }
 
+.wp-block-search.wp-block-search__button-inside .wp-block-search__inside-wrapper .wp-block-search__input {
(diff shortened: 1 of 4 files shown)


carolinan:
Adjusted margin and made it a little less ugly
search-focus-indicator-after

PR #910, opened 2020-12-01
Prevent link focus styles from affecting the buttons block

carolinan:
Fixes an issue where the link focus style is affecting the button block and the file download button

## Summary

Adds a :not for the button link and the file block download button.

## Test instructions

1. Add a button, make sure the button has a link.
1. Focus on the button.
1. Confirm that the link text has no underline, and that the background color for the link is not changed.

## Quality assurance


kjellr (approved):
This seems to work fine on my end.

PR #911, opened 2020-12-01
Prevent link focus styles from affecting the buttons block (Dark mode)

kjellr:
Same as #910, but for dark mode.

### DIFF
--- a/assets/sass/style-dark-mode.scss
+++ b/assets/sass/style-dark-mode.scss
@@ -12,7 +12,7 @@
 		--button--color-background-active: var(--global--color-background);
 		--global--color-border: #9ea1a7;
 
-		.site a:focus,
+		.site a:focus:not(.wp-block-button__link):not(.wp-block-file__button),
 		.site a:focus .meta-nav {
 			background: #000;
 			color: #fff;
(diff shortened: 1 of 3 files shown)


carolinan:
Ah, there is one link text that has the wrong color:
image

kjellr:
I think that's unrelated to this PR. I'm seeing these in `trunk` too:

Screen Shot 2020-12-01 at 11 51 51 AM
 too: 

Screen Shot 2020-12-01 at 11 50 09 AM

carolinan:
it is because of the :not style added to the link focus style

carolinan:
I mean for the meta nav

assets/sass/style-dark-mode.scss
```diff
@@ -12,7 +12,7 @@
 		--button--color-background-active: var(--global--color-background);
 		--global--color-border: #9ea1a7;
 
-		.site a:focus,
+		.site a:focus:not(.wp-block-button__link):not(.wp-block-file__button),
 		.site a:focus .meta-nav {
```
carolinan:
```suggestion
		.site a:focus:not(.wp-block-button__link):not(.wp-block-file__button) .meta-nav {
```

carolinan: This feels very hacky but this is the best I can come up with because of stress.

PR #912, opened 2020-12-01
Try fixing the border for search fields with inset buttons

kjellr:
Before: 

Screen Shot 2020-12-01 at 11 47 48 AM

After:

Screen Shot 2020-12-01 at 11 47 38 AM

### DIFF
--- a/assets/css/ie.css
+++ b/assets/css/ie.css
@@ -4568,12 +4568,6 @@ pre.wp-block-preformatted {
 	}
 }
 
-.wp-block-search.wp-block-search__button-inside .wp-block-search__inside-wrapper {
-	background-color: #fff;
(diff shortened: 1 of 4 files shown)


allancole:
Looks good! Only thing is the hover state looks like it doesn’t have enough contrast. 




Maybe on hover, the background-color can be that dark grey?

luminuu:
This PR looks good to me, I do not have these hover issues @allancole posted. WP 5.6 RC1-49715 without Gutenberg plugin.

carolinan:
I don't see the hover issue either

allancole:
Ahhh, I needed to turn off the Gutenberg plugin! Looks good to me now too :-)

